PR #1, opened 2022-09-17
Remove trailing whitespaces.

eyalz800:
Hi,

Great work, hope you don't mind to remove trailing whitespaces as it can be distracting in some IDEs.

By the way - it could be nice to sometime convert the cppfront code itself to cpp2, and see how it goes.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -34,16 +34,16 @@ auto cmdline_processor::print(std::string_view s, int width) -> void
 
 
 //-----------------------------------------------------------------------
-// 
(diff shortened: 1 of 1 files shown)


hsutter:
@eyalz800 Thanks! I noticed that this affects all the files, so I just went through and re-saved them all myself in VS Code with the trim-trailing-whitespace option set. I'll check in that commit instead of this PR. Thanks for pointing this out!

Yes, I intend to start writing parts of cppfront itself in Cpp2 syntax in the medium term.

PR #3, opened 2022-09-17
fix badge link (for real this time)

Zt-freak:


hsutter:
@Zt-freak Thanks! I'm just doing a housekeeping checkin to fix #1 so I just included this, the fix will be applied in that commit instead of this PR. Thanks for pointing it out!

Zt-freak:
very pog

PR #4, opened 2022-09-19
Added Build System Integration

Eshanatnight:
Added `Cmakelists.txt`, `.clang-format`, `.clang-tidy` and `premake5.lua` .

Because of this .gitignore file was modified to not include the build directory.
and a `.gitattributes` file was added


Eshanatnight:
forgot to push `premake5.lua`

Frityet:
Why premake?

Eshanatnight:
Actually i also wrote a cmakelists.txt but for some reason it got deleted? 

And as for why premake, after no specific reason I started using premake instead of cmake because i wanted to learn Lua and i thought the fastest way would be to use it

CMakelists.txt
```diff
@@ -0,0 +1,22 @@
...
+	source/sema.h
+	include/cpp2util.h
+)
+
+add_executable(${PROJECT_NAME} ${SRC_FILES} ${HEADER_FILES})
+target_include_directories(${PROJECT_NAME} PUBLIC include)
+
+# Tests are yet to be added
+# not exactly sure on how to do it yet
+# as it is author compuer dependent
```
JohelEGP, line 22:
```suggestion
# as it is author computer dependent
```

Eshanatnight:
Just added the Cmakelists file. Found why it was not being tracked by git, .txt files were in the `.gitignore`

JohelEGP (changes requested):
I find this unsatisfactory because the executable isn't exported. I can only use this by `add_subdirectory` and not `find_package`.

Eshanatnight:
Need to find a way to make the tests 

> I find this unsatisfactory because the executable isn't exported. I can only use this by `add_subdirectory` and not `find_package`.

Then would we not need to make the tests in to a "sub-projects" of sorts?

JohelEGP:
My comment is not in regard to tests, but to consumers of the `cppfront` executable.

Eshanatnight:
> My comment is not in regard to tests, but to consumers of the `cppfront` executable.

Oh I understand You mean you find the CMakeLists File. The executable can easily be copied/moved with a post build command. To a standard location. This should also help in making and running the tests easier, in my opinion.

.gitignore
```diff
@@ -1,4 +1,5 @@
-
+/.vs
+/.vscode
```
Colliflower, line 2:
```suggestion
```
I feel like adding editor specific folders to the .gitignore is a bad rabbit hole, you end up adding them for every IDE. This belongs in a user level .gitignore instead.

Eshanatnight: Yeah you are right. I'll change it then.

Eshanatnight:
I would like to ask for some suggestions regarding the tests. Should I classify them as a sub-project to `cppfront`. I tried running it via the `add_custom_command` but that is resulting in either by failing to find the files or giving a bunch of errors.

Colliflower:
I think it might be better to split some of the content from this PR into multiple PRs btw. There are a few things that seem to be unrelated to build systems such as the .md file changes, clang-format and clang-tidy. Those getting their own PRs would clean things up and make this PR more attractive imo.

Eshanatnight:
I apologise, my brain gets all over the place when I work, so it got a bit out of hand it that way. 

What i thought of about the testing, is that it's better if we just keep the Tests as they are atleast for now.

Eshanatnight:
The CMakeLists is complete, but I am keeping The `premake5.lua` file as imo it creates a better visual studio `sln` file than cmake.

filipsajdak:
@Eshanatnight your PR is a collection of unrelated changes (adding build system files, minor edits in Markdown files) - consider moving them to separate PRs. It will make reviewing them easier.

Eshanatnight:
> @Eshanatnight your PR is a collection of unrelated changes (adding build system files, minor edits in Markdown files) - consider moving them to separate PRs. It will make reviewing them easier.

I apologize, I will remove the misc files and edits and remove the premake file

Eshanatnight:
I have made the asked changes. 
Again apologies for the mis steps.

Arthapz:
I have made the support for XMake https://github.com/TapzCrew/cppfront-xmake

CMakelists.txt
```diff
@@ -0,0 +1,20 @@
...
+# issue with the standard library of MSVC
+# https://stackoverflow.com/questions/64889383/how-to-enable-stdclatest-in-cmake
+set(CMAKE_CXX_STANDARD 23)
+
+set(SRC_FILES
+	"source/cppfront.cpp"
+)
+
+add_executable(${PROJECT_NAME} ${SRC_FILES})
+target_include_directories(${PROJECT_NAME} PUBLIC "include")
```
JohelEGP, line 20: Compiling `cppfront` doesn't require this.

Eshanatnight: Can you please clarify what is not required?

JohelEGP: Line 20. You can compile `cppfront` without `cpp2util.h`.

hsutter:
Thanks for this idea and energy.

How about making this a separate companion project, and I can add a wiki page of links to known companion projects to help make people aware of them?

My concern about putting it here is that it complicates what I will need to maintain, particularly at this early stage when a lot of things are still fluid and subject to change. And while I realize CMake is usually the most popular and it's a meta-build system that can generate others, if we add CMake I would expect to get questions like "why not Ninja, msbuild, Xcode, Bazel, Gradle, ...?" which I'd have to answer.

Right now I've been considering "no CMake/Bazel/..." to be a feature, in that the project is simple enough that it's easy for anyone to separately build and integrate with whatever build tools and project system they're using. Is that reasonable for now?

JohelEGP:
Yes, please. I'd like to add https://github.com/JohelEGP/jegp.cmake_modules/#jegpcpp2 to the wiki, which allows to compile CMake targets with Cpp2 source files. You can see it working on CE: https://godbolt.org/z/cYf1MrPM4.

CMakelists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.2)
+project(cppfront)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY 					${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY 					${CMAKE_BINARY_DIR}/bin)
```
alexreinking, line 5: This is super unfriendly to FetchContent users and totally unnecessary.

CMakelists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.2)
```
alexreinking, line 1: CMake 3.2 predates even C++17, so this minimum is bogus.

CMakelists.txt
Eshanatnight: I am using MSVC cmd line. And the cmd line tool has a tendency to make to obfuscate the build dir imo thats why I added that.

CMakelists.txt
Eshanatnight: I am sorry I had no idea that was the case. Thanks For Letting me know.

hsutter:
> Yes, please. I'd like to add https://github.com/JohelEGP/jegp.cmake_modules/#jegpcpp2 to the wiki, which allows to compile CMake targets with Cpp2 source files. You can see it working on CE: https://godbolt.org/z/e1r3qsE8M.

Done, thanks. Added to [wiki home page](https://github.com/hsutter/cppfront/wiki)

PR #9, opened 2022-09-21
Update cpp2util.h

unknown442:
The "args" not forwarding correctly
```
auto construct(auto ...args) -> void { Default.expects(!init);  new (&t) T(std::forward<decltype(args)>(args)...);  init = true; }
auto construct_list(auto ...args) -> void { Default.expects(!init);  new (&t) T{ std::forward<decltype(args)>(args)... };  init = true; }
```

```
union 
{
   int i;
   T t; 
};
```

Also, I'm curious what is the reason for using int instead of a 1-byte wide type.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -386,8 +386,8 @@ public:
    ~deferred_init() noexcept       { if (init) t.~T(); }
     auto value()    noexcept -> T& { Default.expects(init);  return t; }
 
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks, applied! Re `int`: I can't remember any good reason why I used a `union` instead of `aligned_storage`, so I switched to `aligned_storage` which avoids the small-type issue, and also fixed the forwarding parameters in that commit too while I was there. Thanks again.

PR #10, opened 2022-09-21
Wrong index

unknown442:
Also, for this situation, I would like to use the recursive template approach because of less repetition


hsutter:
Thanks! cut-and-paste FTW

PR #12, opened 2022-09-23
Correct includes in cpp2util.h for MacOSX12.3 SDK

filipsajdak:
On MacOSX 12.3 SDK there is no <memory_resource> header there is <experimental/memory_resource> - check with __has_include to include existing one.

There was no <cuchar> and <stop_token> - add checking if it is present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ I'm not posting much documentation because that would imply this project is inte
 
 To learn more, please see:
 
-- **My CppCon 2022 talk, "Can C++ be 10x simpler and safer ...?**" [link coming soon]
+- [**My CppCon 2022 talk, "Can C++ be 10x simpler and safer ...?**"](https://www.youtube.com/watch?v=ELeZAKCN4tY)
 - **The [cppfront regression tests](https://github.com/hsutter/cppfront/tree/main/regression-tests/test-results)** which show dozens of working examples, each with a`.cpp2` file and the `.cpp` file it is translated to. Each filename briefly describes the language features the test demonstrates (e.g., contracts, parameter passing, bounds safety, type-safe `is` queries and `as` casts, initialization safety, and generalized value capture including in function expressions ('lambdas'), postconditions, and string interpolation).
 - The list of papers and talks below.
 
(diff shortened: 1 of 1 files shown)


JohelEGP:
LLVM16's Libc++ doesn't have `<memory_resource>` yet.

jcanizales:
What's the difference between `__has_include` and the `__cpp_lib_NAME` that's already used for other includes?

JohelEGP:
Seehttps://en.cppreference.com/w/cpp/preprocessor/include and https://en.cppreference.com/w/cpp/header/version.

`#include <version>` should be moved before the first use of a `__cpp_lib_`-prefixed macro.

include/cpp2util.h
```diff
@@ -87,7 +87,11 @@
         #include <variant>
         #include <version>
         #include <memory>
-        #include <memory_resource>
+        #if __has_include(<memory_resource>)
```
alexreinking, line 90: need to check `#if defined __has_include`

filipsajdak: I was thinking if should I add that check before using it. If `__has_include` is not present it means that we are dealing with so old C++ that it will not compile the rest of the file.

But you are right. For the reason of completeness I will add that.

JohelEGP: I'd wait for the owner's opinion before doing that.

alexreinking: Actually, I forgot that `__has_include` is standard as of C++17, so I retract my comment.

alexreinking: It might still be a good idea to `#error` if the fallback fails, too. Better to fail when you know there's a problem than later, on a missing symbol error or some such.

hsutter:
I see this PR is closed but I'm now adding this in a separate commit, and put `<version>` up higher. Thanks!

filipsajdak:
@hsutter sorry I was making changes in my repository and I messed things up. I will prepare a new one.

filipsajdak:
@hsutter The new one is here: https://github.com/hsutter/cppfront/pull/16

Sorry for messing things.

PR #14, opened 2022-09-24
Add CMake file + GitHub actions

dacap:
A little CMakeLists.txt file that is used by new GitHub action workflows to build cppfront on different platforms. The [only one failing](https://github.com/dacap/cppfront/actions/runs/3118528674/jobs/5057882349#step:5:11) right now is in Ubuntu with gcc 9 (it looks like the <compare> header file is not available). I can add an extra check for that header file in the CMakeLists.txt, but I think it would be better if it can be detected with a preprocessor macro (so cppfront could be still compiled with CMake).

After this is merged (or some other/modified version of this) I can add the automatic compilation and execution of all regression tests on each pushed commit or PR with cmake/ctest.

I accept the Contributor License Agreement: https://github.com/hsutter/cppfront/blob/fa65d346996ec472e16c61838fbc7a47736d7872/CONTRIBUTING.md#contributor-license-agreement

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,6 @@
+project(cppfront CXX)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+add_executable(cppfront source/cppfront.cpp)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+name: ci
+on: [push, pull_request]
+jobs:
+  build:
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+project(cppfront CXX)
```
alexreinking, line 1: This entire file is UB without `cmake_minimium_required` at the top

dacap:
Seeing Herb’s answer [here](https://github.com/hsutter/cppfront/pull/4#issuecomment-1257006060), it looks like CMake will be more a problem than a solution for him in this early stage. Closing this PR at the moment.

PR #16, opened 2022-09-25
Correct includes in cpp2util.h for MacOSX12.3 SDK

filipsajdak:
On MacOSX 12.3 SDK there is no `<memory_resource>` header there is `<experimental/memory_resource>` - check with __has_include to include existing one.

There was no `<cuchar>` and `<stop_token>` - add checking if it is present.

This PR was previously: https://github.com/hsutter/cppfront/pull/12

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -88,8 +88,10 @@
         #include <utility>
         #include <variant>
         #include <memory>
-        #ifdef __cpp_lib_memory_resource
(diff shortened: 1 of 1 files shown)


alexreinking (approved):

hsutter:
Thanks! I committed a change that adds three of the four requested feature test guards, but didn't add the fallback to `experimental/memory_resource`... I'm trying to avoid the experimental headers.

PR #17, opened 2022-09-26
UFCS: add method chaining

filipsajdak:
I have been playing with UFCS. I found that it doesn't support method chaining yet. So I have added it.

Former implementation was not supporting the method chaining - the below code failed to be compiled
```cpp
v.a().b(1,"arg").c(); // failed to use UFCS
```

The proposed change converts the above code into a nested call of `CPP2_UFCS` macros:
```cpp
CPP2_UFCS_0(c, CPP2_UFCS_0(b, CPP2_UFCS_0(a, v), 1, "arg" ) );
```

This allows using ranges library without the need of `operator|` for chaining. `operator|` does not work in cppfront at the moment. What I realized is that having UFCS that supports chaining solves the problem.

I don't have ranges working on my current system, so I put a custom example in the regression test directory.

Having the possibility of method chaining allows us to use smart pointers and pass them to the functions that need a raw pointers.

```cpp
smart.get().func(); // UFCS will be executed on raw pointer (thanks to chaining)
```

### DIFF
--- a/regression-tests/mixed-ufcs-method-chaining.cpp2
+++ b/regression-tests/mixed-ufcs-method-chaining.cpp2
@@ -1,0 +1,28 @@
+#include <vector>
+
+add: (in v : _, in a : int) -> auto = {
+    cv := v;
+    for cv do :(inout e : _) = {
+        e += a;
+    }
+    return cv;
+}
+
+dbl: (in v : _) -> auto = {
+    cv := v;
+    for cv do :(inout e : _) = {
+        e *= 2;
+    }
+    return cv;
+}
+
+main: () -> int = {
+
+    v : std::vector<int> = (1,2,3,4,5,6,7,8,9,10);
+
+    for v.dbl().add(1).dbl() do :(in e : _) = {
+        std::cout << e << ", ";
+    }
+
+    std::cout << std::endl;
+}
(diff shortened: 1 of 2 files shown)


hsutter:
Interim ack: Thanks! Yes, I've been meaning to add chaining. I was thinking of implementing it later in the same function and remove this special case part of the function (I dislike that there are two paths that deal with `.` operator differently)... let me keep this PR open without action until I get a chance to look at it again later.

filipsajdak:
One thing I found is that it is not working when called on a constructor call - maybe I will try to add it.

```cpp
std::string("this will not work").method();
```
and this
```cpp
s := std::string("this will work");
s.method();
```
will work.

hsutter:
Picking this up: How does this relate to #18?

filipsajdak:
It is for the cases when the first call is done on the object by calling its method.

#18 is for a case when there is a free function call and the chaining starts on the function result.

filipsajdak:
@hsutter sorry that I did not deliver it yet. I am trying to catch up with other changes and resolve all issues found in regression tests. I am working on it.

I will try to merge this with https://github.com/hsutter/cppfront/pull/18

filipsajdak:
@hsutter  I managed to rewrite my code to match the latest changes. It allows me to simplify, make it more generic, and merge it with #18. I will clean it up and push the change... It's already 3:30 am... I need to go to sleep and will finalize it in a couple of hours.

filipsajdak:
This PR is replaced by: https://github.com/hsutter/cppfront/pull/169

PR #37, opened 2022-10-01
Add overload for string_view for cpp2::to_string

filipsajdak:
There is no overload for `string_view` that makes code

```cpp
sv : std::string_view = "world";
std::cout << "Hello (sv)$!" << std::endl;
```
result with outcome
```
Hello (customize me - no cpp2::to_string overload exists for this type)!
```

This change adds overload to `cpp2::to_string` function for `string_view` that avoids ambiguity with `std::string` version for `const char[]`. 

The function returns `std::string` to be consistent with its name and other overloads.


filipsajdak:
I create implementation that returns `std::string`. It could return `std::string_view const&` to avoid unnecessary copy. If that is more preferable I will update this PR.

include/cpp2util.h
```diff
@@ -746,6 +746,14 @@ auto to_string(std::optional<T> const& o) -> std::string {
...
 }
 
+template<typename T>
+auto to_string(T const& t) -> std::string
+    requires (!std::is_array_v<T> // avoid ambiguity with to_string(std::string const&) in case of const char arrays
+        && std::is_same_v<T, std::string_view>)
+{
+    return {t.data(), t.size()};
+}
+
```
JohelEGP, line 756: Make it more like [this constructor](https://eel.is/c++draft/string.cons#lib:basic_string,constructor____).
```suggestion
template<typename T>
auto to_string(T const& t) -> std::string
    requires (std::is_convertible_v<const T&, std::string_view> && !std::is_convertible_v<const T&, const char*>)
{
    return std::string{t};
}

```

JohelEGP (changes requested):

include/cpp2util.h
filipsajdak: It's the middle of the night for me... but Herb is reviewing the PRs so... I will fix it :)

This PR is part of my another one -> https://github.com/hsutter/cppfront/pull/40 (the complete support for standard types). It has a similar issue -> I will fix in both PRs)

filipsajdak: OK, I will correct it to this:

```cpp
template<typename T>
auto to_string(T const& sv) -> std::string
    requires (std::is_convertible_v<T, std::string_view> && !std::is_convertible_v<T, const char*>)
{
    return std::string{sv};
}
```

filipsajdak: Now it is fixed.

I have added examples to the regression tests as well.

hsutter:
Good point, I think the basic types are useful to be able to interpolate, but I skipped the ranges support for now and tweaked a few details (notably changed the implementation for `variant` to be consistent with `is empty`). Applied as a separate commit that covers these three overlapping PRs and that I could test separately. Thanks!

PR #39, opened 2022-10-01
Fix cpp2::to_string to call cpp2::to_string instead of std::to_string

### LINKED ISSUE #38: [BUG] cpp2::to_string(std::optional<T>) is calling std::to_string
Specialisation of `cpp2::to_string` function for `std::optional` is using `std::to_string` on optional value. 
```cpp
template<typename T>
auto to_string(std::optional<T> const& o) -> std::string {
    if (o.has_value()) {
        return std::to_string(o.value());
    }
    return "(empty)";
}
```
That makes the below code fail to compile:
```cpp
v : std::optional<std::string> = "this will not work";
std::court << "(v)$" << std::endl;
```
cppfront will generate
```cpp
std::optional<std::string> v { "this will not work" }; 
std::cout << "My map: " + cpp2::to_string(v) + "\n" << std::endl;
```

The result is that there is a call to `std::to_string(std::string)` and there is no such specialization in the standard library.

filipsajdak:
Specialisation of `cpp2::to_string` function for `std::optional` is using `std::to_string` on optional value. 
```cpp
template<typename T>
auto to_string(std::optional<T> const& o) -> std::string {
    if (o.has_value()) {
        return std::to_string(o.value());
    }
    return "(empty)";
}
```
That makes the below code fail to compile:
```cpp
std::optional<std::string> v { "this will not work" }; 
std::court << "(v)$" << std::endl;
```
cppfront will generate
```cpp
std::optional<std::string> v { "this will not work" }; 
std::cout << "My map: " + cpp2::to_string(v) + "\n" << std::endl;
```

The result is that there is a call to `std::to_string(std::string)` and there is no such specialization in the standard library.

Closes #38


include/cpp2util.h
```diff
@@ -741,7 +741,7 @@ auto to_string(std::string const& s) -> std::string const&
 template<typename T>
 auto to_string(std::optional<T> const& o) -> std::string {
     if (o.has_value()) {
-        return std::to_string(o.value());
+        return to_string(o.value());
```
JohelEGP, line 744: This opens up ADL. For now, it should be
```suggestion
        return cpp2::to_string(o.value());
```

JohelEGP (changes requested):

include/cpp2util.h
filipsajdak: fixed

filipsajdak:
I have added examples of use in regression tests and results of execution.

PR #40, opened 2022-10-01
cpp2::to_string specialisation for standard types

filipsajdak:
`cpp2::to_string` function is important for easy string interpolation. The current implementation handles only types that are convertible with `std::to_string` and `std::string`. Current change adds support for:
* `std::string_view`
* `std::pair`
* `std::tuple`
* standard containers (that fulfills can be used with `std::begin()` and `std::end()`
* `std::variant`

Elements from containers are put in curly braces `{ }`, `std::string_view` is printed as `std::string`, and the rest of the types use normal braces `( )`.

## std::string_view

```cpp
sv : std::string_view = "World";
std::cout << "Hello (sv)$!\n"; // prints: Hello World!

osv : std::optional<std::string_view> = ();
std::cout << "osv = (osv)$\n"; // prints: osv = (empty)
osv = "It works, too";
std::cout << "osv = (osv)$\n"; // prints: osv = It works, too
```

## std::pair

```cpp
p : std::pair<std::string_view, std::optional<std::string>> = ("first", std::nullopt);
std::cout << "p = (p)$\n"; // prints: p = (first, (empty))
p.second = "second";
std::cout << "p = (p)$\n"; // prints: p = (first, second)
```

## std::tuple

```cpp
t : std::tuple<double, std::optional<std::pair<std::string_view, int>>, std::optional<std::tuple<int, int, int>>> = (3.14, std::nullopt, std::nullopt);
std::cout << "t = (t)$\n"; // prints: t = (3.140000, (empty), (empty))
std::get<1>(t) = std::pair(std::string_view("answer"), 42);
std::cout << "t = (t)$\n"; // prints: t = (3.140000, (answer, 42), (empty))
std::get<2>(t) = std::make_tuple(2022,10,1);
std::cout << "t = (t)$\n"; // prints: t = (3.140000, (answer, 42), (2022, 10, 1))
```

## Standard containers

```cpp
v : std::vector<int> = (1,2,3,4,5);
std::cout << "v = (v)$\n"; // prints: v = {1, 2, 3, 4, 5}

m : std::map<std::string, int> = (std::pair("A", 1), std::pair("B", 2), std::pair("C", 3));
std::cout << "m = (m)$\n"; // prints: m = {(A, 1), (B, 2), (C, 3)}

us : std::unordered_set<int> = (1,2,3,2,3,4,5,2,3,1,2,3,4,5);
std::cout << "us = (us)$\n"; // prints: us = {5, 4, 3, 2, 1}

ums : std::unordered_multiset<int> = (1,2,3,2,3,4,5,2,3,1,2,3,4,5);
std::cout << "ums = (ums)$\n"; // prints: ums = {5, 5, 4, 4, 3, 3, 3, 3, 2, 2, 2, 2, 1, 1}

mix : std::vector = (std::optional(std::make_tuple(v,m,us)));
std::cout << "mix = (mix)$\n"; // prints: mix = {({1, 2, 3, 4, 5}, {(A, 1), (B, 2), (C, 3)}, {5, 4, 3, 2, 1})}
```

## std::variant

```cpp
vv : std::variant<int, std::string, std::vector<int>> = ();
std::cout << "vv = (vv)$\n"; // prints: vv = 0
vv = 1;
std::cout << "vv = (vv)$\n"; // prints: vv = 1
vv = "string";
std::cout << "vv = (vv)$\n"; // prints: vv = string
vv = v;
std::cout << "vv = (vv)$\n"; // prints: vv = {1, 2, 3, 4, 5}
```

Examples of use are added in regression tests with results of an execution (consistently with @hsutter way). All changes are in separate commits that should help in reviewing.

The above types allow one to compose one with another - this change works for composition as well (it runs recursively).

Depends on https://github.com/hsutter/cppfront/pull/39


JohelEGP:
If only 3rd party libraries were OK, most, if not everything, could be delegated to `::fmt::print`.

hsutter:
Good point, I think the basic types are useful to be able to interpolate, but I skipped the ranges support for now and tweaked a few details (notably changed the implementation for `variant` to be consistent with `is empty`). Applied as a separate commit that covers these three overlapping PRs and that I could test separately. Thanks!

jcanizales:
> If only 3rd party libraries were OK, most, if not everything, could be delegated to `::fmt::print`.

Is there a reason to not use 3rd party libraries in principle? As long as they're C++ libraries, it's "things C++ can already do" and reimplementing them isn't useful on its own. I'm aware of practical considerations not to use them (e.g. might make things more complicated for the current user of cppfront) - but maybe in the initial stages of this project the provided boost is worth it?

filipsajdak:
@hsutter happy that it is useful. I see that you also skipped forward declarations of to_string specializations - it makes it impossible to compose these types e.g. `std::optional<std::pair<int,double>>` will not compile when trying to use with `cpp2::to_string` function.

Added https://github.com/hsutter/cppfront/pull/49

Issue #42, opened 2022-10-03
[BUG] Duplicate symbols from cpp2util.h


filipsajdak:
**Describe the bug**
In cpp2util.h file there are duplicated symbols of functions:
* `cpp2::to_string`
* `cpp2::report_and_terminate`
* `cpp2::fopen`

**To Reproduce**
1. Sample code - distilled down to minimal essentials please
`main.cpp2` file:
```cpp
#include <iostream>

main: () -> int = {
    std::cout << "Running my new app\n" << std::endl;
}
```
`test.cpp2` file:
```cpp
#include <string>

message: () -> std::string = {
    return "hello from the cpp2";
}
```
3. Command lines including which C++ compiler you are using
```
src % clang++ --version
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

src % cppfront main.cpp2 test.cpp2
main.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

test.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

src % clang++ -std=c++20 -Icppfront/include main.cpp test.cpp 
duplicate symbol 'cpp2::to_string(...)' in:
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/main-b0e56a.o
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/test-c3d2ea.o
duplicate symbol 'cpp2::report_and_terminate(std::__1::basic_string_view<char, std::__1::char_traits<char> >, char const*)' in:
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/main-b0e56a.o
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/test-c3d2ea.o
duplicate symbol 'cpp2::fopen(char const*, char const*)' in:
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/main-b0e56a.o
    /var/folders/52/hk4pdfpj0qx83161z_d05ckm0000gp/T/test-c3d2ea.o
duplicate symbol 'cpp2::to_string(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)' in:
[...]

PR #43, opened 2022-10-03
Fix duplicate symbols for functions in cpp2util.h

filipsajdak:
`cpp2::report_and_terminate`, `cpp2::to_string`, and `cpp2::fopen` duplicate symbols and compilation fails on linking.

This change makes them `inline`.

closes https://github.com/hsutter/cppfront/issues/42

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -249,7 +249,7 @@ private:
     handler reporter;
 };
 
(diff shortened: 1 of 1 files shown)


JohelEGP:
That fixes it: https://godbolt.org/z/aWvc57G8W.

On #42, you mentioned
> I need to check if there are no other duplicated symbols excluded due to preprocessor rules.

Perhaps you could use Compiler Explorer to investigate that.

hsutter:
Looks good, thanks.

PR #46, opened 2022-10-04
[REFACTOR] Change CPP2_UFCS macro to work for both cases - no need for CPP2_UFCS_0

filipsajdak:
The current implementation has two macros `CPP2_UFCS` and `CPP2_UFCS_0`. One for functions/methods with arguments and the other for no arguments cases. There is a special macro `__VA_OPT__()` that includes its argument only if `__VA_ARGS__` has more than zero arguments.

In this change `CPP2_UFCS_0` is no longer needed and is removed. `CPP2_UFCS` works for both cases.

pure2-stdio.cpp2
```cpp

//  "A better C than C" ... ?
//
main: () -> int = {
    s: std::string = "Fred";
    myfile := fopen("xyzzy", "w");
    myfile.fprintf( "Hello %s with UFCS!", s.c_str() );
    myfile.fclose();
}

```
generates
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 4 "external/cppfront/regression-tests/pure2-stdio.cpp2"
#line 10 "external/cppfront/regression-tests/pure2-stdio.cpp2"


#line 1 "external/cppfront/regression-tests/pure2-stdio.cpp2"

//  "A better C than C" ... ?
//
    std::string s { "Fred" }; 
    auto myfile { fopen("xyzzy", "w") }; 
    CPP2_UFCS(fprintf, myfile, "Hello %s with UFCS!", CPP2_UFCS(c_str, s));
    CPP2_UFCS(fclose, myfile);
}
```
compiles and run fine.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -459,22 +459,13 @@ public:
 //-----------------------------------------------------------------------
 //
 #define CPP2_UFCS(FUNCNAME,PARAM1,...) \
-[](auto&& obj, auto&& ...params) { \
-    if constexpr (requires{ std::forward<decltype(obj)>(obj).FUNCNAME(std::forward<decltype(params)>(params)...); }) { \
-        return std::forward<decltype(obj)>(obj).FUNCNAME(std::forward<decltype(params)>(params)...); \
+[](auto&& obj __VA_OPT__(, auto&& ...params)) { \
+    if constexpr (requires{ std::forward<decltype(obj)>(obj).FUNCNAME( __VA_OPT__(std::forward<decltype(params)>(params)...) ); }) { \
+        return std::forward<decltype(obj)>(obj).FUNCNAME( __VA_OPT__(std::forward<decltype(params)>(params)...) ); \
     } else { \
-        return FUNCNAME(std::forward<decltype(obj)>(obj), std::forward<decltype(params)>(params)...); \
+        return FUNCNAME(std::forward<decltype(obj)>(obj) __VA_OPT__(, std::forward<decltype(params)>(params)...) ); \
     } \
-}(PARAM1, __VA_ARGS__)
-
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks! This works on GCC and Clang. However, MSVC support for `__VA_OPT__` still requires the `/Zc:preprocessor` switch which puts the compiler into an experimental mode that emits noisy disclaimers. I'm trying to keep my build instructions clean and not rely on experimental switches.

That said, I'll do a separate commit to put this into `cpp2util.h` and `cppfront.cpp` with a note to enable it as soon as MSVC supports it without experimental extensions. Again, thanks!

gregmarr:
Interesting, https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-170 makes it sound like `/Zc:preprocessor` is fully supported as of Visual Studio 2019 version 16.5 when it replaced the earlier /experimental:preprocessor switch.

filipsajdak:
@hsutter happy to help. If you have any work that you would like to have done please let me know. Currently I rely on the TODO comments.

I am willing to help and can try to work on the things that matters at the moment. I am still learning the structure of the cppfront but can handle some of the tasks right a way.

hsutter:
@filipsajdak Ah, good thing you mentioned the TODOs... Please don't worry about the TODO notes, those are mostly just reminders to myself, as I leave some placeholders for things I deliberately leave in a partial state that can be polished later, so that I can make progress on other features/work instead. As reminders to myself they're also often incomplete, because I have other context in my head about things I intend to do in that place when I come back to it.

What would be useful is bug reports for problems you find as you try the compiler (thanks for those!) and missing features that you wanted yourself (such as interpolation for basic `std` types like `pair` and `optional`). Please keep those comings. Thanks!

PR #47, opened 2022-10-04
[REFACTOR] Remove the duplication for cpp2::lexeme enum

filipsajdak:
In `lex.h` file there was a comment
```cpp
//  TODO: Remove the duplication above/below with a variadic macro,
//        but for now it's simpler just to write it out
```
In the proposed change I have introduced variadic macro that creates:
* `enum class lexeme` (with all enum values),
* implementation of `cpp2::as(lexeme) -> std::string` for all enum values,

`EXPAND()` macro scans the content 1024 times - it means that 1024 enum values are supported.

All macros are removed with `#undef` after use.

Working implementation: https://godbolt.org/z/fd7bc4rMx

### DIFF
--- a/source/lex.h
+++ b/source/lex.h
@@ -32,7 +32,34 @@ namespace cpp2 {
 //-----------------------------------------------------------------------
 //
 
-enum class lexeme : std::int8_t {
+#define CPP2_ADD_PARENS ()
+
+#define CPP2_SWITCHCASE(N) break; case N: return #N;
+
+// CPP2_EXPAND scans macro 1024 times that allows to create a enum that contains 1024 values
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks, and now I'm torn... I did have the TODO to explore generating these, but this version seems to make the code less readable than the existing duplication (which also allows me to comment out specific cases I plan to do something special with later).

Sorry but for now I think I'll stick with the existing code despite the duplication, and maybe once cppfront gets reflection/generation and can be self-hosted I can generate these without macros.

Thanks for trying it out though!

filipsajdak:
I don't like these macros either. Don't worry about it. 

Working on it I have learned about `__VA_OPT__` that makes me propose simplification for `CPP2_UFCS`. No time waste.

Nevertheless, I am helping not to have to write such ugly code in the new syntax. Happy to help!

filipsajdak:
@hsutter I just came up with some ideas on how we can generate these enums better. The implementation uses the creation of an enum, `constexpr` string parsing, a list of disabled enums, and the creation of a `to_string()` function that searches for the proper `string_view` to use;

Working implementation is here: https://godbolt.org/z/9db79G1EP

And you can use it like this
```cpp
CPP2_MAKE_LEXEME_ENUM_WITH_DISABLED((
    SlashEq,
    Slash,
    LeftShiftEq,
    RightShift,
    GreaterEq,
    Greater,
), 
/*DISABLE to_string() for:*/ 
    RightShift,
);

int main() {
    std::cout << to_string(lexeme::LeftShiftEq) << std::endl; // -> LeftShiftEq
    std::cout << to_string(lexeme::RightShift) << std::endl; // you can disable specific enum -> INTERNAL_ERROR
    std::cout << to_string(static_cast<lexeme>(10)) << std::endl; //  -> INTERNAL_ERROR
    std::cout << to_string(lexeme::Greater) << std::endl; // -> Greater
}
```

The current implementation uses `__VA_OPT__` but I can work on preparing something that is not using it. Just let me know if I should prepare the pull request for it.

filipsajdak:
OK, here is a version without `__VA_OPT__` -> https://godbolt.org/z/KnY3ev96x

Issue #56, opened 2022-10-08
[BUG] Correct #line directives to point to cpp2 source when output option is used


filipsajdak:
After introducing `output` option the `#line` directives points to the output file path instead of cpp2 source path.

### Cpp2 source file
```cpp
i : int = 42;
```
### Steps
```
1. build/external/cppfront external/tests/easy.cpp2 -o build/easy.cpp
```

`build/easy.cpp` content
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 1 "build/easy.cpp2"

//=== Cpp2 definitions ==========================================================

#line 1 "build/easy.cpp2"
    int i { 42 }; 
```

### Error & expected behaviour
`#line` directive points to file `build/easy.cpp2` that does not exists. The expected path is `external/tests/easy.cpp2`

Expected file content:
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 1 "external/tests/easy.cpp2"

//=== Cpp2 definitions ==========================================================

#line 1 "external/tests/easy.cpp2"
    int i { 42 }; 
```

PR #57, opened 2022-10-08
Correct #line directive to cpp2 source file

filipsajdak:
Closes https://github.com/hsutter/cppfront/issues/56

After introducing `output` option the `#line` directives points to the output file path instead of cpp2 source path.

### Cpp2 source file
```cpp
i : int = 42;
```
### Steps
```
1. build/external/cppfront external/tests/easy.cpp2 -o build/easy.cpp
```

`build/easy.cpp` content
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 1 "build/easy.cpp2"

#line 1 "build/easy.cpp2"
    int i { 42 }; 
```

### Error & expected behaviour
`#line` directive points to file `build/easy.cpp2` that does not exists. The expected path is `external/tests/easy.cpp2`

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -121,6 +121,7 @@ class positional_printer
 {
     //  Core information
     std::ofstream               out       = {};     // Cpp1 syntax output file
+    std::string                 cpp2source = {};
     std::string                 filename  = {};
     std::vector<comment> const* pcomments = {};     // Cpp2 comments data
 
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks! I saw this issue via email and did the fix, then noticed this was a PR and your fix was essentially identical to what I did. :) But please check, thanks.

Issue #45, opened 2022-10-04
[SUGGESTION] in cpp2 mode, string literals should be string_literal by default


neumannt:
Raw pointers are fundamentally problematic if we have to do pointer arithmetic or offset based access. A major source for pointers, that we inherited from C, are strings. There are safer alternatives to C strings (i.e., string_view), but they are not used by default.
We still need classic C strings for compatibility reasons, but the default should be a safe construct.

Thus, I would suggest that a string literal in cpp2 mode is considered a string_view by default. Traditional C strings could be constructed by using, e.g., the c suffix:

cpp2: "abc" -> cpp: "abc"sv
cpp2: "abc"c -> cpp: "abc"

That would make string handling much cleaner and safer. Compatibility with existing code is a concern, but as cpp2 code is by definition new anyway, we can always add the 'c' suffix if needed to get a traditional C string.

marioarbras:
I think this is a great idea, but I propose a slightly different approach. Instead of having string literals be `string_view`'s in Cpp2 syntax, I propose creating a new string literal type.

The reason for this is because, in general, `string_view` does not represent a null-terminated string. It will if you initialize it with a C string literal, but it's not something that's true for all use cases, and might introduce bugs if it gets misused.

Consider how you would interface with a C library assuming Cpp2 string literals are `string_view`'s:
```cpp
str:= "a Cpp2 string literal"; // actually a std::string_view

some_c_function(str.data()); // OK, str.data() is null-terminated
```

You have no option but to call the member function `.data()` to get a `const char*` out. A user might be tempted to do the same thing when passing a `string_view` that was not initialized from a string literal:
```cpp
str: std::string = "a regular Cpp string";
substr: std::string_view = (str.begin(), std::next(str.begin(), 3));

some_c_function(substr.data()); // ¯\_(ツ)_/¯
```

I suggest implementing strongly-typed string literals in Cpp2. This type could have an implicit conversion operator to `string_view` and should only have a constructor from a string literal. This type can also have a `.c_str()` member function, just like `std::string`, that returns a `const char*` guaranteed to be null-terminated. I've seen some implementations of such types, sometimes called `fixed_string` or `string_literal`. Internally, this type is just a wrapper around a reference to the string literal (guaranteed to be valid for the duration of the program) with a compile-time known size and so does not add any overhead.

The code above then becomes
```cpp
str:= "a Cpp2 string literal"; // actually a cpp2::string_literal
[...]

PR #63, opened 2022-10-09
Improves main function

mordante:
Two patches:
1. Removes the forward declaration of `main` and its `[[nodiscard]]`.
2. Allows `main` to return `void` and allows a `std::span<std::string_view>` argument instead of `int, char**`.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -199,6 +199,9 @@
     #include <cstddef>
     #include <utility>
     #include <cstdio>
+    #include <vector>
+    #include <string_view>
+    #include <span>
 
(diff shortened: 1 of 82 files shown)


hsutter:
Thanks for the suggestion. I considered something like this (and am still considering it) but there are subtle issues... for example, using a `span<string_view>` breaks the use of the command-line arguments as null-terminated strings. See also [this comment on issue 45](https://github.com/hsutter/cppfront/issues/45#issuecomment-1267800958) for discussion on safety of string literals in Cpp2... command-line arguments are a lot like string literals in this regard.

Issue #66, opened 2022-10-10
[BUG] parens are not preserved when used for calculating argument value


filipsajdak:
Parens disappear when used inside function call to calculate value of argument.

Given cpp2 code:
```cpp
f: (x:_) = {}

main: () -> int = {
    i := 2;
    f((i+(i+1)*2)/2);
}
```
When compiled with cppfront we get (I am skipping the boilerplate)
```cpp
[[nodiscard]] auto main() -> int{
    auto i { 2 }; 
    f( i + i + 1 * 2 / 2);
}
```
all parens inside the function call are gone. `(i+(i+1)*2)/2` is not equal `i + i + 1 * 2 / 2`.

The expectation is that all parens used to calculate arguments will last cppfront compilation.

PR #67, opened 2022-10-11
[FIX] preserve parens around expressions calculating value of argument for a function call

filipsajdak:
The current implementation removes parentheses used in expression that calculates the value of arguments for a function.

cpp2 code:
```cpp
f: (x:_) = {}

main: () -> int = {
    i := 2;
    f((i+(i+1)*2)/2);
}
```
When compiled with cppfront we get (I am skipping the boilerplate)
```cpp
    auto i { 2 }; 
    f( i + i + 1 * 2 / 2);
}
```
All parentheses inside the function call are gone. `(i+(i+1)*2)/2` is not equal `i + i + 1 * 2 / 2`.

This fix corrects that behavior and preserves these parentheses. The generated code after the fix:
```cpp
    auto i { 2 }; 
    f((i + (i + 1) * 2) / 2);
}

```

Close https://github.com/hsutter/cppfront/issues/66

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1980,7 +1980,9 @@ public:
     auto emit(expression_node const& n) -> void
     {
         assert(n.expr);
+        push_need_expression_list_parens(true);
         emit(*n.expr);
+        pop_need_expression_list_parens();
     }
 
 
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1990,7 +1990,7 @@ public:
     //
     auto emit(expression_list_node const& n) -> void
     {
-        if (should_add_expression_list_parens() && !n.expressions.empty()) {
+        if (should_add_expression_list_parens() && !n.expressions.empty() && !n.inside_initializer) {
             printer.print_cpp2("(", n.position());
         }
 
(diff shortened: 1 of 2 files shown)


filipsajdak:
This fix breaks 
```cpp
v : std::vector<int> = ( 1, 2, 3 );
```
cppfront generates
```cpp
std::vector<int> v { (1, 2, 3) };
```
which compiles but thanks to the coma operator it will compile by cpp1 compiler to a `vector` with only `3` value.

filipsajdak:
OK, above issue is now fixed.

filipsajdak:
Some issue still exists

cpp2 code
```cpp
ddd := fun(fun("abc").substr(1));
```
compiles to
```cpp
auto ddd { fun(CPP2_UFCS(substr, fun(("abc")), 1)) };
```

filipsajdak:
The issue with additional parens around function arguments is solved.

The issue showed up in the code merged with https://github.com/hsutter/cppfront/pull/18 (I was not able to trigger it on the main branch). I remove that commit from the patch set.

hsutter:
Good fix -- this actually corrects the out for two existing test cases, one of which is `mixed-test-parens.cpp2` that I hadn't noticed had the wrong output. As Bjarne would say: "Eek!" Thanks.

Issue #62, opened 2022-10-09
[BUG] In mixed mode cppfront can interfere with string done with raw string literal


filipsajdak:
Cppfront recognized a line as a cpp2 line but it is in the middle of a raw string literal.

Cpp2 code:
```cpp
auto a = R"(
i:int = 42;
)";
```
after compiling with cppfront we get:
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"

auto a = R"(

)";

//=== Cpp2 definitions ==========================================================

int i { 42 }; 
```
https://godbolt.org/z/qMnerz8cM

Expected: cppfront should treat raw string literals similarly as it treats `/* */` comments - there is a need to check if it is in the middle of a raw string literal (like we test if it is in the middle of the comment).

PR #69, opened 2022-10-11
[SUGGESTION][FIX] Add support for raw string literals in mixed mode

filipsajdak:
Current implementation in mixed mode is not parsing raw string literals well.

Cpp2 source code
```cpp
auto r = R"(
    this is raw string literal
    \n
    /*   */
    //
i: int = 42;
)";

/*
 *
 *   this is a comment
 *
 */

main: () -> int = {
    std::cout << "Hello world\n\n" << std::quoted(r) << std::endl;
}
```
Compiles to
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"

#line 1 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

auto r = R"(
    this is raw string literal
    \n
    /*   */
    //

#line 7 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"
#line 8 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"
)";

#line 16 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

#line 7 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"
int i { 42 }; 
#line 9 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

/*
 
    this is a comment
 
 */

    std::cout << "Hello world\n\n" << std::quoted(r) << std::endl;
}
```

The proposed change makes cppfront parse raw string literals in a similar way as it parses `/* */` comments. After this change cppfront produces:
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"

#line 1 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

auto r = R"(
    this is raw string literal
    \n
    /*   */
    //
i: int = 42;
)";

#line 16 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

#line 9 "/Users/filipsajdak/dev/execspec/external/tests/raw_string.cpp2"

/*
 
    this is a comment
 
 */

    std::cout << "Hello world\n\n" << std::quoted(r) << std::endl;
}
```
Debug info shows that the lines are recognized as `/* R */` which was added symbol to mark raw string literals lines.
```
/*   */ 
/* 1 */ auto r = R"(
/* R */     this is raw string literal
/* R */     \n
/* R */     /*   */
/* R */     //
/* R */ i: int = 42;
/* 1 */ )";
/* 2 */ 
/* 2 */ /*
/* 2 */  *
/* 2 */  *   this is a comment
/* 2 */  *
/* 2 */  */
/* 2 */ 
/* 2 */ main: () -> int = {
/* 2 */     std::cout << "Hello world\n\n" << std::quoted(r) << std::endl;
/* 2 */ }
```

Close https://github.com/hsutter/cppfront/issues/62

### DIFF
--- a/source/common.h
+++ b/source/common.h
@@ -46,7 +46,7 @@ struct source_line
 {
     std::string text;
 
-    enum class category { empty, preprocessor, comment, import, cpp1, cpp2 }
+    enum class category { empty, preprocessor, comment, import, cpp1, cpp2, rawstring }
         cat = category::empty;
 
     auto prefix() const -> std::string
@@ -58,6 +58,7 @@ struct source_line
         break;case category::import:        return "/* i */ ";
         break;case category::cpp1:          return "/* 1 */ ";
         break;case category::cpp2:          return "/* 2 */ ";
+        break;case category::rawstring:     return "/* R */ ";
         break;default: assert(!"illegal category"); abort();
         }
     }
--- a/source/load.h
+++ b/source/load.h
@@ -196,27 +196,29 @@ auto starts_with_identifier_colon(std::string const& line) -> bool
 struct process_line_ret {
     bool all_comment_line;
     bool empty_line;
+    bool all_rawstring_line;
 };
 auto process_cpp_line(
     std::string const&  line,
     bool&               in_comment,
     bool&               in_string_literal,
+    bool&               in_raw_string_literal,
     std::vector<int>&   brace_depth,
     lineno_t            lineno,
     std::vector<error>& errors
 )
     -> process_line_ret
 {
-    if (!in_comment && !in_string_literal && starts_with_whitespace_slash_slash(line)) {
-        return { true, false };
+    if (!in_comment && !in_string_literal && !in_raw_string_literal && starts_with_whitespace_slash_slash(line)) {
+        return { true, false, false };
     }
 
-    if (!in_comment && !in_string_literal && starts_with_whitespace_slash_star_and_no_star_slash(line)) {
+    if (!in_comment && !in_string_literal && !in_raw_string_literal && starts_with_whitespace_slash_star_and_no_star_slash(line)) {
         in_comment = true;
-        return { true, false };
+        return { true, false, false };
     }
 
-    struct process_line_ret r { in_comment, true };
+    struct process_line_ret r { in_comment, true , in_raw_string_literal};
 
     auto prev = ' ';
     for (auto i = colno_t{0}; i < ssize(line); ++i)
@@ -225,7 +227,7 @@ auto process_cpp_line(
         //  Note: in_literal is for { and } and so doesn't have to work for escaped ' characters
         //
         auto peek       = [&](int num) {  return (i+num < std::ssize(line)) ? line[i+num] : '\0';  };
-        auto in_literal = [&]()        {  return in_string_literal || (prev == '\'' && peek(1) == '\'');  };
+        auto in_literal = [&]()        {  return in_string_literal || in_raw_string_literal || (prev == '\'' && peek(1) == '\'');  };
 
         //  Process this source character
         //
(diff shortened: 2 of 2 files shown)


filipsajdak:
In the last push, I have added also support for `d-char-sequence` from spec:
```
prefix(optional) R"d-char-sequence(optional)
(r-char-sequence(optional))d-char-sequence(optional)"
```
The only thing I did not check yet is the prefix.

The current implementation can handle:
```cpp
auto r = R"abc(
    this is raw string literal
    \n
    /*   */
    //
i: int = 42;
)abc";
```

filipsajdak:
I have tested the prefixes and it seems to work, too.

hsutter:
I actually completely forgot about raw string literals... thanks for doing this. At first blush this looks good, before merging let me take a deeper look with a fresh brain and a fresh cup of coffee...

Did you check the three `/passthrough-tests` files to make sure they still are diff-identical when passed through cppfront?

filipsajdak:
No, unfortunately I did not do this. 

Will check that.

JohelEGP:
I think https://github.com/modern-cmake/cppfront does that.

filipsajdak:
I have rebased my changes to the main and done the passthrough tests.

clang & GCC files are the same.

MSVC ends with the error:
```
cppfront % ../../build/external/cppfront passthrough-tests/msvc-msstl-e.cpp2      
passthrough-tests/msvc-msstl-e.cpp2...
msvc-msstl-e.cpp2(174324,0): error: closing } does not match a prior {
```

Need to check what caused the error.

filipsajdak:
OK, found a bug and fix it.

The issue was here
```diff
diff --git a/source/load.h b/source/load.h
index 71fd857..9954a02 100644
--- a/source/load.h
+++ b/source/load.h
@@ -248,7 +248,7 @@ auto process_cpp_line(
                 return r;
             }
             in_raw_string_literal = false;
-            i += end_pos+d_char_sequence.size()-1;
+            i = end_pos+d_char_sequence.size()-1;
         }
         else {
             r.all_comment_line = false;
```

Small but nasty mistake. Now all files compiles and generate identical output.

filipsajdak:
Rebased to main - all passthrough-tests and regression-tests passed.

hsutter:
I tried this out with the `passthrough-tests`, starting with running `cppfront msvc-msstl-e.cpp2`, but it crashed.

Running it under the debugger, I see that the crash is an out-of-bounds access attempt in the second-last line in `process_cpp_line`:

    prev = line[i];

where `i` was about 20-30 higher than `line.size()`.

I didn't investigate further, but this should give you a repro and a roadmap to find the bug... can you take a look?

filipsajdak:
Hmm... I am surprised that it crashed. I will investigate

filipsajdak:
The interesting fact is that only MSVC is using raw string literals in its lib. It is always inline and finalised with the string_view suffix. E.g.:
```cpp
_Result += R"(: ")"sv;
```
so `d_char_sequence` will be empty.

filipsajdak:
@hsutter I have rebased my changes on the newest main and still cannot reproduce. I will probably need to check on other compilers.

filipsajdak:
@hsutter I am confused. I have tried it on MSVC and it works as well. I will need more information To be able to reproduce it. 

My very first guess is that it looks like the previous bug that I have fixed (check above). Some merge issues might also cause it.

hsutter:
OK, I'll take another look. Thanks!

filipsajdak:
@hsutter if it will not work, please send me the code you have via email I will verify what is wrong.

hsutter:
Good news: It works fine!

The test failure was almost certainly my error when I opened the branch in GitHub Desktop and must just not have had all the latest commits somehow. (I did try pulling the later commits, but there was a conflict and I'm just not familiar enough with GitHub Desktop.)

So I have verified the PR by the time-honored method of manually applying it to my main branch and running the passthrough tests, and they work fine. So now I'll revert that and merge this PR. :) Thanks again.

filipsajdak:
Happy to hear that it's working.

The good thing is that this situation motivated me to run sanitizers which helped in discovering another issue.

Issue #73, opened 2022-10-13
[BUG] Unable to use enums in inspect expressions - `no matching function for call to 'is'`


filipsajdak:
The current implementation is not supporting enums in inspect expressions.

This cpp2 code
```cpp
enum class lexeme : std::uint8_t {
    hash,
};

to_string: (e:lexeme) -> auto = {
    return inspect (e) -> std::string {
        is lexeme::hash = "hash";
        is _ = "INTERNAL_ERROR";
    };
}
```
compiles by cppfront to (skipping enum declaration):
```cpp
[[nodiscard]] auto to_string(cpp2::in<lexeme> e) -> auto{
    return [&] () -> std::string { auto&& __expr = (e);
        if (cpp2::is<lexeme::hash>(__expr)) { if constexpr( requires{"hash";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF("hash"),std::string> ) return "hash"; else return std::string{}; else return std::string{}; }
        else return "INTERNAL_ERROR"; }
    ()
; }
```
and fails to compile with the cpp1 compiler with the error:
```
main.cpp2:26:13: error: no matching function for call to 'is'
[build]         if (cpp2::is<lexeme::hash>(__expr)) { if constexpr( requires{"hash";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF("hash"),std::string> ) return "hash"; else return std::string{}; else return std::string{}; }
[build]             ^~~~~~~~~~~~~~~~~~~~~~
[build] /Users/filipsajdak/dev/execspec/external/cppfront/include/cpp2util.h:517:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'C'
```

The issue is caused that enums cannot match to any provided `is` function overloads. To handle that we need special overload for enums.

hsutter:
This now works with the latest checkin:

```
enum class lexeme : std::uint8_t {
    hash,
};

main: ()->int = {
    e := lexeme::hash;
    std::cout << inspect (e) -> std::string {
        is (lexeme::hash) = "hash";
        is _ = "INTERNAL_ERROR";
    };
}
```

Thanks!

filipsajdak:
@hsutter I am testing the main branch.

Below code
```cpp
#include <variant>
#include <string_view>
#include <iostream>

main: () -> int = {

    v : std::variant<int, double, std::string_view> = 42;

    std::cout << inspect v -> std::string {
        is 42 = "this is 42";
[...]

PR #74, opened 2022-10-13
[FIX] Add missing overload for `is` function to handle enums

filipsajdak:
The current implementation does not have an overload for the `is` function to handle `enums`.

The below code compiles by cppfront but the result code fails to compile with the cpp1 compiler.
```cpp
enum class lexeme : std::uint8_t {
    hash,
};

to_string: (e:lexeme) -> auto = {
    return inspect (e) -> std::string {
        is lexeme::hash = "hash";
        is _ = "INTERNAL_ERROR";
    };
}
```
cppfront result
```cpp
[[nodiscard]] auto to_string(cpp2::in<lexeme> e) -> auto{
    return [&] () -> std::string { auto&& __expr = (e);
        if (cpp2::is<lexeme::hash>(__expr)) { if constexpr( requires{"hash";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF("hash"),std::string> ) return "hash"; else return std::string{}; else return std::string{}; }
        else return "INTERNAL_ERROR"; }
    ()
; }
```

Unfortunately, there is no overload that can match using an enum value. This change provides additional overload that works with enums - it checks if the value is an enum type and if the compared value is the same type as the enum value provided.

Working prototype: https://godbolt.org/z/1c4o8qG6P

Close https://github.com/hsutter/cppfront/issues/73

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -548,6 +548,14 @@ auto is( X const& x ) -> bool {
     return x == X();
 }
 
+template< auto value, typename X >
+auto is( X const& x ) -> bool {
+    if constexpr (std::is_convertible_v<decltype(value),X>) {
+        return x == value;
+    } else {
+        return false;
+    }
+}
 
 //-------------------------------------------------------------------------------------------------------------
 //  Built-in as (partial)


filipsajdak:
I have made a small change to the `is` function overload to work with various of enums. You can now compare values of two different types of enums and have a `false` as a return - it will compile in inspect like below:
```cpp
v := lexeme::hash;

  is lexeme::hash  = "original lexeme";
  is lexeme2::hash = "fake lexeme2";
  is _             = "don't know";
} << std::endl;
```
and will match the proper alternative. Of course, it will also work for the `lexeme2::hash` value.

filipsajdak:
I am closing it as it is also covered by https://github.com/hsutter/cppfront/pull/79

PR #86, opened 2022-10-20
[SUGGESTION] Add inspection of `std::variant` and `std::optional` values

filipsajdak:
Current implementation of `inspect` and `is` allows to inspect types. https://github.com/hsutter/cppfront/pull/79 add inspection capabilities to check values of variables (what allows to replace cpp1 switches). 

This change introduce possibility to inspect values stored in `std::variant` and `std::optional` what make given code work:
```cpp
v : std::variant<lexeme, lexeme2, std::string, int, double, char, my_struct> = lexeme2::hash;

    is lexeme::hash  = "original lexeme";
    is lexeme2::hash = "fake lexeme2";
    is "hash"        = "it is string 'hash'";
    is 42            = "the answer is 42";
    is 1.23          = "the answer is 1.23";
    is 'a'           = "this is 'a'";
    is _             = "don't know";
} << std::endl;
```
or you can use it as if statement with `std::optional`
```cpp
o : std::optional<std::string> = "hash";

if o is "hash" {
    std::cout << "Optional has 'hash' value" << std::endl;
} else {
    std::cout << "There is something else" << std::endl;
}
```
or `std::variant`
```cpp
if v is 1.23 {
  std::cout << "This is 1.23" << std::end;
}
```

This PR is based on https://github.com/hsutter/cppfront/pull/79

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -615,6 +615,94 @@ auto is( X const& x ) -> bool {
     return x == X();
 }
 
+//-------------------------------------------------------------------------------------------------------------
+//  Built-in is (literals)
+//
+
+template< auto value, typename X, typename V = CPP2_TYPEOF(value)>
+    // This requires are needed for gcc compiler to avoid ambiguity when value is double
+    requires (
+        (std::floating_point<V> && !std::floating_point<X>) ||
+        (std::integral<V> && !std::integral<X>) ||
+        (std::is_enum_v<V> && !std::is_same_v<V, X> )
+    )
+constexpr auto is( X const& x ) -> bool {
+    return false;
+}
+
+template< auto value, typename X >
+    requires (std::is_enum_v<CPP2_TYPEOF(value)> && std::is_same_v<X,CPP2_TYPEOF(value)>)
+constexpr auto is( X const& x ) -> bool {
+    return x == value;
+}
+
+template< auto value, typename X >
+    requires std::integral<CPP2_TYPEOF(value)> && std::integral<X>
+constexpr auto is( X const& x ) -> bool {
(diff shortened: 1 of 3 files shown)


filipsajdak:
Rebase to main. Make `is()` functions `constexpr`

filipsajdak:
Might require https://github.com/hsutter/cppfront/pull/90 for some use cases.

hsutter:
> Might require https://github.com/hsutter/cppfront/pull/90 for some use cases.

Should these be merged? Should all the `is`/`as` PRs be merged (also #79, #106, #108)? (Not suggesting, just asking. As I look through them I'm trying to grok the overlap.)

fecjanky:
Conceptually can we make `std::optional<T>` to be isomorphic to `std::variant<std::monostate,T>` from the cpp2 language/library point of view as these types are isomorphic mathematically as well? This would definitely simplify things on multiple levels.

hsutter:
Thanks! About half of this (parsing and literals) seems to be duplicated from #79, so I'll focus on the `variant`, `optional`, and `any` parts... more soon...

hsutter:
OK, half of this which wasn't about `variant`/`optional`/`any` was merged in the commit c77de70270ff35fdb7484adcadc1c4b66c461d41 that closed #79, and I've now applied the rest for those three standard types in a separate commit to follow the pattern of commit c77de70270ff35fdb7484adcadc1c4b66c461d41 to use the two-parameter "is value" pattern.

Thanks again!

hsutter:
I've now added this with 2f37277d3d4844ef5a904213409ac4c657f6a858 - thanks again.

Issue #91, opened 2022-10-23
[BUG] Parens disappear in `std::cout` expressions


filipsajdak:
Parens are not preserved in one of the regression-tests: `mixed-test-parens.cpp2`

The code
```cpp
std::cout << (1+2)*(3+v[0]);
```
is compiled by the cppfront to:
```cpp
std::cout << (1+2)* 3 + v[0];
```

filipsajdak:
Fixed in https://github.com/hsutter/cppfront/pull/92

Issue #81, opened 2022-10-18
[BUG] Expected outputs in tests need to be updated


alexreinking:
**Describe the bug**
Two expected outputs are stale
```
mixed-function-expression-with-pointer-capture
mixed-test-parens
```

**To Reproduce**
See the failing workflow run here.

https://github.com/modern-cmake/cppfront/pull/9

**Additional context**
The workflow doesn't show the diff of the failure, yet, unfortunately.

alexreinking:
Fixed in latest commits... modern-cmake/cppfront#10 passing regression tests.

Issue #72, opened 2022-10-13
[BUG] Lambda that use capture `i$` with lambda argument `n` inside if statement failed with assertion


filipsajdak:
Using capture and if failed if done in some order.

In this cpp2 code cppfront fail on `if i$ + n < std::ssize(line$)`:
```cpp
lex_process_line: (line: std::string) = {
    i := 0;
    peek := :(n : int) -> char = {
        if i$ + n < std::ssize(line$) {} // it triggers assertion
        return '\0';
    };
}
```
https://godbolt.org/z/bTvzW4P6z

On Compiler Explorer it fails with the error:
```
cppfront: source/cppfront.cpp:1561: void cpp2::cppfront::emit(cpp2::postfix_expression_node&, bool): Assertion `n.expr' failed.
Compiler returned: 139
```
On my setup, it fails with the error:
```
Assertion failed: (cap.capture_expr->cap_grp == &captures), function build_capture_lambda_intro_for, file cppfront.cpp, line 1495.
```

The issue can be solved by changing the order of variables in the if statement:
```cpp
lex_process_line: (line: std::string) = {
    i := 0;
    peek := :(n : int) -> char = {
        if n + i$ < std::ssize(line$) {} // works
        return '\0';
    };
}
```
https://godbolt.org/z/h8PGEs39K

by using parentheses:
```cpp
lex_process_line: (line: std::string) = {
    i := 0;
    peek := :(n : int) -> char = {
        if (i$ + n) < std::ssize(line$) {} // works
        return '\0';
    };
}
```
https://godbolt.org/z/4v7YxevY7

Replacing `n` with the constant also works:
```cpp
lex_process_line: (line: std::string) = {
    i := 0;
    peek := :(n : int) -> char = {
        if i$ + 2 < std::ssize(line$) {} // works
        return '\0';
    };
}
```
https://godbolt.org/z/hcefPbers

My expectation is that cppfront does not crash. It should work with various orders of variables or should show the error message if there is some preferred order.

hsutter:
Thanks! Found it... when first attempting to parse `n < line$` we first try to parse it as a template-argument-list, then correctly realize it isn't that and back out, but the backing-out cleanup didn't unregister the capture. Now it will, in the next commit.

filipsajdak:
Superb!

hsutter:
[...]

PR #92, opened 2022-10-23
[FIX] preserving parens for std::cout expressions

filipsajdak:
There is a bug in expressions inside `std::cout` shift expressions. The example is in `regression-tests/mixed-test-parens.cpp2`

The code:
```cpp
std::cout << (1+2)*(3+v[0]);
```
currently compiles to
```cpp
std::cout << (1+2)* 3+v[0];
```

After this change the cppfront preserves the parens and returns:
```cpp
std::cout << (1+2)*(3+v[0]);
```

This PR closes https://github.com/hsutter/cppfront/issues/91

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1857,7 +1857,9 @@ public:
             }
         }
         assert(n.expr);
+        push_need_expression_list_parens(true);
         emit(*n.expr);
+        pop_need_expression_list_parens();
         printer.print_cpp2(suffix, n.position());
     }
 


filipsajdak:
Probably we need to check if there is more expressions that might suffer from disappearing parens.

hsutter:
Thanks, this looks reasonable. I'm slightly confused because I remember that these parens were correctly emitted (for the first time) with a commit about a week ago. Curious what caused the regression?

filipsajdak:
@hsutter it looks like this test case was broken by my previous PR about parens https://github.com/hsutter/cppfront/pull/67 (that one you refers to). It looks like everything works fine before it. Sorry, my mistake I should have run more regression tests. I already do it before sending PRs (that is why I found this issue).

Previous PR fixes different scenario:
```cpp
f((i+(i+1)*2)/2);
```
I will investigate it a little more later.

JohelEGP:
The issue had been known since https://github.com/hsutter/cppfront/issues/81. From the [CI run](https://github.com/modern-cmake/cppfront/actions/runs/3270631069/jobs/5379479602#step:11:179):
> ```
>        Start  72: codegen/check/mixed-test-parens
> 85/118 Test  #72: codegen/check/mixed-test-parens ................................................***Failed    0.00 sec
> ```

Issue #78, opened 2022-10-17
[BUG] Add support for defining pointer to pointers. 


leejy12:
Hi, I am very interested in this project's future!

Cppfront fails to compile this code
```cpp
main: () -> int = {
    a:     int = 2;
    pa:   *int = a&;
    ppa: **int = pa&;
    return a*pa**ppa**; // 8
}
```

As of now, `ppa: **int = pa&;` does not work.

filipsajdak:
That is true. As a workaround you can use type aliases.

hsutter:
@leejy12 Thanks, fixed with commit e1b57cf026041cef33d4f95ae230513ada29fa51 .

This now works (note spacing around `*`, otherwise it's your example):

```
main: () -> int = {
    a:     int = 2;
    pa:   *int = a&;
    ppa: **int = pa&;
    std::cout << a * pa* * ppa**; // 8
}
```

Thanks!

PR #93, opened 2022-10-24
[FIX] Add support for multi-level pointers

filipsajdak:
Current implementation supports only pointer of the form
```cpp
p: *int = ...;
```

This change extends that to allow multi-level pointers and make below code compilable
```cpp
main: (argc : int, argv : **char) -> int = {
    a:     int = 2;
    pa:   *int = a&;
    ppa: **int = pa&;

    return a*pa**ppa**; // 8
}
```
and generates the cpp1 code
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 1 "tests/pointer-to-pointer.cpp2"
[[nodiscard]] auto main(cpp2::in<int> argc, cpp2::in<char * *> argv) -> int;

#line 1 "tests/pointer-to-pointer.cpp2"
[[nodiscard]] auto main(cpp2::in<int> argc, cpp2::in<char * *> argv) -> int{
    int a { 2 }; 
    int* pa { &a }; 
    int** ppa { &pa }; 

    return a * *pa * **ppa; // 8
}
```

Closes https://github.com/hsutter/cppfront/issues/78 and makes possible to implement correct main function with arguments.

# Open issues

This change still does not support deduced pointer types (just like the current implementation). That means:
```cpp
pa = 0;
```
will cause an error as cppfront knows that pa is a pointer to pointer
```
tests/pointer-to-pointer.cpp2...
pointer-to-pointer.cpp2(6,8): error: = - pointer assignment from null or integer is illegal
  ==> program violates lifetime safety guarantee - see previous errors
```
but this code:
```cpp
pa2 := ppa*;
pa2 = 0;

pa3 := a&;
pa3 = 0;
```
will pass lifetime safety guarantee checks and will compile to cpp1 code:
```cpp
auto pa2 { *ppa }; 
pa2 = 0;

auto pa3 { &a }; 
pa3 = 0;
```

Having multi-level pointer makes that issue more important to be fixed as it is easier to deduce a pointer from pointer to pointer.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1071,6 +1071,10 @@ public:
         assert(n.identifier);
         emit(*n.identifier);
 
+        for (auto _ : n.pointer_declarators) {
+            printer.print_cpp2("*", n.position());
+        }
+
         if (!n.template_args.empty()) {
             printer.print_cpp2("<", n.open_angle);
             auto first = true;
@@ -1572,9 +1576,19 @@ public:
                 auto& unqual = std::get<id_expression_node::unqualified>(id->id);
                 assert(unqual);
                 auto decl = sema.get_local_declaration_of(*unqual->identifier);
-                //  TODO: Generalize this -- for now we detect only cases of the form "p: *int = ...;"
-                //        We don't recognize pointer types that are deduced, multi-level, or from Cpp1
-                if (decl && decl->declaration && decl->declaration->pointer_declarator) {
+
+                bool is_pointer = false;
+                if (decl && decl->declaration) {
+                    if (auto* obj_id_expr = std::get_if<declaration_node::object>(&decl->declaration->type)) {
+                        if (auto* unqual = std::get_if<id_expression_node::unqualified>(&(*obj_id_expr)->id)){
+                            is_pointer = !(*unqual)->pointer_declarators.empty();
+                        }
+                    }
+                }
+
+                //  TODO: Generalize this -- for now we detect only multi-level cases of the form "p: ***int = ...;"
+                //        We don't recognize pointer types that are deduced or from Cpp1
+                if (!unqual->pointer_declarators.empty() || is_pointer) {
                     if (n.ops.empty()) {
                         last_postfix_expr_was_pointer = true;
                     }
@@ -2155,10 +2169,6 @@ public:
         }
         else {
             emit( id_expr );
-            if (n.declaration->pointer_declarator) {
-                printer.print_cpp2(" ", n.declaration->pointer_declarator->position());
-                emit(*n.declaration->pointer_declarator);
-            }
         }
 
         //  Then any suffix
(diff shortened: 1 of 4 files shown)


source/cppfront.cpp
```diff
@@ -2538,7 +2538,7 @@ class cppfront
                 }
                 printer.preempt_position(n.position());
                 emit( *type );
-                if (n.pointer_declarator) {
+                for (int i = 0; i < std::ssize(n.pointer_declarators); ++i) {
```
gregmarr, line 2541: Seems like a range based for would be better here so there are no index type issues.

filipsajdak: Actually, my first implementation was done with range-for 
```cpp
for (auto _ : n.pointer_declarators) { /**/}
```
but I was not using the argument so I decided to switch to this one.

JohelEGP: Maybe the `printer` is missing this functionality, or a `repeat` function is missing.

filipsajdak: I am trying to write consistent code with @hsutter style - there is a code with indexes so I am coping the style
```cpp
for(auto i = 0;
    i < std::ssize(prev_line_info.requests) && prev_line_info.requests[i].requested <= pos.colno;
    ++i
    )
```

but I am open to change that if needed.

filipsajdak:
On the second thought, I am also not using index so nevertheless the range-for is better
```cpp
for (auto _ : n.pointer_declarators) {
    printer.print_cpp2("*", n.position());
}
```

I was thinking of using `std::ignore` but it is not valid in this context. I pushed the corrected version.

filipsajdak:
I just spend couple of minutes meditating @hsutter comment (that I have modified for this change)
```cpp
//  TODO: Generalize this -- for now we detect only multi-level cases of the form "p: ***int = ...;"
//        We don't recognize pointer types that are deduced or from Cpp1
```
and I did some tests that makes me add `Open points` to the description of this PR as this needs to be clear that this PR introduce support for multi-level pointers but does not address otter issues. One of them: deduced pointers becomes more an issue after this PR as it will be easier to create a deduced pointer from pointer to pointer.

filipsajdak:
Handling of pointer deduced types are done here: https://github.com/hsutter/cppfront/pull/94

filipsajdak:
Pointers from cpp1 are handled here https://github.com/hsutter/cppfront/pull/96

filipsajdak:
Add support for pointers as template arguments
```cpp
s : std::span<*char> = (argv, gsl::narrow_cast<std::size_t>(argc)); // works
```

hsutter:
@filipsajdak Would you be interested in updating the good/current parts of #93, #94, and #96 and merging them into a new PR to replace these?

filipsajdak:
Yes. I will.

filipsajdak:
Replaced by https://github.com/hsutter/cppfront/pull/196

PR #101, opened 2022-10-28
Minor corrections in terms of readability and performance

MarchMore:
Fixed implicit casts to bool.
Replaced "auto" with "auto*" for readability.
Added "noexcept" to move constructor and move assignment operator at c_raii class.

Signed-off-by: MarchMore <59848602+MarchMore@users.noreply.github.com>

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -291,7 +291,7 @@ auto inline Testing = contract_group(
 );
 
 constexpr auto contract_group::set_handler(handler h) -> handler {
-    Default.expects(h != nullptr);
+    Default.expects(h);
     auto old = reporter;
     reporter = h;
     return old;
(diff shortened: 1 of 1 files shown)


JohelEGP (commented):
For the `nullptr` tests, see [ES.87: Don’t add redundant == or != to conditions](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions). I personally break it for numbers, but not for pointers or types contextually convertible to `bool`.

MarchMore:
I wasn't aware of this one. Reverted corresponding code.

hsutter:
Thanks! I like the two `noexcept`s and will add those separately in another commit I'm doing, should land in the next push.

PR #102, opened 2022-10-29
[FIX] Correct get_declaration_of not to skip first symbol

filipsajdak:
While looking for the symbol there is a loop that searches backward using normal iterators. The loop will never check symbols.cbegin() as it is excluded by `i != symbols.cbegin()` condition.

That makes in the below code we will never find the `fun` symbol
```cpp
fun: (inout i : int) -> * const int = {
    return i&;
}

main: () -> int = {
    i := 42;
    p := fun(i);
    // ...
}
```

This change replaces normal iterators with `std::reverse_iterator`.

### DIFF
--- a/source/sema.h
+++ b/source/sema.h
@@ -244,11 +244,11 @@ public:
 
         //  Then look backward to find the first declaration of
         //  this name that is not deeper (in a nested scope)
-        for ( ; i != symbols.cbegin(); --i )
+        for (auto ri = std::make_reverse_iterator(i+1); ri != symbols.crend(); ++ri )
         {
-            if (i->sym.index() == symbol::active::declaration && i->depth <= depth)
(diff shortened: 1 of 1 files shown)


filipsajdak:
I made small mistake that cause skipping one symbol while using `std::reverse_iterator` fixed in last push.

hsutter:
Thanks! I applied the basic change as a separate commit to also rename the function.

filipsajdak:
Verified on my side on my tests - works!

filipsajdak:
I wondered why I had added a bool value. I just realized that I needed to look outside of the named function in another PR: https://github.com/hsutter/cppfront/pull/96 (I needed that to deduce the type of the symbol - it was needed to deduce if we deal with a pointer or not).

I will rebase other PRs and will see what need to be adjusted

PR #105, opened 2022-11-01
add fish shell equivalent for run-tests.bat

kitanatahu:


hsutter:
Thanks! But I'm trying to keep dependencies a minimal as possible at this stage, and so the minimal ones I'm taking are on those Windows `.bat` batch files (and soon, for GCC and Clang compilation only, a Bash `.sh` file each). Sorry! but in the future if this goes past the personal incubation stage it will be time to add things like this (and CI, ...).

Issue #104, opened 2022-10-31
[SUGGESTION] How to cast int to std::size_t


filipsajdak:
Thanks to https://github.com/hsutter/cppfront/pull/93 I was able to implement the `main` function that takes arguments.

```cpp
main: (argc : int, argv : **char) -> int = {
    // ...
}
```
but then `argv` is a pointer and the pointer arithmetic checks will kick in (another PR is needed to achieve that - already on the list). From the messages provided by the safety check, I know that we recommend using `std::span`.

When I try to construct `std::span` I can use pointer and `std::size_t` argument to construct the span but `argc` is an `int`. How to do narrowing conversion in cpp2 way? using `argc as std::size_t` is not working as it matches 
```cpp
static std::nullptr_t nonesuch = nullptr;

template< typename C >
auto as(...) -> auto {
    return nonesuch;
}
```
and fails miserably as a runtime crash :(.

What worked for me right now is
```cpp
    args : std::span<Arg> = (argv, static_cast<std::size_t>(argc));
    // Arg is `using Arg = char*;` as there is an issue with std::span<char*> 
    // or std::span<*char> (star disappears) 
```
but I believe we can do better.

By the way, returning `nonesuch` without a warning when something doesn't match defined cases is a really bad idea - I was looking for an error why the below code ends with a crash:
```cpp
    args : std::span<Arg> = (argv, argc as std::size_t);
    c : std::ifstream = args.back(); // <= here it crashes
```

Probably it is better to do `static_assert` for cases that don't match or something that will give a hint of what is not working.

hsutter:
This is now resolved, and the example below works:

```
main: (argc: int, argv: **char) -> int = {
    args : std::span<*char> = (argv, argc as std::size_t);
    c : std::ifstream = args.back(); // <= now ok
}
```

Thanks!

filipsajdak:
Excellent solution! Thanks for cooperating on that!

gregmarr:
I thought that there had been a discussion about how to make the `main` parameters safe by default, but I wasn't able to find it.  Is there a plan already spelled out somewhere?

JohelEGP:
There was https://github.com/hsutter/cppfront/pull/63.

gregmarr:
Thanks Johel, I think that's what I was thinking of.

PR #106, opened 2022-11-01
[SUGGESTION][FIX] Add narrowing 'as' cast and static assert failure instead of nonesuch…

filipsajdak:
The current implementation allows for writing below code
```cpp
args : std::span<Arg> = (argv, argc as std::size_t);
c : std::ifstream = args.back(); // <= here it crashes
```
It fails as `as<std::size_t>(argc)` matches `as(...)` that returns `nonesuch`. 

This change introduces `as` casts:

1. `as` - static asserted casts,
2. `as std::optional<T>` - optional cast,
3. `cpp2::unsafe_narrow<T>()` - narrow cast,

If the safe cast is not possible, the `as` cast will fail to compile with information to use optional cast `as std::optional<T>`.
```
error: static_assert failed due to requirement 'program_violates_type_safety_guarantee<long, double>' "No safe 'as' cast available please use optional cast `as std::optional<T>`"
```

The specialization of the `as()` functions try to match arithmetic types (algorithm based on the `gsl::narrow()` function). It also removes the version of the `as(...)` function that silently returns `nullptr` - now it fails at compile time thanks to `static_assert`.

After this change, the below code does not compile all narrowing cases, and will trigger static assertion with the information to use two other possible casts:

```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := max_uint8  as std::int8_t;  // static_assert failure
    v2  := -1         as std::uint8_t; // static_assert failure
    v3  := -1.0       as std::uint8_t; // static_assert failure
    v4  := max_double as float;        // static_assert failure

    v5  := -1.0      as long;          // static_assert failure
    v6  := -1        as double;        // OK
    v7  := max_float as double;        // OK
    v8  := max_uint8 as std::uint16_t; // OK
    v9  := max_int8  as std::uint16_t; // static_assert failure
    v10 := 10        as std::size_t;   // OK
    v11 := "string"  as std::string;   // OK
    v12 := 'a'       as int;           // OK
    v13 := 36        as char;          // OK

    std::cout << " v6 = (v6)$" << std::endl;  //  v6 = -1.000000
    std::cout << " v7 = (v7)$" << std::endl;  //  v7 = 340282346638528859811704183484516925440.000000
    std::cout << " v8 = (v8)$" << std::endl;  //  v8 = 255
    std::cout << "v10 = (v10)$" << std::endl; // v10 = 10
    std::cout << "v11 = (v11)$" << std::endl; // v11 = string
    std::cout << "v12 = (v12)$" << std::endl; // v12 = 97
    std::cout << "v13 = (v13)$" << std::endl; // v13 = 36
}
```

Thanks to optional as `as std::optional<T>` we can rewrite it to:
```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := max_uint8  as std::optional<std::int8_t>;  // OK
    v2  := -1         as std::optional<std::uint8_t>; // OK
    v3  := -1.0       as std::optional<std::uint8_t>; // OK
    v4  := max_double as std::optional<float>;        // OK

    v5  := -1.0      as std::optional<long>;          // OK
    v6  := -1        as std::optional<double>;        // OK
    v7  := max_float as std::optional<double>;        // OK
    v8  := max_uint8 as std::optional<std::uint16_t>; // OK
    v9  := max_int8  as std::optional<std::uint16_t>; // OK
    v10 := 10        as std::optional<std::size_t>;   // OK
    v11 := "string"  as std::optional<std::string>;   // OK
    v12 := 'a'       as std::optional<int>;           // OK
    v13 := 36        as std::optional<char>;          // OK

    std::cout << " v1 = (v1)$" << std::endl;  //  v1 = (empty) 
    std::cout << " v2 = (v2)$" << std::endl;  //  v2 = (empty)
    std::cout << " v3 = (v3)$" << std::endl;  //  v3 = (empty)
    std::cout << " v4 = (v4)$" << std::endl;  //  v4 = (empty)
    std::cout << " v5 = (v5)$" << std::endl;  //  v5 = (empty)
    std::cout << " v6 = (v6)$" << std::endl;  //  v6 = -1.000000
    std::cout << " v7 = (v7)$" << std::endl;  //  v7 = 340282346638528859811704183484516925440.000000
    std::cout << " v8 = (v8)$" << std::endl;  //  v8 = 255
    std::cout << " v9 = (v9)$" << std::endl;  //  v9 = 127
    std::cout << "v10 = (v10)$" << std::endl; // v10 = 10 
    std::cout << "v11 = (v11)$" << std::endl; // v11 = string
    std::cout << "v12 = (v12)$" << std::endl; // v12 = 97 
    std::cout << "v13 = (v13)$" << std::endl; // v13 = 36 
}
```

Thanks to `unsafe_narrow()`:

```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := cpp2::unsafe_narrow<std::int8_t>(max_uint8);   // OK
    v2  := cpp2::unsafe_narrow<std::uint8_t>(-1);         // OK
    v3  := cpp2::unsafe_narrow<std::uint8_t>(-1.0);       // OK
    v4  := cpp2::unsafe_narrow<float>(max_double);        // OK

    v5  := cpp2::unsafe_narrow<long>(-1.0);               // OK
    v6  := cpp2::unsafe_narrow<double>(-1);               // OK
    v7  := cpp2::unsafe_narrow<double>(max_float);        // OK
    v8  := cpp2::unsafe_narrow<std::uint16_t>(max_uint8); // OK
    v9  := cpp2::unsafe_narrow<std::uint16_t>(max_int8);  // OK
    v10 := cpp2::unsafe_narrow<std::size_t>(10);          // OK
    v11 := cpp2::unsafe_narrow<std::string>("string");    // OK
    v12 := cpp2::unsafe_narrow<int>('a');                 // OK
    v13 := cpp2::unsafe_narrow<char>(36);                 // OK

    std::cout << " v1 = (v1)$" << std::endl;  //  v1 = -1
    std::cout << " v2 = (v2)$" << std::endl;  //  v2 = 255
    std::cout << " v3 = (v3)$" << std::endl;  //  v3 = 0
    std::cout << " v4 = (v4)$" << std::endl;  //  v4 = inf
    std::cout << " v5 = (v5)$" << std::endl;  //  v5 = -1
    std::cout << " v6 = (v6)$" << std::endl;  //  v6 = -1.000000
    std::cout << " v7 = (v7)$" << std::endl;  //  v7 = 340282346638528859811704183484516925440.000000
    std::cout << " v8 = (v8)$" << std::endl;  //  v8 = 255
    std::cout << "v10 = (v10)$" << std::endl; // v10 = 10
    std::cout << "v11 = (v11)$" << std::endl; // v11 = string
    std::cout << "v12 = (v12)$" << std::endl; // v12 = 97
    std::cout << "v13 = (v13)$" << std::endl; // v13 = 36
}
```

That will also fix below code:
```cpp
args : std::span<Arg> = (argv, (argc as std::optional<std::size_t>).value()); // using `argc as std::size_t` will fail with static_assert
c : std::ifstream = args.back(); // no crash, works as expected
```
Or with `unsafe_narrow`:
```cpp
args : std::span<Arg> = (argv, cpp2::unsafe_narrow<std::size_t>(argc));
c : std::ifstream = args.back(); // no crash, works as expected
```

Closes https://github.com/hsutter/cppfront/issues/104

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1160,6 +1160,18 @@ inline constexpr auto as() -> O
     }
 }
 
+template< typename IsChecked, typename C, typename X >
+auto inspect_as( X&& x ) -> auto
+{
+    using T = std::remove_cvref_t<X>;
+
+    if constexpr (is_optional_v<C> ) {
+        return as<C, T>(std::forward<X>(x));
+    } else {
+        return as<std::optional<C>, T>(std::forward<X>(x)).value();
+    }
+}
+
 template<typename T, auto x>
 constexpr auto as( ) -> decltype(auto)
     { return as<T, x, decltype(x)>();  }
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -740,6 +740,21 @@ auto as(T const& t) -> auto {
     }
 }
 
+template< typename C, auto x, typename X >
+inline constexpr bool is_castable_v = std::is_arithmetic_v<C>
+                                        && std::is_arithmetic_v<X>
+                                        && !(static_cast<X>(static_cast<C>(x)) != x || ((std::is_signed_v<C> != std::is_signed_v<X>) && ((static_cast<C>(x) < C{}) != (x < X{}))));
+
+template< typename C, auto x, typename X>
+inline constexpr auto as() -> auto
+{
+    if constexpr ( is_castable_v<C, x, X> ) {
+        return static_cast<C>(X(x));
+    } else {
+        static_assert(program_violates_type_safety_guarantee<C, X>, "No safe 'as' cast available please use optional cast `as std::optional<T>` or `unsafe_narrow<T>()`");
+    }
+}
+
 template< typename C, typename X >
     requires std::is_same_v<C, X>
 auto as( X const& x ) -> decltype(auto) {
@@ -1132,6 +1147,24 @@ auto as( const X& x ) -> auto
         return O{};
     }
 }
+
+// for literals - compile-time cast
+template< typename O, auto x, typename X, typename C = typename O::value_type >
+    requires (is_optional_v<O> && std::is_arithmetic_v<C> && std::is_arithmetic_v<X>)
+inline constexpr auto as() -> O
+{
+    if constexpr ( is_castable_v<C, x, X> ) {
+        return O{static_cast<C>(X(x))};
+    } else {
+        return O{};
+    }
+}
+
+template<typename T, auto x>
+constexpr auto as( ) -> decltype(auto)
+    { return as<T, x, decltype(x)>();  }
+
+//-----------------------------------------------------------------------
 //
 template<typename T, typename X>
     requires std::is_same_v<X,std::optional<T>>
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1178,11 +1178,13 @@ constexpr auto as( ) -> decltype(auto)
 
 //-----------------------------------------------------------------------
 //
-template<typename T, typename X>
-    requires std::is_same_v<X,std::optional<T>>
-constexpr auto as( X const& x ) -> decltype(auto)
-    { return x.value(); }
+//  Unsafe narrow cast
 
+template <typename C, typename X>
+auto unsafe_narrow( X&& x ) noexcept -> decltype(auto)
+{
+    return static_cast<C>(std::forward<X>(x));
+}
 
 //-----------------------------------------------------------------------
 //
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -208,6 +208,7 @@
     #include <utility>
     #include <cstdio>
     #include <cstdint>
+    #include <cmath>
 
     #if defined(CPP2_USE_SOURCE_LOCATION)
         #include <source_location>
@@ -721,9 +722,22 @@ inline constexpr auto is( auto const& x, auto const& value ) -> bool
 //-------------------------------------------------------------------------------------------------------------
 //  Built-in as
 //
-template< typename C >
-auto as(auto const&) -> auto {
-    return nonesuch;
+template <typename... Ts>
+inline constexpr auto program_violates_type_safety_guarantee = sizeof...(Ts) < 0;
+
+template <typename O>
+inline constexpr bool is_optional_v = requires{ typename O::value_type; requires std::same_as<std::optional<typename O::value_type>, O>;}; 
+
+template< typename C, typename T >
+auto as(T const& t) -> auto {
+    if constexpr (is_optional_v<C>) {
+        return C{};
+    } else if constexpr (is_optional_v<T>) {
+        static_assert(program_violates_type_safety_guarantee<C, T>, "No safe 'as' cast from optional");
+    }
+    else {
+        static_assert(program_violates_type_safety_guarantee<C, T>, "No safe 'as' cast available please use optional cast `as std::optional<T>` or `unsafe_narrow<T>()`");
+    }
 }
 
 template< typename C, typename X >
@@ -734,7 +748,7 @@ auto as( X const& x ) -> decltype(auto) {
 
 template< typename C, typename X >
 auto as( X const& x ) -> auto
-    requires (!std::is_same_v<C, X> && requires { C{x}; })
+    requires (!is_optional_v<C> && !std::is_same_v<C, X> && requires { C{x}; })
 {
     return C{x};
 }
@@ -744,31 +758,6 @@ template< typename C, typename X >
 auto as( X&& x ) -> C&& {
     return std::forward<X>(x);
 }
-
-template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
-auto as( X& x ) -> C& {
-    return dynamic_cast<C&>(x);
-}
-
-template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
-auto as( X const& x ) -> C const& {
-    return dynamic_cast<C const&>(x);
-}
-
-template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
-auto as( X* x ) -> C* {
-    return dynamic_cast<C*>(x);
-}
-
-template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
-auto as( X const* x ) -> C const* {
-    return dynamic_cast<C const*>(x);
-}
-
 
 //-------------------------------------------------------------------------------------------------------------
 //  std::variant is and as
@@ -890,8 +879,10 @@ auto is( std::variant<Ts...> const& x ) {
     return false;
 }
 
-template<typename T, typename... Ts>
-auto as( std::variant<Ts...> const& x ) {
+template< typename O, typename X, typename... Ts >
+    requires is_optional_v<O>
+auto as( std::variant<Ts...> const& x ) -> O {
+    using T = typename O::value_type;
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as< 0>(x)), T >) { if (x.index() ==  0) return operator_as<0>(x); }
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as< 1>(x)), T >) { if (x.index() ==  1) return operator_as<1>(x); }
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as< 2>(x)), T >) { if (x.index() ==  2) return operator_as<2>(x); }
@@ -912,7 +903,7 @@ auto as( std::variant<Ts...> const& x ) {
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as<17>(x)), T >) { if (x.index() == 17) return operator_as<7>(x); }
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as<18>(x)), T >) { if (x.index() == 18) return operator_as<8>(x); }
     if constexpr (std::is_same_v< CPP2_TYPEOF(operator_as<19>(x)), T >) { if (x.index() == 19) return operator_as<9>(x); }
-    throw std::bad_variant_access();
+    return O{};
 }
 
 
@@ -957,10 +948,16 @@ inline constexpr auto is( std::any const& x, auto const& value ) -> bool
 
 //  as
 //
-template<typename T, typename X>
-    requires (!std::is_reference_v<T> && std::is_same_v<X,std::any> && !std::is_same_v<T,std::any>)
-constexpr auto as( X const& x ) -> T
-    { return std::any_cast<T>( x ); }
+template< typename O, typename X, typename T = typename O::value_type>
+    requires (is_optional_v<O> && std::is_same_v<X, std::any> && !std::is_same_v<T,std::any>)
+constexpr auto as( X const& x ) -> O
+    { 
+        try {
+            return std::any_cast<T>( x ); 
+        } catch (...) {
+            return O{};
+        }
+    }
 
 
 //-------------------------------------------------------------------------------------------------------------
@@ -1001,7 +998,140 @@ constexpr auto is( std::optional<T> const& x, auto const& value ) -> bool
 }
 
 
-//  as
+//-------------------------------------------------------------------------------------------------------------
+//  Optional as
+//
+
+template< typename O, typename X, typename C = typename O::value_type >
+    requires ( is_optional_v<O>
+               && std::is_arithmetic_v<C> && std::is_arithmetic_v<X>
+             )
+auto as( const X& x ) -> auto
+{
+    constexpr const bool is_different_signedness = (std::is_signed_v<C> != std::is_signed_v<X>);
+    const C value = static_cast<C>(x);
+    auto is_equal_after_casting = false;
+    if constexpr (std::is_floating_point_v<C> || std::is_floating_point_v<X>) {
+        is_equal_after_casting = std::fabs(static_cast<X>(value) - x) < std::numeric_limits<C>::epsilon();
+    } else {
+        is_equal_after_casting = static_cast<X>(value) == x;
+    }
+
+    if ( !is_equal_after_casting || (is_different_signedness && ((value < C{}) != (x < X{}))))
+    {
+        return O{};
+    }
+
+    return O{value};
+}
+
+template< typename O, typename X, typename C = typename O::value_type >
+auto as( X& x ) -> O
+    requires (is_optional_v<O> && !std::is_same_v<X, std::any> && !is_optional_v<X>
+              && !std::is_arithmetic_v<X> && !std::is_arithmetic_v<C>
+              && !std::is_pointer_v<C> && !std::is_pointer_v<X>
+             )
+{
+    if constexpr ( 
+        std::is_base_of_v<X, C>
+        || (std::is_polymorphic_v<X> && std::is_polymorphic_v<C>)
+    ) {
+        if (auto* ptr = dynamic_cast<C*>(&x); ptr) {
+            return O{*ptr};
+        }
+    }
+    else if constexpr ( requires { C{x}; } ) {
+        return O{x};
+    }
+
+    return O{};
+}
+
+template< typename O, typename X, typename C = typename O::value_type >
+auto as( X const& x ) -> O
+    requires (is_optional_v<O> && !std::is_same_v<X, std::any> && !is_optional_v<X>
+              && !std::is_arithmetic_v<X> && !std::is_arithmetic_v<C>
+              && !std::is_pointer_v<C> && !std::is_pointer_v<X>
+             )
+{
+    if constexpr ( 
+        std::is_base_of_v<X, C>
+        || (std::is_polymorphic_v<X> && std::is_polymorphic_v<C>)
+    ) {
+        if (auto* ptr = dynamic_cast<C const*>(&x); ptr) {
+            return O{*ptr};
+        }
+    }
+    else if constexpr ( requires { C{x}; } ) {
+        return O{x};
+    }
+
+    return O{};
+}
+
+template< typename O, typename X, typename C = typename O::value_type >
+auto as( X const* x ) -> O
+    requires (is_optional_v<O> 
+              && !std::is_arithmetic_v<X> && !std::is_arithmetic_v<C>
+             )
+{
+    if constexpr ( std::is_pointer_v<C> ) {
+        using T = std::remove_pointer_t<C>;
+        if constexpr ( 
+            std::is_base_of_v<X, T>
+            || (std::is_polymorphic_v<X> && std::is_polymorphic_v<T>)
+        ) {
+
+            if (auto* ptr = dynamic_cast<C>(x); ptr) {
+                return O{ptr};
+            }
+        }
+        return O{};
+    }
+    else if constexpr ( requires { C{x}; } ) {
+        return O{x};
+    } else {
+        return O{};
+    }
+}
+
+template< typename O, typename X, typename C = typename O::value_type >
+auto as( X* x ) -> O
+    requires (is_optional_v<O> 
+              && !std::is_arithmetic_v<X> && !std::is_arithmetic_v<C>
+             )
+{
+    if constexpr ( std::is_pointer_v<C> ) {
+        using T = std::remove_pointer_t<C>;
+        if constexpr ( 
+            std::is_base_of_v<X, T>
+            || (std::is_polymorphic_v<X> && std::is_polymorphic_v<T>)
+        ) {
+
+            if (auto* ptr = dynamic_cast<C>(x); ptr) {
+                return O{ptr};
+            }
+        }
+        return O{};
+    }
+    else if constexpr ( requires { C{x}; } ) {
+        return O{x};
+    } else {
+        return O{};
+    }
+}
+
+// optional as optional
+template< typename O, typename X >
+    requires ( is_optional_v<O> && is_optional_v<X> && !std::is_same_v<O, X>)
+auto as( const X& x ) -> auto
+{
+    if (x.has_value()) {
+        return as<O>(x.value());
+    } else {
+        return O{};
+    }
+}
 //
 template<typename T, typename X>
     requires std::is_same_v<X,std::optional<T>>
--- a/regression-tests/test-results/mixed-inspect-values.cpp
+++ b/regression-tests/test-results/mixed-inspect-values.cpp
@@ -50,7 +50,7 @@ auto test(auto const& x) -> void{
         else if (cpp2::is(__expr, in_2_3)) { if constexpr( requires{"3";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(("3")),std::string> ) return "3"; else return std::string{}; else return std::string{}; }
         else if (cpp2::is(__expr, std::move(forty_two))) { if constexpr( requires{"the answer";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(("the answer")),std::string> ) return "the answer"; else return std::string{}; else return std::string{}; }
         else if (cpp2::is<int>(__expr)) { if constexpr( requires{"integer " + cpp2::to_string(x);} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(("integer " + cpp2::to_string(x))),std::string> ) return "integer " + cpp2::to_string(x); else return std::string{}; else return std::string{}; }
-        else if (cpp2::is<std::string>(__expr)) { if constexpr( requires{cpp2::as<std::string>(x);} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::as<std::string>(x))),std::string> ) return cpp2::as<std::string>(x); else return std::string{}; else return std::string{}; }
+        else if (cpp2::is<std::string>(__expr)) { if constexpr( requires{cpp2::inspect_as<std::string, std::string>(x);} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::inspect_as<std::string, std::string>(x))),std::string> ) return cpp2::inspect_as<std::string, std::string>(x); else return std::string{}; else return std::string{}; }
         else return "(no match)"; }
     () << "\n";
 }
--- a/regression-tests/test-results/pure2-inspect-expression-in-generic-function-multiple-types.cpp
+++ b/regression-tests/test-results/pure2-inspect-expression-in-generic-function-multiple-types.cpp
@@ -35,8 +35,8 @@ auto test_generic(auto const& x, auto const& msg) -> void{
         << std::setw(30) << msg 
         << " value is " 
         << [&] () -> std::string { auto&& __expr = x;
-            if (cpp2::is<int>(__expr)) { if constexpr( requires{"integer " + std::to_string(cpp2::as<int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(("integer " + std::to_string(cpp2::as<int>(x)))),std::string> ) return "integer " + std::to_string(cpp2::as<int>(x)); else return std::string{}; else return std::string{}; }
-            else if (cpp2::is<std::string>(__expr)) { if constexpr( requires{'"' + cpp2::as<std::string>(x) + '"';} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(('"' + cpp2::as<std::string>(x) + '"')),std::string> ) return '"' + cpp2::as<std::string>(x) + '"'; else return std::string{}; else return std::string{}; }
+            if (cpp2::is<int>(__expr)) { if constexpr( requires{"integer " + std::to_string(cpp2::inspect_as<int, int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(("integer " + std::to_string(cpp2::inspect_as<int, int>(x)))),std::string> ) return "integer " + std::to_string(cpp2::inspect_as<int, int>(x)); else return std::string{}; else return std::string{}; }
+            else if (cpp2::is<std::string>(__expr)) { if constexpr( requires{'"' + cpp2::inspect_as<std::string, std::string>(x) + '"';} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(('"' + cpp2::inspect_as<std::string, std::string>(x) + '"')),std::string> ) return '"' + cpp2::inspect_as<std::string, std::string>(x) + '"'; else return std::string{}; else return std::string{}; }
             else return "not an int or string"; }
         () 
         << "\n";
--- a/regression-tests/test-results/pure2-inspect-expression-with-as-in-generic-function.cpp
+++ b/regression-tests/test-results/pure2-inspect-expression-with-as-in-generic-function.cpp
@@ -22,7 +22,7 @@ auto print_an_int(auto const& x) -> void{
         << std::setw(30) << cpp2::to_string(x) 
         << " value is " 
         << [&] () -> std::string { auto&& __expr = x;
-            if (cpp2::is<int>(__expr)) { if constexpr( requires{std::to_string(cpp2::as<int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((std::to_string(cpp2::as<int>(x)))),std::string> ) return std::to_string(cpp2::as<int>(x)); else return std::string{}; else return std::string{}; }
+            if (cpp2::is<int>(__expr)) { if constexpr( requires{std::to_string(cpp2::inspect_as<int, int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((std::to_string(cpp2::inspect_as<int, int>(x)))),std::string> ) return std::to_string(cpp2::inspect_as<int, int>(x)); else return std::string{}; else return std::string{}; }
             else return "not an int"; }
         () 
         << "\n";
--- a/regression-tests/test-results/pure2-type-safety-2-with-inspect-expression.cpp
+++ b/regression-tests/test-results/pure2-type-safety-2-with-inspect-expression.cpp
@@ -35,7 +35,7 @@ auto test_generic(auto const& x, auto const& msg) -> void{
         << std::setw(30) << msg 
         << " value is " 
         << [&] () -> std::string { auto&& __expr = x;
-            if (cpp2::is<int>(__expr)) { if constexpr( requires{std::to_string(cpp2::as<int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((std::to_string(cpp2::as<int>(x)))),std::string> ) return std::to_string(cpp2::as<int>(x)); else return std::string{}; else return std::string{}; }
+            if (cpp2::is<int>(__expr)) { if constexpr( requires{std::to_string(cpp2::inspect_as<int, int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((std::to_string(cpp2::inspect_as<int, int>(x)))),std::string> ) return std::to_string(cpp2::inspect_as<int, int>(x)); else return std::string{}; else return std::string{}; }
             else return "not an int"; }
         () 
         << "\n";
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1335,13 +1335,39 @@ public:
                     statement.pop_back();
                 }
 
+                auto replace_all = [](std::string inout, std::string_view what, std::string_view with){
+                    for (std::string::size_type pos{};
+                        inout.npos != (pos = inout.find(what.data(), pos, what.length()));
+                        pos += with.length()) {
+                        inout.replace(pos, what.length(), with.data(), with.length());
+                    }
+                    return inout;
+                };
+
+                auto is_statement_type = [&]() -> std::string {
+                    if (!is_expression) return {};
+                    auto statement = std::string{};
+                    printer.emit_to_string(&statement);
+                    if (alt->type_id) {
+                        emit(*alt->type_id);
+                    } else if (alt->value) {
+                        printer.print_cpp2("CPP2_TYPEOF((", alt->position());
+                        emit(*alt->value);
+                        printer.print_cpp2("))", alt->position());
+                    }
+                    printer.emit_to_string();
+                    return statement;
+                }();
+
+                auto internal_statement = replace_all(statement, "cpp2::as<", "cpp2::inspect_as<" + (is_statement_type.empty() ? "" : is_statement_type + ", "));
+
                 //  If this is an inspect-expression, we'll have to wrap each alternative
                 //  in an 'if constexpr' so that its type is ignored for mismatches with
                 //  the inspect-expression's type
                 auto return_prefix = std::string{};
                 auto return_suffix = std::string{";"};   // use this to tack the ; back on in the alternative body
                 if (is_expression) {
-                    return_prefix = "{ if constexpr( requires{" + statement + ";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((" + statement + "))," + result_type + "> ) return ";
+                    return_prefix = "{ if constexpr( requires{" + internal_statement + ";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((" + internal_statement + "))," + result_type + "> ) return ";
                     return_suffix += " }";
                 }
 
@@ -1363,7 +1389,7 @@ public:
                     printer.print_cpp2(return_prefix, alt->position());
                 }
 
-                printer.print_cpp2(statement, alt->position());
+                printer.print_cpp2(internal_statement, alt->position());
 
                 if (is_expression && id != "_") {
                     assert(alt->statement->statement.index() == statement_node::expression);
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1996,6 +1996,16 @@ public:
     {
         std::string prefix = {};
         std::string suffix = {};
+        bool        as_on_literal = false;
+
+        assert(n.expr);
+        assert(n.expr->get_postfix_expression_node()->expr);
+        if (auto t = n.expr->get_postfix_expression_node()->expr->get_token();
+            t && is_literal(t->type()) && t->type() != lexeme::StringLiteral && t->type() != lexeme::FloatLiteral
+            && std::ssize(n.ops) > 0 && *n.ops[0].op == "as"
+        ) {
+            as_on_literal = true;
+        }
 
         for (auto i = n.ops.rbegin(); i != n.ops.rend(); ++i)
         {
@@ -2014,8 +2024,14 @@ public:
             }
         }
 
+        if (as_on_literal) {
+            auto last_pos = prefix.rfind('>'); assert(last_pos != prefix.npos);
+            prefix.insert(last_pos, ", " + print_to_string(*n.expr));
+        }
+
         printer.print_cpp2(prefix, n.position());
-        emit(*n.expr);
+        if(!as_on_literal)
+            emit(*n.expr);
         printer.print_cpp2(suffix, n.position());
     }
 


JohelEGP:
```C++
    v1  := max_uint8  as std::int8_t;   // throws narrowing_error
```

That seems like a departure from https://wg21.link/P2392, which matches
```
Otherwise, if x can be bound to a refto(T,x), then x as T means a refto(T,x) bound to x.
```

I think you can confirm this with Circle.

hsutter:
Thanks! General timing note (maybe I should paste/link this in a few other PRs/issues if this isn't visible enough):

My responses will be delayed because of the Kona meeting next week, which will be the first face-to-face ISO C++ meeting in nearly three years, and the week before and after the meeting are generally also consumed with committee-related work. I plan to be back to a normal schedule after that and will be catching up with PRs and issues then. Thanks!

hsutter:
But quick ack on this: The intent is to require `narrow_cast` for narrowing conversions, but I haven't fleshed that out with `as` yet.

filipsajdak:
Thanks for the feedback. I assumed that you are preparing for the Kona meeting and I am not expecting that you will process these PRs at the moment. I hope I will be able to join one of the committee meetings in the future.

Regarding this PR. I like the cleanliness of the `x as T` syntax (easy to learn and easy to read). I agree that narrowing operations shall be visible and shall stand out. I was thinking about making it

```cpp
x narrowed_to T; // or something that you can easily read
```

The good thing is that we can have `static_assert` that can inform not to use the `as` keyword for narrowing cases and to use `narrowed_to` (similar case as with pointer arithmetic where we recommend using `std::span` or `gsl::span`). That approach will make writing generic code harder but will guide the programmer and will stand out like this:

```cpp
args : std::span<Arg> = (argv, argc narrowed_to std::size_t);
c : std::ifstream = args.back();
```

filipsajdak:
We can brainstorm on the keyword name to make it clear. Let's first agree on the general idea of how to tackle that cases.

feature-engineer:
> 1. Require the explicit "narrow" syntax because potentially-lossy conversions should stand out, and make `as` not support narrowing conversions.
> 
>     2. The one being suggested here, which is to have `as` do a `narrow` which throws on data loss, but this adds a branch and makes the operation throwing.
> 
>     3. Have `as` do a `narrow_cast` which is a "trust me" operation.

I have a broader language suggestion that may be useful in more cases than just narrow casts:

Maybe we could indicate whether we want to throw on mismatch, static assert, or return an optional instead by adding `?` for optional/ and `!` for static assert (leaving throwing as default)?
e.g. `x as? T;` would return optional which is empty or some lightweight default exception with error_condition if no match exists, `x as! T;` would fail at compile time if no match exists (or if `x`'s value is not defined at compile time), and `x as T;` would throw an exception if no match exists. 

The motivation for having built-in defaults for matching when no pattern applies is easily delegating control over what happens to the caller. 

In the context of this problem, this would mean we just match on cases where there's no data loss, and have the caller decide if they want an optional (or lightweight exception), a thrown exception, or a static assertion in case there's data loss by selecting the type of match in the calling site.

JohelEGP:
I think it makes sense for
```C++
    v1  := max_uint8  as std::int8_t;   // throws narrowing_error
```
to use `narrow` instead. After all, it's safe, just like when casting `std::variant`s and class hierarchies with `as`. Perhaps a difference is that the latter two the compiler can optimize when the `as` is preceded by a corresponding `is` check. Would the same work with `v1`? Looking at the paper, in a generic context, `is` would do something different, as it matches
> Otherwise, if C(x) is valid and convertible to bool, then x is C means C(x).

> @JohelEGP P2392 is a living document that will be updated based on this and other experience (and committee feedback) and can be expected to be a little out of sync sometimes. But for `max_uint8 as std::int8_t` I don't think the bullet you quoted relates because you can't bind a `uint8_t` to an `int8_t&` (in Cpp1 syntax)? But I'm writing this quickly and may be missing something.

You're right. I think it actually matches
> Otherwise, if typeof(x) is implicitly convertible to T, then x as T means to convert x to an rvalue of type
T (e.g., including the case where both are Pointer types and this is a static upcast).

filipsajdak:
Thanks to @feature-engineer I decided to play with exclamation marks and question marks. I don't like `as!` or `as?` - what I like with the pure `as` is that you can read it and it will make sense:
```cpp
pass(x as std::uint16_t);
```
I decided to try `!` and `?` at the end.

So if you want to be strict about casting you can write code adding exclamation marks after the type:
```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v   := max_uint8  as std::uint16_t!; // OK

    v1  := max_uint8  as std::int8_t!;   // compile-time error
    v2  := -1         as std::uint8_t!;  // compile-time error
    v3  := -1.0       as std::uint8_t!;  // compile-time error
    v4  := max_double as float!;         // compile-time error
 
    v5  := -1.0      as long!;           // compile-time error
    v6  := -1        as double!;         // compile-time error
    v7  := max_float as double!;         // OK
    v8  := max_uint8 as std::uint16_t!;  // OK
    v9  := max_int8  as std::uint16_t!;  // compile-time error
    v10 := 10        as std::size_t!;    // compile-time error
    v11 := max_uint8 as std::int16_t!;   // OK
}
```

If you don't want static asserts but you don't want to have exceptions you can use question marks to get `std::optional`:
```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v   := max_uint8  as std::uint16_t?; // OK, 255

    v1  := max_uint8  as std::int8_t?;   // OK, (empty)
    v2  := -1         as std::uint8_t?;  // OK, (empty)
    v3  := -1.0       as std::uint8_t?;  // OK, (empty)
    v4  := max_double as float?;         // OK, (empty)
 
    v5  := -1.0      as long?;           // OK, -1
    v6  := -1        as double?;         // OK, -1.000000
    v7  := max_float as double?;         // OK, 340282346638528859811704183484516925440.000000
    v8  := max_uint8 as std::uint16_t?;  // OK, 255
    v9  := max_int8  as std::uint16_t?;  // OK, 127
    v10 := 10        as std::size_t?;    // OK, 10
    v11 := max_uint8 as std::int16_t?;   // OK, 255

    std::cout << "v = (v)$" << std::endl; 
    std::cout << "v1 = (v1)$" << std::endl; 
    std::cout << "v2 = (v2)$" << std::endl; 
    std::cout << "v3 = (v3)$" << std::endl; 
    std::cout << "v4 = (v4)$" << std::endl; 
    std::cout << "v5 = (v5)$" << std::endl; 
    std::cout << "v6 = (v6)$" << std::endl; 
    std::cout << "v7 = (v7)$" << std::endl; 
    std::cout << "v8 = (v8)$" << std::endl; 
    std::cout << "v9 = (v9)$" << std::endl; 
    std::cout << "v10 = (v10)$" << std::endl; 
    std::cout << "v11 = (v11)$" << std::endl; 
}
```

If you don't want to have `static_assert` checks nor use `std::optional` you can stay with just plain `as` and you will get a runtime check with an exception on narrowing error.

Currently, `optional_as` is using exceptions (captures them) - it is only a prototype to play with it. We shall use `std::expected` or something similar to handle narrowing error cases and throw only in specific cases.

My only doubt is that there is a very subtle difference between these cases:
```cpp
pass(x as std::uint8_t);
pass(x as std::uint8_t!);
pass(x as std::uint8_t?);
```
probably a better option would be to have something like
```cpp
pass(x as_optional std::uint8_t);
pass(x as_strict std::uint8_t);
pass(x as std::uint8_t);
```

PS: have you noticed how easy it is to print `std::optional` - it is awesome!

gregmarr:
I don't like the `!` and `?` after the type, too easy to miss.  I was thinking along the lines of `x as optional int` gives `std::optional<int>`, `x as expected int` gives `std::expected<int>`, `x as narrow int` allows narrowing conversions.

jcanizales:
My understanding of the future of exceptions (per https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0709r4.pdf ) is that `std::expected` is not going to be a recommended tool. So enshrining it in the language itself might be counterproductive.

filipsajdak:
Ok. Will read that through.

filipsajdak:
Continuing my experiments with the syntax I have pushed the new update that allows using two syntaxes:

```cpp
pass(x as std::uint8_t?);
pass(x as std::uint8_t!);
pass(x as std::uint8_t);
```
or
```cpp
pass(x as std::optional<std::uint8_t>);
pass(x as cpp2::strict<std::uint8_t>);
pass(x as std::uint8_t);
```

Taking a look at the possible options and current experiments probably strict mode shall be the default one as it is not narrowing. 

Possibly narrowing the cast shall be emphasized instead of a strict case so maybe we should change it to something similar to:
```cpp
pass(x as std::optional<std::uint8_t>);
pass(x as std::uint8_t); // static_assert failure if possibly dangerous
pass(x as cpp2::narrowing<std::uint8_t>);
```

feature-engineer:
> My understanding of the future of exceptions (per https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0709r4.pdf ) is that `std::expected` is not going to be a recommended tool. So enshrining it in the language itself might be counterproductive.

I don't think your assessment is correct - according to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/n4907.pdf, P0323R12 (std::expected) is going to be included in the language for C++23. 

It seems that it'd be quite some time before p0709, or P2232, or some other yet to be proposed proposal are going to deprecate std::expected.

jcanizales:
I guess it's just a matter of how forward you look. The paper I linked is clear about what you said:

> In the committee, we are advancing the proposed std::experimental::expected<SuccessResult,ErrorResult> [P0323R3]. As noted in the paper: “C++ already supports exceptions and error codes, expected would be a third kind of error handling.”

But that paper is Herb writing 60 pages about how he dislikes adding (with `std::expected`) yet another way of handling errors, and would like instead a unification that covers all requirements.

filipsajdak:
Yes, you are right. I did the experiments and came to that conclusion as well.

To be honest I am waiting for a feedback from Herb as I don't know if has something in mind regarding this feature.

Probably I can make all these syntaxes working. I am currently waiting to clarify the defaults and how to distinguish other cases from the default behavior.

filipsajdak:
One thing I have learned is that we probably want to have default version that is statically checked on compile time. If that is not working than we shall be able to choose if we want to have exception or e.g. optional.

switch-blade-stuff:
> I guess it's just a matter of how forward you look. The paper I linked is clear about what you said:
> 
> > In the committee, we are advancing the proposed std::experimental::expected<SuccessResult,ErrorResult> [P0323R3]. As noted in the paper: “C++ already supports exceptions and error codes, expected would be a third kind of error handling.”
> 
> But that paper is Herb writing 60 pages about how he dislikes adding (with `std::expected`) yet another way of handling errors, and would like instead a unification that covers all requirements.

From my understanding, `std::expected` is created as the "modern" way of handling error codes, instead of using a C-style out parameter. While it is an alternative to exceptions, i personally see a case for both types of error handling: exceptions being used for "unexpected" error conditions, while error codes being used when the error condition is "expected".

Ex. when connecting to a web socket, you expect to be returned some kind of a handle object, however you also may expect an error code. IMO error codes fit this use-case more since a network error is an actually expected (although erroneous) result. Normally, you would use a nullable return value or `std::optional` and an out parameter for the error code, but now you can handle both with `std::expected`.

Although, granted, out `std::expected` prevents RVO if you need to further work with the stored value.

jcanizales:
The way I read that proposal I linked, "unexpected errors" as in "programmer errors" should be asserting, rather than throwing exceptions that propagate all the way out and terminate the program there.

If you're programmaticaly handling errors, they are by definition errors that you anticipate could happen. Like your socket example. I can't do what I wanted to do (normal control flow), but I can do something about the error. The proposal argues in several places, and I agree, that normal and error control flow in C++ should be separate. I think _separable_ is a better way to call it (as nothing ever stops you from handling errors immediately next to the call site). It has to be that way for constructors, and constructors are central for RAII. And it has to be that way to let you do compositions as simple as chaining methods, `f(g(x))`.

That said, the new exceptions mechanism proposed is very much like "using `std::expected` behind the scenes", just with language support to act like exceptions. The details on that behind-the-scenes are in section 4.1.1. Section 4.1.6 shows the new mechanism side-by-side as a replacement for `std::expected`, and section 5 says one way to measure the success of the new exceptions would be if they actually replace the use of `std::expected` for handling errors.

So _maybe_ the best course for cppfront _is_ to enshrine `std::expected` as the way to return errors in the generated C++ code. And then eventually to have the new syntax hide it.

switch-blade-stuff:
> So maybe the best course for cppfront is to enshrine std::expected as the way to return errors in the generated C++ code. And then eventually to have the new syntax hide it.

I would definitely not want to have some kind of 3d way of handling runtime errors in addition to classic exceptions and `std::expected`, so this approach seems to be the most optimal as of now.

Perhaps opening a new issue would be best? @jcanizales

switch-blade-stuff:
> "unexpected errors" as in "programmer errors" should be asserting, rather than throwing exceptions that propagate all the way out and terminate the program there

Personally, I think classic exceptions are better in this case, since they handle RAII and can be used to "catch all non-fatal errors".

Ex. in a game engine's editor you could catch all non-fatal exceptions and log the error message, then pause the game and allow the user to fix it without crashing.

Assertions are great for debugging, but do not allow you to preform this kind of error handling and may cause resource leaks because of unhandled RAII destructors.

jcanizales:
Classic exceptions are already forbidden in many types of codebases, for the range of issues the paper describes. The point of the proposal is to fix those issues (like `std::expected` does) but without losing the advantages they have.

When I wrote "programmer error" I meant "the code is buggy" errors. The user can't recover from that, and trying to recover programmatically is problematic because there's no knowing how much of your program state is already corrupted by the time you detect the issue. Assertion libraries like glog usually do print an error message and a stack trace before terminating the program. Chrome survives assertions in tabs without crashing by isolating each in its own OS process. That said, like with all guidance I'm sure there's no shortage of cases in which it's not the best. New exceptions are also better for this than old exceptions in any case.

filipsajdak:
I was not thinking about runtime assertions but static assertions that will make your build to fail. No impact on runtime error handling. It should prevent developer from writing potentially buggy code. Additionally it should provide guidance what to do instead.

filipsajdak:
I have done more experiments and tests around that topic. I like the idea of having default `as` as strict as possible that ends with a static assertion if it is a suspicious expression that may not work.

Currently, it works like that:
```cpp
template< typename C, typename... Ts >
auto as(Ts... ts) -> auto {
  static_assert(program_violates_type_safety_guarantee<Ts...>, "safe 'as' cast is not defined");
}
```

I realized that this interfered with the `inspect` implementation as there are used constexpr ifs and requires expressions that are not working when used on expression that ends up with static assertions.

In `pure...` file there is a code:

```cpp
inspect x -> std::string {
  is int = "integer " + std::to_string(x as int);
  is std::string = '"' + x as std::string + '"';
  is _   = "not an int or string";
}
```
`is int = "integer " + std::to_string(x as int);` line compiles to (the original code is oneliner I refortmated it for easier reading)
```cpp
if (cpp2::is<int>(__expr)) {
  if constexpr( requires{ "integer " + std::to_string(cpp2::as<int>(x)); } )
    if constexpr( std::is_convertible_v<CPP2_TYPEOF( "integer " + std::to_string( cpp2::as<int>(x) ) ),
                                        std::string> )
      return "integer " + std::to_string(cpp2::as<int>(x)); 
    else
      return std::string{};
  else 
    return std::string{}; 
}
```

If `cpp2::as<int>(x)` will end up with static assert requires expression will not return false and compilation will fail. I have made an workaround for it - I am preparing the clean commit to update this PR.


TLDR: I have renamed all `cpp2::as` function to `cpp2::internal_as` and added additional template argument that specify failure policy (static assertions or throwing expceptions). `internal_as` is called by one `as` function with default policy:
```cpp
enum class cast_failure_policy {
    static_assertions,
    throws
};

// ...

auto as(X const& x) -> decltype(auto) {
    return internal_as<cast_failure_policy::static_assertions, C>(x);
}
```

cppfront emits statement to the string that is later printed using that string:
```cpp
return_prefix = "{ if constexpr( requires{" + statement + ";} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((" + statement + "))," + result_type + "> ) return ";
```

In `statement` string I am replacing all `cpp2::as` to `cpp2::internal_as<cpp2::cast_failure_policy::throws, ` - when expression throws inside requires call it will end up with false.
```cpp
auto internal_statement = replace_all(statement, "cpp2::as<", "cpp2::internal_as<cpp2::cast_failure_policy::throws, ");
```

filipsajdak:
@hsutter I have made it a draft to emphasize that I am working on it.

filipsajdak:
rebased to main

hsutter:
Thanks, everyone. Now that I've had time to reread this thread, I see @filipsajdak narrowed (pun intended) this down to several suggestions, one of which is the one I was thinking of.

### Briefly: What adding `?` would entail

If we allow `x as? T` to mean an `optional<T>`, then
- If we use a `?` it seems more natural for it to be on the type, i.e., `x as T?`. That would make sense if we wanted to pursue the concept of an optional type as a language feature hardwired throughout the type system... but I'm not inclined to do that because I think would be a major invasive change. C# 2.0 did something similar when it added a `Nullable` language supported concept also spelled `T?`, and the experience I've heard from C# designers was that it was way more work than expected and complicated the type system and compiler, and in retrospect is not clear it was a good idea to add `T?` throughout the type system. Besides, ...
- We already have a way to spell that, namely `x as optional<T>`. Let's try that.

Which brings me to the same idea that @filipsajdak included as one of his options...

### Spelling "safe as"

```
// I like this
x as std::uint8_t // static_assert failure if possibly dangerous
```

That makes sense to me, especially with a nice error message to "use fallible-as or lossy-as" if you really want to do this.

### Spelling "fallible as"

```
// I like this
x as std::optional<std::uint8_t>
```

I like this direction. Note this doesn't require adding new `?`/`!` grammar or a reporting mode enum configuration parameter or anything complex like that. It should work fine with the current compiler just by adding overloads of `as<optional<T>>(x)` in `cpp2util.h`.

### Spelling "lossy as"

Note that this case doesn't just come up with narrowing conversions, but also with (other) slicing that loses part of a value. Derived-to-base slicing is the most well-known, and yes we tell people not to make base classes copyable, but (a) they do and (b) maybe it can even sometimes make sense to slice off part of a larger derived-type value.

As long as it's more visible and opt-in, I think that's fine... so perhaps if we have this:

```
// Hypothetical, but a bit inconsistent (see next paragraph)
x as narrow_to<std::uint8_t>
```

we could implement it as overloads of `auto as<cpp2::narrow_to<T>>(x) -> T` ... but note the expression's type should be `T`, which is different from all the other `as` cases where `as<T>` yields a `T`, and that's one reason at least for now I'd be happy to make this not be `as` at all but leave it as something uglier and more visibly different like GSL already has,

```
// I like this, and GSL has it already as "narrow_cast"
unsafe_narrow<std::uint8_t>(x)
```

Naming it differently also preserves "`as` is safe" as a stake in the ground.

### Additional value constraints?

In some (many?) cases, the calling code actually has more information on the valid/expected value range than just "fits into a `std::uint8_t`"... there may be business logic limits/constraints, so perhaps we should allow optional `min, max` parameters that default to the full value range? Instead of writing an `as` preceded/followed by explicit tests for the business logic limits value range, perhaps allowing the programmer to express it in one shot would make it convenient and easy for programmers to do the right thing.

Example:

    // let's say that x and y shouldn't only fit into a uint8_t,
    // but x should be >=10 and y should be <=127

    // without (min,max)
    a := x as std::optional<uint8_t>; // fallible as
    if a && a.value() >= 10 {
        /*...*/
    }
    b := unsafe_narrow<uint8_t>(y); // lossy "as"
    if b && b <= 127 {
        /*...*/
    }

    // with (min,max)
    a := x as std::optional<uint8_t>(10 /*, default max value*/ ); // fallible as
    if a {
        /*...*/
    }
    b := unsafe_narrow<uint8_t>(y,0,127); // lossy "as"
    if b {
        /*...*/
    }

... and now the interface of `unsafe_narrow` converges with that of [C++17 `std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) which is interesting and makes a lot of sense to me, since narrowing really is (or heavily overlaps with) the concept of value clamping.

hsutter:
Thanks, everyone. Now that I've had time to reread this thread, I see @filipsajdak narrowed (pun intended) this down to several suggestions, one of which is the one I was thinking of.

### Briefly: What adding `?` would entail

If we allow `x as? T` to mean an `optional<T>`, then
- If we use a `?` it seems more natural for it to be on the type, i.e., `x as T?`. That would make sense if we wanted to pursue the concept of an optional type as a language feature hardwired throughout the type system... but I'm not inclined to do that because I think would be a major invasive change. C# 2.0 did something similar when it added a `Nullable` language supported concept also spelled `T?`, and the experience I've heard from C# designers was that it was way more work than expected and complicated the type system and compiler, and in retrospect is not clear it was a good idea to add `T?` throughout the type system. Besides, ...
- We already have a way to spell that, namely `x as optional<T>`. Let's try that.

Which brings me to the same idea that @filipsajdak included as one of his options...

### Spelling "safe as"

```
// I like this
x as std::uint8_t // static_assert failure if possibly dangerous
```

That makes sense to me, especially with a nice error message to "use fallible-as or lossy-as" if you really want to do this.

### Spelling "fallible as"

```
// I like this
x as std::optional<std::uint8_t>
```

I like this direction. Note this doesn't require adding new `?`/`!` grammar or a reporting mode enum configuration parameter or anything complex like that. It should work fine with the current compiler just by adding overloads of `as<optional<T>>(x)` in `cpp2util.h`.

### Spelling "lossy as"

Note that this case doesn't just come up with narrowing conversions, but also with (other) slicing that loses part of a value. Derived-to-base slicing is the most well-known, and yes we tell people not to make base classes copyable, but (a) they do and (b) maybe it can even sometimes make sense to slice off part of a larger derived-type value.

As long as it's more visible and opt-in, I think that's fine... so perhaps if we have this:

```
// Hypothetical, but a bit inconsistent (see next paragraph)
x as narrow_to<std::uint8_t>
```

we could implement it as overloads of `auto as<cpp2::narrow_to<T>>(x) -> T` ... but note the expression's type should be `T`, which is different from all the other `as` cases where `as<T>` yields a `T`, and that's one reason at least for now I'd be happy to make this not be `as` at all but leave it as something uglier and more visibly different like GSL already has,

```
// I like this, and GSL has it already as "narrow_cast"
unsafe_narrow<std::uint8_t>(x)
```

Naming it differently also preserves "`as` is safe" as a stake in the ground.

### Additional value constraints?

In some (many?) cases, the calling code actually has more information on the valid/expected value range than just "fits into a `std::uint8_t`"... there may be business logic limits/constraints, so perhaps we should allow optional `min, max` parameters that default to the full value range? Instead of writing an `as` preceded/followed by explicit tests for the business logic limits value range, perhaps allowing the programmer to express it in one shot would make it convenient and easy for programmers to do the right thing.

Example:

    // let's say that x and y shouldn't only fit into a uint8_t,
    // but x should be >=10 and y should be <=127

    // without (min,max)
    a := x as std::optional<uint8_t>; // fallible as
    if a && a.value() >= 10 {
        /*...*/
    }
    b := unsafe_narrow<uint8_t>(y); // lossy "as"
    if b && b <= 127 {
        /*...*/
    }

    // with (min,max)
    a := x as std::optional<uint8_t>(10 /*, default max value*/ ); // fallible as
    if a {
        /*...*/
    }
    b := unsafe_narrow<uint8_t>(y,0,127); // lossy "as"
    if b {
        /*...*/
    }

... and now the interface of `unsafe_narrow` converges with that of [C++17 `std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) which is interesting and makes a lot of sense to me, since narrowing really is (or heavily overlaps with) the concept of value clamping.

JohelEGP:
> We already have a way to spell that, namely `x as optional<T>`. Let's try that.

> It should work fine with the current compiler just by adding overloads of `as<optional<T>>(x)` in `cpp2util.h`.

This is all about some fundamental types, right? What about UDTs? Rather than throwing constructors, some (niches) use named functions that report failure. In Cpp2, those could be spelled `x as optional<T>`.

But then I think about generic programming, and how `std::optional` isn't always enough to report failure.

This seems like it could create a generic programming pitfall by giving the `as` expression different meanings depending on the arguments.

> Naming it differently also preserves "`as` is safe" as a stake in the ground.

That can't work, right? `is` and `as` being overloadable by necessity, because it needs users to define semantics, means it can do anything within the constraints of the language and the feature themselves.

> ### Spelling "lossy as"

Both this "lossy as" and the previous "fallible as" could be spelled as `x as fallible<T>`, where the `fallible` overload returns a `std::optional<T>` (or a more appropriate failure type for program-defined specializations). Or, perhaps, make use of the support for "throwing values".

I'm not sure over-overloading `as` is a good idea, in face of the compile-times impact of `operator<<` and `invoke_tag`. Constructors are named after the type, which limits the overloads. @hsutter presented something like "all functions that return a `T` can be used to construct a `T`". So perhaps Cpp2 can do better.

JohelEGP:
> This seems like it could create a generic programming pitfall by giving the `as` expression different meanings depending on the arguments.

There are 2 ways the meaning could change. `x as std::optional<T>` is disallowed if `T` is not a fundamental type to which it is given a special meaning, or rather than being disallowed, it has today's meaning. The former could result in the use of `std::optional` deep in a library to fail, while the latter would, in the same case, change the meaning from the intuitive `std::optional<T>(x)` to the special meaning.

A special tag like `fallible` could prevents this. It also sounds more like an error situation, making it more apparent that it is for the user to handle.

What would `x is fallible<T>` do?

hsutter:
@JohelEGP 
> This is all about some fundamental types, right? What about UDTs? 

This is what I meant by narrowing being more general and including slicing: It applies to any lossy type conversion.

Lossy conversions can include:
- lossy built-in narrowing (e.g., `i16` to `u8`)
- lossy `Derived`-to-`Base` slicing (when `Base` is copyable, which is discouraged but sometimes intentional and can be useful)
- lossy conversion operators to narrower-range-or-precision types (e.g., `Employee` might provide an `/*explicit*/ operator std::string` to get the employee's name, essentially throwing away the other information about the employee, which is often discouraged but sometimes intentional and can be useful; or conversion to a related type, such as a high-precision UDT with a conversion operator to a lower-precision UDT that may have the same range (number of exponent bits) but just lower precision)

I would want a design that treats these uniformly, and I think `as std::optional<T>` does that... it provides a common spelling (customization point) for such types to hook into and express a "lossy as" conversion. One thing to learn and teach.

> This seems like it could create a generic programming pitfall by giving the as expression different meanings depending on the arguments.

I do hear that sometimes about `is` and `as`, but I haven't seen it... you know statically from the types what the meaning is, so there isn't a confusion that it could mean different things. IMO one advantage of using `as std::optional<T>` is that it hews strictly to the existing single meaning of `as T`... it always means a conversion to `T`, and if you want it to be optional, you just say so. That's why I like it.

filipsajdak:
I like `x as std::optional<T>`. One issue I have struggled with while implementing it is a problem that `std::optional` cannot hold references.

filipsajdak:
Regarding `unsafe_narrow<std::uint8_t>(x)`. Will it throw on error?

hsutter:
> I like `x as std::optional<T>`. One issue I have struggled with while implementing it is a problem that `std::optional` cannot hold references.

[That's a feature, not a bug.][https://herbsutter.com/2020/02/23/references-simply/] :) It's deliberate that Cpp2 has no references.

IMO lossy-`as` is a cast that should return an optional copy, so `x as optional<T>`.

If we want to support in-place semantics, we could express that using `x& as optional<*T>`, which is nice and explicit. Or just `x& as *T` (since pointers can be null as a result of dynamic tests), which I think we already support for the dynamic cast cases.

> Regarding `unsafe_narrow<std::uint8_t>(x)`. Will it throw on error?

No, it's an unsafe forced conversion. As long as the conversion is statically valid (ignoring value lossiness), this would be a syntax to force it to happen and ignore the consequences, as with `gsl::narrow_cast`. Hence "unsafe."

filipsajdak:
Also, casting a pointer when failing should probably return `nullptr` instead of `std::optional`.

hsutter:
> Also, casting a pointer when failing should probably return `nullptr` instead of `std::optional`.

Right, this is P2392 and what's implemented now for raw pointer dynamic casts:

```
auto as( X* x ) -> C* {
    return dynamic_cast<C*>(x);
}

```

filipsajdak:
Ok, so we miss sidecast.

filipsajdak:
And another thought. It is not safe cast as it might fail. How should we emphasize that? 

In my current implementation I am using `as?` for `std::optional` cast. I was thinking to make it return pointer (or `nullptr`) when casting pointer. 

It would be good not to use `x as T*` or `x as T&` for casts that might fail.

JohelEGP:
> Does that make sense and cover most of your questions?

Yes. Thank you very much.

hsutter:
> Ok, so we miss sidecast.

Good point. That should be added in that existing function too, right?

> It would be good not to use `x as T*` or `x as T&` for casts that might fail.

Good point. Sticking with `x as optional<*T>` is more consistent with other fallible casts. That's a simple consistent place to start with and later see if it needs to be changed.

filipsajdak:
Yes, sidecast is missing in current implementation.

filipsajdak:
Short update after my work on this PR.

I managed to rework the code to handle `safe as`:

```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := max_uint8  as std::int8_t;  // static_assert failure
    v2  := -1         as std::uint8_t; // static_assert failure
    v3  := -1.0       as std::uint8_t; // static_assert failure
    v4  := max_double as float;        // static_assert failure

    v5  := -1.0      as long;          // OK
    v6  := -1        as double;        // OK
    v7  := max_float as double;        // OK
    v8  := max_uint8 as std::uint16_t; // OK
    v9  := max_int8  as std::uint16_t; // static_assert failure
    v10 := 10        as std::size_t;   // OK
    v11 := "string"  as std::string;   // OK
    v12 := 'a'       as int;           // OK

    std::cout << " v5 = (v5)$" << std::endl;
    std::cout << " v6 = (v6)$" << std::endl;
    std::cout << " v7 = (v7)$" << std::endl;
    std::cout << " v8 = (v8)$" << std::endl;
    std::cout << "v10 = (v10)$" << std::endl;
    std::cout << "v11 = (v11)$" << std::endl;
    std::cout << "v12 = (v12)$" << std::endl;
}
```

Optional as:

```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := max_uint8  as std::optional<std::int8_t>;  // OK
    v2  := -1         as std::optional<std::uint8_t>; // OK
    v3  := -1.0       as std::optional<std::uint8_t>; // OK
    v4  := max_double as std::optional<float>;        // OK

    v5  := -1.0      as std::optional<long>;          // OK
    v6  := -1        as std::optional<double>;        // OK
    v7  := max_float as std::optional<double>;        // OK
    v8  := max_uint8 as std::optional<std::uint16_t>; // OK
    v9  := max_int8  as std::optional<std::uint16_t>; // OK
    v10 := 10        as std::optional<std::size_t>;   // OK
    v11 := "string"  as std::optional<std::string>;   // OK
    v12 := 'a'       as std::optional<int>;           // OK

    std::cout << " v1 = (v1)$" << std::endl;  //  v1 = (empty) 
    std::cout << " v2 = (v2)$" << std::endl;  //  v2 = (empty)
    std::cout << " v3 = (v3)$" << std::endl;  //  v3 = (empty)
    std::cout << " v4 = (v4)$" << std::endl;  //  v4 = (empty)
    std::cout << " v5 = (v5)$" << std::endl;  //  v5 = -1
    std::cout << " v6 = (v6)$" << std::endl;  //  v6 = -1.000000
    std::cout << " v7 = (v7)$" << std::endl;  //  v7 = 340282346638528859811704183484516925440.000000
    std::cout << " v8 = (v8)$" << std::endl;  //  v8 = 255
    std::cout << " v9 = (v9)$" << std::endl;  //  v9 = 127
    std::cout << "v10 = (v10)$" << std::endl; // v10 = 10 
    std::cout << "v11 = (v11)$" << std::endl; // v11 = string 
    std::cout << "v12 = (v12)$" << std::endl; // v12 = 97 
}
```

Plus some polymorphic casts (with sidecast):
```cpp
struct B1 { virtual ~B1() = default; };

struct B2 { virtual ~B2() = default; };

struct D : B1, B2 {};

template <typename T>
using ptr = T*;

main: () -> int = {
    d: D = ();
    p: * B1 = d&;

    std::cout << p* is B2 << std::endl;

    dp := p as std::optional<ptr<B2>>;
    b2p := p* as std::optional<B2>;

    std::cout << "dp  = (dp)$" << std::endl;
    std::cout << "b2p = (b2p)$" << std::endl;
}
```

Side note: I am not able to use `std::optional<*B2>` there is an error with parsing:
```
tests/is_polymorph.cpp2...
is_polymorph.cpp2(16,30): error: invalid expression after Less: < (at '*')
is_polymorph.cpp2(16,30): error: expected ; at end of statement (at '*')
is_polymorph.cpp2(16,30): error: ill-formed initializer (at '*')
is_polymorph.cpp2(16,8): error: expected ; at end of statement (at ':')
is_polymorph.cpp2(16,8): error: invalid statement in compound-statement (at ':')
is_polymorph.cpp2(16,30): error: invalid expression after Less: < (at '*')
is_polymorph.cpp2(16,30): error: expected ; at end of statement (at '*')
is_polymorph.cpp2(16,30): error: ill-formed initializer (at '*')
is_polymorph.cpp2(16,31): error: ill-formed initializer (at 'B2')
is_polymorph.cpp2(10,1): error: unexpected text at end of Cpp2 code section (at 'main')
is_polymorph.cpp2(9,0): error: parse failed for section starting here
```
My other PRs will solve it (those regarding pointers).

I still did not touch the issue with `inspect expressions` that use `require` expressions to deduce what branch to choose. Unfortunately, when `static_assert` is triggered in the `require` clause, the compilation stops with an error. In my previous implementation, I replace all "safe as" with "throwable as" (in inspect) that can fail safely in the `require` expression.

filipsajdak:
I have pushed the code that passes all regression tests and implements:

* static asserted as - aka safe as,
* `as std::optional<T>`- aka optional as,

It requires:

* adding `unsafe_narrow<T>(x)`,
* some cleanups,

@hsutter I have added `inspect_as()` to make `inspect` expression work - it is just `as()` that throws on failure (that can be used in `requires` expressions. I am thinking if we need `throwing_as`. Currently the code that in my previous implementation looked like
```cpp
args : std::span<Arg> = (argv, argc as! std::optional<std::size_t>); // `as!` stands for `throwing as`
```
now will look
```cpp
args : std::span<Arg> = (argv, (argc as std::optional<std::size_t>).value());
```

filipsajdak:
I have added `cpp2::unsafe_narrow` that works as shown below:

```cpp
main: () -> int = {
    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := cpp2::unsafe_narrow<std::int8_t>(max_uint8);   // OK
    v2  := cpp2::unsafe_narrow<std::uint8_t>(-1);         // OK
    v3  := cpp2::unsafe_narrow<std::uint8_t>(-1.0);       // OK
    v4  := cpp2::unsafe_narrow<float>(max_double);        // OK

    v5  := cpp2::unsafe_narrow<long>(-1.0);               // OK
    v6  := cpp2::unsafe_narrow<double>(-1);               // OK
    v7  := cpp2::unsafe_narrow<double>(max_float);        // OK
    v8  := cpp2::unsafe_narrow<std::uint16_t>(max_uint8); // OK
    v9  := cpp2::unsafe_narrow<std::uint16_t>(max_int8);  // OK
    v10 := cpp2::unsafe_narrow<std::size_t>(10);          // OK
    v11 := cpp2::unsafe_narrow<std::string>("string");    // OK
    v12 := cpp2::unsafe_narrow<int>('a');                 // OK

    std::cout << " v1 = (v1)$" << std::endl;  //  v1 = -1
    std::cout << " v2 = (v2)$" << std::endl;  //  v2 = 255
    std::cout << " v3 = (v3)$" << std::endl;  //  v3 = 0
    std::cout << " v4 = (v4)$" << std::endl;  //  v4 = inf
    std::cout << " v5 = (v5)$" << std::endl;  //  v5 = -1
    std::cout << " v6 = (v6)$" << std::endl;  //  v6 = -1.000000
    std::cout << " v7 = (v7)$" << std::endl;  //  v7 = 340282346638528859811704183484516925440.000000
    std::cout << " v8 = (v8)$" << std::endl;  //  v8 = 255
    std::cout << "v10 = (v10)$" << std::endl; // v10 = 10
    std::cout << "v11 = (v11)$" << std::endl; // v11 = string
    std::cout << "v12 = (v12)$" << std::endl; // v12 = 97
}
```

hsutter:
Thanks, quick ack:

> My other PRs will solve it (those regarding pointers).

I just pushed a commit that moved `*` to _type-id_, which was overdue cleanup needed for other things (including multilevel pointers). And that's a step toward allowing `std::optional<*B1>` and also `as *B1` (both of which don't parse today but I'm working on it...).

filipsajdak:
Sorry that I did not push those pointers PRs, yet. 

I have stuck with this `as` PR that I want to finalize. Almost there. I will push it soon for review.

I have doubts about casting polymorphic types with `as optional<T>` that make a copy of a casted object.

Probably it should be banned with information to use `as optional<*T>`.

filipsajdak:
I have cleaned the PR - I think it is ready for review.

hsutter:
Another quick ack: FYI, among other things today's commit also allows `as *Base` and similar. No more need for the `ptr<Base>` workaround in test code.

filipsajdak:
Ah... I see that you have changed `as` generation. I need to rebase and align with the new way.

It is right direction. I was hoping to switch to pointers PRs. Not today.

hsutter:
Yes, I realized you were working on a PR but it seemed right that I should (finally) fix the grammar and parsing issues to correctly support multi-`*`, `*` with `const`, `as *T`, and similar that would affect what you're doing. Hopefully this will help your PR even though it needs a rebase to pull these in. Thanks!

Just let me know anytime if I can coordinate a better way in the future, and thanks for the help.

---

Also I just checked in support for the C-style multi-token fundamental types... things like `unsigned long long` are now merged into a single Cpp2 token that contains internal whitespace, which I think is an elegant solution for those and that didn't disturb the rest of the code. With those fundamental types now all supported, and the multi-`*` pointers supported, I think we now can utter all the C/C++ types for full compatibility... please open an issue if I missed any.

filipsajdak:
@hsutter I have made rebase and merged my code with your latest changes.

The only thing I need to merge is `auto emit(is_as_expression_node const& n) -> void`. I made it quick. This change is needed to be able to handle casting a literal at compile time - it makes the below code work:
```cpp
    v5  := -1.0      as long;          // OK
    v6  := -1        as double;        // OK
    v10 := 10        as std::size_t;   // OK
```

It works as it is checked at compile-time that values fit the casted type. It currently works only for the first cast. 
```cpp
a := -1.0 as long;                 // works, -1.0 fit to long
b := -1.0 as long as float;        // works, -1.0 fit to long, and long fit to float
c := -1.0 as long as float as int; // doesn't work, -1.0 fit to long, long fit float, but float does not fit to int
```

Please review the whole concept. If you like it, I can work on `auto emit(is_as_expression_node const& n) -> void` to make it more generic.

All regression tests pass.

JohelEGP:
> Please review the whole concept.

Do you mean this?

> to be able to handle casting a literal at compile time

If so, I think it's necessary. If `T{Lit}` works, where `Lit` is a literal, `Lit as T` should, too.

filipsajdak:
@JohelEGP currently, it works differently.

```cpp
long l{1.0};
```

Does not compile with error:
```
<source>: In function 'int square(int)':
<source>:4:12: error: narrowing conversion of '1.0e+0' from 'double' to 'long int' [-Wnarrowing]
    4 |     long l{1.0};
      |            ^~~
Compiler returned: 1
```

But it works with cpp2:
```cpp
l := 1.0 as long;
```

JohelEGP:
You're right. Since you mentioned "compile-time" at https://github.com/hsutter/cppfront/pull/106#issuecomment-1368035512

> This change is needed to be able to handle casting a literal at compile time - it makes the below code work:
> 
> ```c++
>     v5  := -1.0      as long;          // OK
>     v6  := -1        as double;        // OK
>     v10 := 10        as std::size_t;   // OK
> ```
> 
> It works as it is checked at compile-time that values fit the casted type. It currently works only for the first cast.

I was thinking of the exceptions in https://eel.is/c++draft/dcl.init.list#7 that makes list-initialization like `short i{1};` not a narrowing conversion. So as a minimum, I'd expect `as` casts to also do the same for those same cases.

hsutter:
Thanks, I'm starting to take a look now...

As a first step, let me try to sum up the error handling policy for dynamic typing -- this may not be identical to P2392 but P2392 does have throwing for failed dynamic casts (and it did not have this "as std::optional" which is new here) and Cpp2 and P2392 will continue to be re-converged to keep the design in sync as it's tweaked:

| Operation  | Static check  | Dynamic check  | Can throw? |
| ----------- | ----------- | ----------- | ----------- |
| Explicit narrowing: `unsafe_narrow<T>(x)` | `x` must be convertible to `T` | None - this is a "trust me, I'm telling you to just throw away the extra values bits if it doesn't fit" operation | No |
| Pointer dynamic downcast: `x as *Derived` | `x` must be a pointer to a base of `Derived` or to `Derived` | If the cast fails, the expression yields `nullptr` | No |
| `variant` dynamic "downcast": `x as T` where `x` is `std::variant<Ts>` | `Ts` must contain `T` | If `x` does not hold a `T`, the cast throws an exception | Yes |
| `variant` dynamic "downcast": `x as std::optional<T>` where `x` is `std::variant<Ts>` | `Ts` must contain `T` | If `x` does not hold a `T`, the resulting optional is empty | No |
| `optional` dynamic "downcast": `x as T` where `x` is `std::optional<U>` | `U` must be `T` | If the optional is empty, the cast throws an exception | Yes |
| `optional` dynamic "downcast": `x as std::optional<T>` where `x` is `std::optional<U>` | `U` must be `T` | No-op | No |
| `any` dynamic "downcast": `x as T` where `x` is `std::any` | None  | If `x` does not hold a `T`, the cast throws an exception | Yes |
| `any` dynamic "downcast": `x as std::optional<T>` where `x` is `std::any` | None | If `x` does not hold a `T`, the resulting optional is empty | No |

Does that make sense?

filipsajdak:
@hsutter thank you for an excellent summary table.

The main difference is that current PR treats the `as` cast as a “safe cast” that is checked at compile time. That means it never throws nor fails in any other way.

The idea is that if cppfront accepts it, it will work. If it fails with static casts, I want to instruct the user on what other kind of cast they should consider.

That is why I think it would be good to have a distinguishable way to say `throwable_as`.

We can quickly identify where the cast might fail by looking at the code, assuming it will be distinguishable and statically checked.

filipsajdak:
At the beginning of this PR, I have proposed three kind of casts:

1. `as` - statically checked safe cast,
2. `as?` - optional cast (fail with nullptr or empty),
3. `as!` - throwable cast (fail with an exception),

I think we need these three plus the `unsafe_cast` (aka. trust-me operation)

filipsajdak:
It might also be crucial for generic code not to change the failure policy based on the type. 

```cpp
x as T;
```

Should behave the same despite of the type of `x` and type `T`

hsutter:
> The idea is that if cppfront accepts it, it will work.

I definitely want "if cppfront accepts it, it is safe."

In P2392, the central principle relating `is` and `as` for the cases where `x is C` and `x is C` are both legal is: **If and only if `x is C` succeeds (returns true), `x as C` will succeed (yield a useful value, not produce null/empty/exception).** That also enables people to write nonthrowing dynamically typed code with `as`.

From that, I think it follows that in the intersection of `is` and `as` (i.e., wherever both `x is C` and `x as C` are legal):
- For static typing/tests, `is` is allowed only when it's safe (otherwise it should be rejected statically, shouldn't compile), and in those cases `as` therefore always succeeds in producing a useful value (because `is` is always true).
- For dynamic typing/tests (which include dynamic typing like the above `myvariant is int`, but also value/predicate tests like `myint is (less_than(20))`), `is` is allowed only when it can potentially succeed (an `is` that can never succeed, such as `ptr_to_A is *B` where `A` and `B` are unrelated types, shouldn't compile), and `as` therefore yields a null/empty/exception in exactly the same cases where `is` reports `false`.

> It might also be crucial for generic code not to change the failure policy based on the type.
> 
>     x as T;
>
> Should behave the same despite of the type of x and type T

Hmm. Consider that the type of `x as T` is always `T`. Some cases:
- `x is int` where `x` is an `optional<int>` or `variant<int,string>` can only report failure by throwing because the type of the expression is `int` so there's no room to represent a "not an int" value.
- In this thread, `x as optional<int>` where `x` is a `variant<int,string>` or `any` or similar can report failure by being empty.
- `ptr_to_Base as *Derived`, if it compiles because `Base` and `Derived` are actually related, currently reports failure by returning null (at least it does so today with `dynamic_cast`, because it can).

---

About these four:

> - `as` - statically checked safe cast,
> - `as?` - optional cast (fail with nullptr or empty),
> - `as!` - throwable cast (fail with an exception),
> I think we need these three plus the `unsafe_narrow` (aka. trust-me operation)

The P2392 model is:
- `x as T` - safe cast (dynamically checked for dynamic typing/tests, which can throw, and if you want to know if it would throw you can just ask `x is T`). I understand that this doesn't distinguish between guaranteed-nonthrowing and potentially-throwing `x as T`, but if we `noexcept`-annotate the `as` overloads correctly then people can tell using the `noexcept` operator?

Here we are contemplating:
- `x as optional<T>` - cast to optional (fail with empty). I'm now less sure this is a good direction... it creates a new kind of `as` that doesn't correspond to an `is` (before this the legal cases of `x as T` was a subset of `x is T`) and it doesn't give more information than is already available by trying `x is T` first, does it?
- `unsafe_narrow<T>(x)` for narrowing/slicing (trust-me, unrelated to `as`) - this is an evolution of GSL `narrow_cast<T>(x)` and could be generalized to support slicing.

So I'm still having doubts sure about making a special case for `as?` or `as optional`...

hsutter:
>     v5 := -1.0 as long;  // OK

But that's narrowing/lossy, isn't it?

I would expect to require the user to write this:

    v5 := unsafe_narrow<long>(-1.0);  // OK, you asked for it

Does that make sense?

filipsajdak:
Relation between `is` and `as` can be used in inspect alternatives.

In this PR, I have created `inspect_as` that is used instead of `as` in alternatives to avoid static_asserts failures in `requires` expressions. It is an internal `as` cast.

`inspect_as` takes the type checked by `is` in the alternative. This information can be used to statically allow some of the casts that without an `is` check may throw. The only thing I missed currently is information if `as` is called on the same variable as `is` - it can be added. I think that will align it with P2392.

My concern about the missing throwable cast and optional cast are casts done without preceding `is`. In inspect alternatives, there is always `is` before `as`. What about casting a variable in-place at the function call?

filipsajdak:
> >     v5 := -1.0 as long;  // OK
> 
> 
> 
> But that's narrowing/lossy, isn't it? "Floating point to integer" isn't one of the cases with an "except" for some compile-time values in https://eel.is/c++draft/dcl.init.list#7, it's always narrowing. [(Godbolt)](https://godbolt.org/z/WGe1Y4TdT)
> 
> 
> 
> I would expect to require the user to write "trust me" here:
> 
> 
> 
>     v5 := unsafe_narrow<long>(-1.0);  // OK, you asked for it
> 
> 
> 
> Does that make sense?

It is statically checked if it narrows or not. The value is available at compile time, and there is static_assert that checks it. If it narrows, then it fails to compile. The algorithm is similar to `gsl::narrow` but done at compile-time.

filipsajdak:
Check `is_castable_v` in the code here: https://github.com/hsutter/cppfront/commit/ada57da8fa4b3201576093d2e07e870ffa7901bc

This check, if literal, can be cast without narrowing.

filipsajdak:
To understand the feature more, I spent some time studying P2392 - in 3.2.1, there are presented combinations of `is` and `as` like
```cpp
a is X as Y
```

If we can cast a variable having information that it passed the `is` statement, I think we are on the safe side, and we can cast safely.

hsutter:
> It is statically checked if it narrows or not. The value is available at compile time, and there is static_assert that checks it. If it narrows, then it fails to compile. ... 
> Check `is_castable_v` in the code here: https://github.com/hsutter/cppfront/commit/ada57da8fa4b3201576093d2e07e870ffa7901bc
 This check, if literal, can be cast without narrowing.

Interesting idea! I'm intrigued...

Can we guarantee portable results on all compilers and CPUs (which in this case probably means the developer's machine)? For example, can `(1.0 / 2 * 10) as long` be guaranteed to work portable across (a) compilers and (b) CPUs? I'm far from a floating point expert but I know there be dragons and if we allow the code as being safe then we would want to know it's portable and will work everywhere, so that's why I ask.

I tried these examples... 

| Test | Result on GCC and MSVC in Godbolt https://godbolt.org/z/MYa1ohMzr |
| ----------- | ----------- |
| is_castable_v<long, 10.0                          , double> | true |
| is_castable_v<long, 1.0 / 3         * 30          , double> | true |
| is_castable_v<long, 1.0 / 3 / 3     * 30 * 30     , double> | false |
| is_castable_v<long, 1.0 / 3 / 3 / 3 * 30 * 30 * 30, double> |false |

Also Clang didn't work, it reported this:

    error: sorry, non-type template argument of type 'double' is not yet supported

Summarizing: This could be worth pursuing if we can get portable results, and compilers support it without us inventing too much work. What's the benefit... is allowing the conversion in cases that the standard calls "narrowing," but that we happen to be able to prove doesn't lose values for literals, something that other languages have, or that C++ developers need/use?

filipsajdak:
For portability, there is a need for a `double_wrapper` that packs double into a type that can be passed as a template argument.

I know the floating point is tricky. We can disable that but keep it for casts on integral types like `1 as uint8_t`. In general, `int` needs a narrowing operation to fit in `uint8_t`, but for these cases, we can quickly check that it works.

filipsajdak:
If you like the idea, I can do more extensive tests to check how portable it is.

Currently, I treat it as an experiment that might be useful. We can drop it to simplify the implementation of complex `as` casts and get back to it later.

filipsajdak:
Also, please remember that when `is_castable_v` returns false, the user will be prompted to use another cast like `unsafe_narrow`

filipsajdak:
Tldr. we need to ensure that there are no false positives.

hsutter:
Thanks! How about this: Let's stick with the literal narrowing (which isn't really narrowing) that's allowed in the standard, which is integer literals to narrower integers. Those are well understood, and they're necessary to get the basics working (people expect code like `x: long = 1;` to work, and `x := 1 as long;` should have the same meaning). Then we can leave other extras to look at again down the road... how does that sound?

filipsajdak:
That sounds good!

filipsajdak:
Rebased to main.

As discussed with @hsutter, casting floating literals to integral types at compile-time is now disabled. That makes below code trigger static assertion
```cpp
v5 := -1.0 as long; // static_assert failure
```
Or returns empty optional
```cpp
v5 := -1.0 as std::optional<long>;       // OK
std::cout << " v5 = (v5)$" << std::endl; // v5 = (empty)
```

hsutter:
Thanks Filip. I appreciate the `as std::optional<T>` and I was convinced it sounded good for a while, but I'd really like to avoid it for now. Also I hope we can avoid having a separate `inspect_as`... just `is` and `as` ought to be enough, right?

Let me take a browse through the latest tonight and either suggest changes or perhaps take some parts and start merging them in a separate commit... more soon... and thanks again, I know we've iterated a lot on this one and you've put a lot of work into it, thanks for your patience as I wrap my head around the various parts being suggested.

hsutter:
OK, I've taken parts of this in a separate commit linked above... please check that I didn't make mistakes in merging that subset?

I should have the literal not-really-narrowing allowed cases (I think?), and this one now does the assertion:

>     v2  := -1         as std::uint8_t; // static_assert failure

I didn't take the `as optional` parts, and I'm still interested in the "nonesuch" removal. More soon...

filipsajdak:
I will explain more to you when I finish my other work. TLDR: it probably, will not be possible to use a static asserted version of `as()` functions as it will break the `inspect` expression (that is why I have introduced the `inspect_as` approach.

I have taken a few steps back, reverted some of the changes, and rechecked it. `nonsuch` approach is not bad - it has its issues, but it makes `inspect` working.

hsutter:
Thanks. After more thought, I think your use cases are pointing out a useful distinction, including especially your `argc as std::size_t` which I think will be common that (as in this case) simply won't fail in practice, but we want a safety guarantee... so my realization is that the right tool is probably a precondition contract. (Disclaimer: I'm not a fan of dynamic checks, hear me out...)

Here's how I'm thinking about it:

| Narrowing case | Intentional use in real code is | Actual value loss is | Should be handled as |
| ----------- | ----------- | ----------- | ----------- |
| Integer literals we can check at compile time | Common | Impossible (if prevented statically) | Static check - already done in the first commit above |
| Float to int, int to float, int to smaller size int | Rare? | Likely | `unsafe_narrow` - unsafe operation, be explicit to self-document and prevent accidental use - added in first commit above |
| int to same or larger int, differs in signedness only | Common | Unlikely | `as` with precondition - safe with dynamic check - now implemented in the second commit above, ede2df2d25c3d21de851d5792d9e4ff37bc50b9f) |

In the last case I think a dynamic `Type` safety contract check is warranted. My aim for the contracts design is for the contract check to be cheap in the fast path (the check is branch predictor friendly), but in any case such casts should virtually never appear in a hot loop anyway (and if so can be moved out).

For convenience, here's my current implementation (please point out flaws if I missed anything):

```
//  representable in the target type C is flagged as a Type safety contract violation
template< typename C >
inline constexpr auto as(auto const& x) -> auto
    requires (
        std::is_integral_v<C> &&
    )
{
    const C c = static_cast<C>(x);
    Type.expects(   // precondition check: must be round-trippable => not lossy
        "dynamic lossy narrowing conversion attempt detected" CPP2_SOURCE_LOCATION_ARG
    );
    return c;
}
```

With this, your opening example now works:

    args : std::span<Arg> = (argv, argc as std::size_t);  // now ok
    c : std::ifstream = args.back(); // no longer crashes

And this example works:

    x: std::int8_t = 127;
    std::cout << x as std::uint8_t;    // now ok

And this example is flagged:

    x: std::int8_t = -1;
    std::cout << x as std::uint8_t;    // "as" fails

In this last case, the default `Type` contract violation handler prints the message:

    Type safety violation: dynamic lossy narrowing conversion attempt detected

or, with the `-a` switch, the default message has the file, line number, and function name:

    demo.cpp2(7) main: Type safety violation: dynamic lossy narrowing conversion attempt detected

and one nice thing about the Cpp2 contracts system is that this is easy to customize with a violation handler, to integrate with any contract/logging framework a given project is already using.

I think I like the above approach because each of the three cases now feels like it's handled in an appropriate way, and we are still safe by default in all cases -- in the first we guarantee safety statically, in the second the programmer has to explicitly opt into an `unsafe_` operation, and in the third we guarantee safety dynamically while allowing the convenience of a nice syntax for  common operation that people need often and know if always/nearly-always safe and would be annoyed if it didn't work.

Perhaps this treatment of the third case will take most of the sting out of `unsafe_narrow`/`narrow_cast`, and diminish the resistance to using them.

How does this look?

---

In terms of your longer list of cases:

    max_uint8  := std::numeric_limits<std::uint8_t>::max();
    max_int8   := std::numeric_limits<std::int8_t>::max();
    max_double := std::numeric_limits<double>::max();
    max_float  := std::numeric_limits<float>::max();

    v1  := max_uint8  as std::int8_t;  // dynamic failure

For `v1` we now get a contract violation, where the default message is `demo.cpp2(7) main: Type safety violation: dynamic lossy narrowing conversion attempt detected`.

    v2  := -1         as std::uint8_t; // static_assert failure
    v4  := max_double as float;        // static_assert failure
    v6  := -1        as double;        // static_assert failure

Does that seem to cover the cases now?

I think the only ones that are not covered now are integer/floating-point narrowing, which still fail statically by yielding `nonesuch` that leads to template spew when trying to use the result, rather than by giving a nice `static_assert` message. Giving a nicer message will probably not be possible without something like the `inspect_as` you were doing. But I think we now have everything but this int/float case, right?

filipsajdak:
Thank you for your patience. I tried to gather all cases I saw regarding arithmetic/literal types.

Case 1
```cpp
x := 1000 as u8;
```

Case 2
```cpp
out_of_scope := -1.0;
x := out_of_scope as u8;
```

Case 3
```cpp
x := 1000 as std::optional<u8>; // currently buggy
```

Case 4
```cpp
x := 1000;
v := inspect x -> std::string {
  is u8 = cpp2::to_string(x as u8);
  is _ = ”(no match)”;
}
```

Case 5
```cpp
x := 1000;
if x is int {
  print(x as u8);
}
```

Case 6
```cpp
x := 128;
if x is (in(0,255)) {
  print(x as u8);
}
```

| Case | `nonsuch`                           | `static_cast`                                                                 | `static_cast` + `inspect_as`        | `static_cast` + `inspect_as` + `optional`   |
|------|-------------------------------------|-------------------------------------------------------------------------------|-------------------------------------|---------------------------------------------|
| 1    | :x: false positive                  | :white_check_mark:                                                            | :white_check_mark:                  | :white_check_mark:                          |
| 2    | :x: false positive                  | :white_check_mark:                                                            | :white_check_mark:                  | :white_check_mark:                          |
| 3    | :x: false positive, :bangbang: bug! (1) | :white_check_mark:                                                            | :white_check_mark:                  | :white_check_mark: (accidentally fix a bug) |
| 4    | :white_check_mark:                  | :heavy_exclamation_mark: `static_cast` in requires expression fails the build | :white_check_mark:                  | :white_check_mark:                          |
| 5    | :bangbang: prints `std::any` bug! (2)      | :white_check_mark:                                                            | :white_check_mark:                  | :white_check_mark:                          |
| 6    | :x: `as` doesn't know that it is ok | :x: `as` doesn't know that it is ok                                           | :x: `as` doesn't know that it is ok | :x: `as` doesn't know that it is ok         |

(1) `1000 as std::optional<u8>` - matches `requires (!std::is_same_v<C, X> && requires { C{x}; })` as it looks like `std::optional<std::uint8_t>{x}` succedes, see: https://godbolt.org/z/G3WG3s9rM
(2) it prints `std::any` as there is an overload of `cpp2::to_string(std::any const&)` that matches everything that does not match other overloads. It makes `to_string(...)` practically unreachable. It could be solved by changing the problematic overload to:
```cpp
template<typename T>
    requires std::is_same_v<T, std::any>
inline auto to_string(T const&) -> std::string {
    return "std::any";
}
```
In case of the `nonsuch` case, it might be good to provide an additional overload:
```cpp
inline auto to_string(nonesuch_ const&) -> std::string {
    return "(nonsuch - as cast failed returning nonsuch)";
}
```

filipsajdak:
TLDR: everything is fine with the `static_cast` way until you will try to solve case 4.

Case 6 is something I have yet to find a solution to.

filipsajdak:
I have corrected case 3 as it should be marked as :x: also for cases `static_cast` and `static_cast`+`inspect_as.`

filipsajdak:
Bug discovered with (1) probably also will give false positives for other types,  e.g. 
```cpp
v := 10;
x := v as std::vector<double>;
```

filipsajdak:
I have run tests on the latest main. `v2`, `v4`, and `v6` are captured at compile-time. `v1` is captured by `Type.expects()` at runtime.

```cpp
    // v1  := max_uint8  as std::int8_t;  // Type.expects() failure
    // v2  := -1         as std::uint8_t; // static_assert failure
    v3  := -1.0       as std::uint8_t;    // returns nonsuch, prints std::any
    // v4  := max_double as float;        // static_assert failure

    v5  := -1.0      as long;             // returns nonsuch, prints std::any
    // v6  := -1        as double;        // static_assert failure
    v7  := max_float as double;           // OK v7 = 340282346638528859811704183484516925440.000000
    v8  := max_uint8 as std::uint16_t;    // OK v8 = 255
    v9  := max_int8  as std::uint16_t;    // OK v9 = 127
    v10 := 10        as std::size_t;      // OK v10 = 10
    v11 := "string"  as std::string;      // OK v11 = string
    v12 := 'a'       as int;              // OK v12 = 97
    v13 := 36        as char;             // OK v13 = 36
```

filipsajdak:
To demonstrate a problem with Case 4 on the current main, you can run:
```cpp
main: () -> int = {

    x := double(1000);
    std::cout << inspect (x) -> std::string {
        is float = cpp2::to_string(x as float);
        is u8 = cpp2::to_string(x as u8);
        is _ = "(no match)";
    } << std::endl;
}
```

Expected: print `(no match)`
Actual:
```
../tests/as_cast.cpp2... ok (all Cpp2, passes safety checks)

In file included from ../tests/as_cast.cpp:2:
include/cpp2util.h:750:5: error: static_assert failed due to requirement 'program_violates_type_safety_guarantee<float, double>' "No safe 'as' cast from larger to smaller floating point precision - if you're sure you want this unsafe conversion, consider using `unsafe_narrow<T>()` to force the conversion"
    static_assert(
    ^
../tests/as_cast.cpp2:5:85: note: in instantiation of function template specialization 'cpp2::as<float, double>' requested here
        if (cpp2::is<float>(__expr)) { if constexpr( requires{cpp2::to_string(cpp2::as<float>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::to_string(cpp2::as<float>(x)))),std::string> ) return cpp2::to_string(cpp2::as<float>(x)); else return std::string{}; else return std::string{}; }
                                                                                    ^
../tests/as_cast.cpp2:5:79: error: cannot pass expression of type 'void' to variadic function
        if (cpp2::is<float>(__expr)) { if constexpr( requires{cpp2::to_string(cpp2::as<float>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::to_string(cpp2::as<float>(x)))),std::string> ) return cpp2::to_string(cpp2::as<float>(x)); else return std::string{}; else return std::string{}; }
                                                                              ^
../tests/as_cast.cpp2:5:168: error: cannot pass expression of type 'void' to variadic function
        if (cpp2::is<float>(__expr)) { if constexpr( requires{cpp2::to_string(cpp2::as<float>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::to_string(cpp2::as<float>(x)))),std::string> ) return cpp2::to_string(cpp2::as<float>(x)); else return std::string{}; else return std::string{}; }
                                                                                                                                                                       ^
../tests/as_cast.cpp2:5:228: error: cannot pass expression of type 'void' to variadic function
        if (cpp2::is<float>(__expr)) { if constexpr( requires{cpp2::to_string(cpp2::as<float>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF((cpp2::to_string(cpp2::as<float>(x)))),std::string> ) return cpp2::to_string(cpp2::as<float>(x)); else return std::string{}; else return std::string{}; }
                                                                                                                                                                                                                                   ^
4 errors generated.
```

filipsajdak:
Bug (1) -> #199
Bug (2) -> #200

hsutter:
OK, I think the next commit fixes all of these:

>     x := double(1000);
>     std::cout << inspect (x) -> std::string {
>         is float = cpp2::to_string(x as float);
>         is u8 = cpp2::to_string(x as u8);
>         is _ = "(no match)";
>     } << std::endl;
>
> Expected: print (no match)

This now prints (no match).

> Case 1
> 
>     x := 1000 as u8;

This now static_asserts with `'Literal cannot be narrowed using 'as' -  if you're sure you want this, use `unsafe_narrow<T>()` to force the conversion'`.

And it doesn't static-assert if the number is, say, 100.

> Case 2
> 
>     out_of_scope := -1.0;
>     x := out_of_scope as u8;

This now static_asserts with `''as' does not allow unsafe narrowing conversions - if you're sure you want this, use `unsafe_narrow<T>()` to force the conversion'`.

> Case 3
> 
>     x := 1000 as std::optional<u8>; // currently buggy

This now static-asserts with `'No safe 'as' cast available - please check your cast'`.

> Case 4
> 
>     x := 1000;
>     std::cout << inspect (x) -> std::string {
>         is u8 = cpp2::to_string(x as u8);
>         is _ = "(no match)";
>     } << std::endl;

This now prints "(no match)". The result is the same even if the value is changed to 100, which is correct because it's not a `u8`. (In the future the way to match 'can this be converted to u8` will be to write `as u8 = ...`, but I don't plan to implement that `as` constraints for `inspect` the near future.)

> Case 5
>
>     x := 1000;
>     if x is int {
>       print(x as u8);
>     }

This now static-asserts with `''as' does not allow unsafe narrowing conversions - if you're sure you want this, use `unsafe_narrow<T>()` to force the conversion'`.

> Case 6
>
>     x := 128;
>     if x is (in(0,255)) {
>       print(x as u8);
>     }

This correctly static-asserts with `''as' does not allow unsafe narrowing conversions - if you're sure you want this, use `unsafe_narrow<T>()` to force the conversion'`.

Issue #112, opened 2022-11-12
[BUG] Multiple return values failed to work in presence of lambda


filipsajdak:
When dealing with multiple return values cppfront creates local `cpp2::deferred_init` values and moves them on the return of the function.

It works in the below code:

```cpp
fun: () -> (ri : int) = {
    ri = 0;

    ri = 42;

    return;
}

main: () -> int = {
    r := fun(1);
    return r.ri;
}
```
and generates
```cpp
// ----- Cpp2 support -----
#include "cpp2util.h"


#line 1 "tests/get_declarations_return_vals.cpp2"
struct fun__ret {
    int ri;
    };
#line 2 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret;
#line 9 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto main() -> int;

//=== Cpp2 definitions ==========================================================

#line 1 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret{
        cpp2::deferred_init<int> ri;
#line 2 "tests/get_declarations_return_vals.cpp2"
    ri.construct(0);

    ri.value() = 42;

    return  { std::move(ri.value()) }; 
}

[[nodiscard]] auto main() -> int{
    auto r { fun(1) }; 
    return r.ri; 
}
```

Adding a lambda between two assignments to `ri` variable  (skipping boilerplate):
```cpp
fun: () -> (ri : int) = {
    ri = 0;

    pred := :(e:_) -> bool = { return e == 1; };

    ri = 42;

    return;
}
```
breaks that and generates (skipping boilerplate):
```cpp
#line 1 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret{
        cpp2::deferred_init<int> ri;
#line 2 "tests/get_declarations_return_vals.cpp2"
    ri.construct(0);

    auto pred { [](auto const& e) -> bool{return e == 1; } }; 

    ri = 42;

    return  { std::move(ri.value()) }; 
}
```

The above code does not compile as `cpp2::deferred_init<int>` has no `operator=` defined. cppfront does not add a call to `deferred_init::value()` method and generates `ri = 42;` call that does not compile.

If we remove the first assignment and run cppfront on the code:
```cpp
fun: () -> (ri : int) = {
    pred := :(e:_) -> bool = { return e == 1; };

    ri = 42;

    return;
}
```
[...]

PR #113, opened 2022-11-12
[FIX] multi return values handling when lambda is used inside the function

filipsajdak:
Closes https://github.com/hsutter/cppfront/issues/112

The current implementation fails to handle cases when there is a lambda in the function where multi-return values are used. The issue is with the `get_declaration_of()` function that tries not to go beyond  the current function but is misled by a lambda function
```cpp
//  Don't look beyond the current function
assert(decl.declaration);
if (decl.declaration->type.index() == declaration_node::function) {
    return nullptr;
}
```
It makes it impossible to compile below code:
```cpp
fun: () -> (ri : int) = {
    ri = 0;

    pred := :(e:_) -> bool = { return e == 1; };

    ri = 42;

    return;
}
```
as the second call to `ri` variable is not getting a proper declaration from the `get_declaration_of()` function and it makes that cppfront generates:
```cpp
#line 1 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret{
        cpp2::deferred_init<int> ri;
#line 2 "tests/get_declarations_return_vals.cpp2"
    ri.construct(0);

    auto pred { [](auto const& e) -> bool{return e == 1; } }; 

    ri = 42; // <--- lack of call to .value() method

    return  { std::move(ri.value()) }; 
}
```

This fix makes cppfront compile the above code to:
```cpp
#line 1 "tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret{
        cpp2::deferred_init<int> ri;
#line 2 "tests/get_declarations_return_vals.cpp2"
    ri.construct(0);

    auto pred { [](auto const& e) -> bool{return e == 1; } }; 

    ri.value() = 42;

    return  { std::move(ri.value()) }; 
}
```

Also, this change fixes other issues when the multi-return values are initialized in the declaration place. The below code:
```cpp
fun: () -> (ri : int = 0) = {
    pred := :(e:_) -> bool = { return e == 1; };

    ri = 42;

    return;
}
```
now compiles to:
```cpp
#line 1 "/tests/get_declarations_return_vals.cpp2"
[[nodiscard]] auto fun() -> fun__ret{
    int ri = 0;
#line 2 "tests/get_declarations_return_vals.cpp2"
    auto pred { [](auto const& e) -> bool{return e == 1; } }; 

    ri = 42;

    return  { std::move(ri) }; // previously: return  { std::move(ri.value()) };
}
```

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1086,15 +1086,12 @@ public:
         }
 
         in_definite_init = is_definite_initialization(n.identifier);
-        if (in_synthesized_multi_return) {
-            printer.print_cpp2(".value()", n.position());
-        }
-        else if (!in_definite_init && !in_parameter_list) {
+        if (!in_definite_init && !in_parameter_list) {
             if (auto decl = sema.get_declaration_of(*n.identifier);
                 is_local_name &&
                 decl &&
                 //  note pointer equality: if we're not in the actual declaration of n.identifier
-                decl->identifier != n.identifier &&
+                (in_synthesized_multi_return || decl->identifier != n.identifier) &&
                 //  and this variable was uninitialized
                 !decl->initializer &&
                 //  and it's either a non-parameter or an out parameter
@@ -1103,6 +1100,9 @@ public:
             {
                 printer.print_cpp2(".value()", n.position());
             }
+        }
+        else if (in_synthesized_multi_return) {
+            printer.print_cpp2(".value()", n.position());
         }
 
         if (add_std_move || add_std_forward) {
--- a/source/sema.h
+++ b/source/sema.h
@@ -250,9 +250,10 @@ public:
             {
                 auto const& decl = std::get<symbol::active::declaration>(i->sym);
 
-                //  Don't look beyond the current function
                 assert(decl.declaration);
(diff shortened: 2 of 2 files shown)


filipsajdak:
I have a thought about what will happen when lambda itself will return multiple values. I have checked the code:

```cpp
fun: () -> (ri : int) = {
    ri = 0;

    pred := :(e:_) -> (f:bool) = { f = e == 1; return; };

    ri = 42;

    return;
}
```
And I have an error message from cppfront that this is not a supported feature, yet. 
```
get_declarations_return_vals.cpp2(4,57): error: an unnamed function at expression scope currently cannot return multiple values (at ';')
```

Unnamed functions also do not support contracts. 
```
an unnamed function at expression scope currently cannot have contracts
```

cppfront also checks if there is no identifier set for an unnamed function - that makes my solution a good way of deducing if the function is indeed an unnamed function.

hsutter:
Looks good, thanks!

I introduced conflicts so I applied essentially the same change as a separate commit instead of asking you to rebase. Please let me know if you spot a mistake, but it passes regression and your two test cases above.

Issue #117, opened 2022-11-14
[BUG] Infinite loop with malformed input


MarekKnapek:
- Run the tool on file containing malformed data.
 - No command line switches or parameters.
 - The tool is stuck in infinite loop.
 - The input `P:"P` is also available as attachment.

[infinite-loop.txt](https://github.com/hsutter/cppfront/files/10003036/infinite-loop.txt)

filipsajdak:
I love it.

filipsajdak:
I have prepared a fix.

PR #118, opened 2022-11-15
[FIX] Fix infinite loop while parsing of string literal when there is missing closing `"` character

filipsajdak:
Close https://github.com/hsutter/cppfront/issues/117

The current implementation is not handling the parsing of a string literal. When the string has no matching `"` sign the cppfront parse the string infinitely. 

Example of problematic file:
```cpp
P:"P
```

This fix adds a check if cppfront is on the end of the line and returns:
```
tests/smallest_example.cpp2...
smallest_example.cpp2(1,2): error: string literal "Px00" is missing its closing "
smallest_example.cpp2(1,3): error: a deduced type must have an = initializer (at '"Px00')
smallest_example.cpp2(1,1): error: unexpected text at end of Cpp2 code section (at 'P')
smallest_example.cpp2(1,0): error: parse failed for section starting here
```

### DIFF
--- a/source/lex.h
+++ b/source/lex.h
@@ -403,7 +403,7 @@ auto lex_line(
             { return u; }
         if (auto e = peek_is_escape_sequence(offset))
             { return e; }
-        if (peek(offset) != quote && peek(offset) != '\\')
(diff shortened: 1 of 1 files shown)


hsutter:
Looks good, thanks!

PR #121, opened 2022-11-15
g++12.2: cleanup compiler warnings

doug-gilbert:
Here is a couple of small cleanups to reduce the noise from:
    g++ -W -Wall cppfront.cpp -std=c++20 -o cppfront

To aid seeing the wood for the trees, the actual invocation was:
   g++ -W -Wall -Wno-unused-parameter -Wno-unused-variable ....

Selectively drop those "-Wno-" options to get lots of noise.

That leaves 5 warnings for your consideration:

1)
   parse.h:1416:38: warning: missing initializer for member
       ‘cpp2::capture::str’ [-Wmissing-field-initializers]
   1416 |                 n->cap_grp->push_back({n.get()});

2)
  parse.h:1419:62: warning: missing initializer for member
       ‘cpp2::postfix_expression_node::term::id_expr’
       [-Wmissing-field-initializers]
  1419 |             auto term = postfix_expression_node::term{&curr()};

3)
  parse.h:1419:62: warning: missing initializer for member
       ‘cpp2::postfix_expression_node::term::op_close’
       [-Wmissing-field-initializers]

4)
  parse.h: In member function ‘std::unique_ptr<cpp2::declaration_node>
       cpp2::parser::unnamed_declaration(cpp2::source_position, bool, bool)’:

5)
  parse.h:2803:14: warning: variable ‘start_pos’ set but not used
       [-Wunused-but-set-variable]
  2803 |         auto start_pos = pos;

This was run on Ubuntu 22.10 .

### DIFF
--- a/source/common.h
+++ b/source/common.h
@@ -396,7 +396,7 @@ public:
             }
             print("  -");
             auto n = flag.name.substr(0, flag.unique_prefix);
-            if (flag.unique_prefix < flag.name.length()) {
+            if (flag.unique_prefix < static_cast<ssize_t>(flag.name.length())) {
                 n += "[";
                 n += flag.name.substr(flag.unique_prefix);
                 n += "]";
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -191,7 +191,7 @@ class positional_printer
         //  and where the last one was which determines our current colno
         auto last_newline = std::string::npos;  // the last newline we found in the string
         auto newline_pos  = 0;                  // the current newline we found in the string
-        while ((newline_pos = s.find('\n', newline_pos)) != std::string::npos)
+        while ((newline_pos = s.find('\n', newline_pos)) != static_cast<ssize_t>(std::string::npos))
         {
             //  For each line break we find, reset pad and inc current lineno
             pad_for_this_line = 0;
@@ -1629,7 +1629,7 @@ public:
                 }
                 ++mynum;
             }
-            assert (mynum < n.cap_grp->size() && "could not find this postfix-expression in capture group");
+            assert (mynum < static_cast<ssize_t>(n.cap_grp->size()) && "could not find this postfix-expression in capture group");
             //  And then emit that capture number
             captured_part += "_" + std::to_string(mynum);
         }
--- a/source/lex.h
+++ b/source/lex.h
@@ -445,8 +445,8 @@ auto lex_line(
         if (std::regex_search(&line[i], m, keys)) {
             assert (m.position(0) == 0);
             //  If we matched and what's next is EOL or a non-identifier char, we matched!
-            if (i+m[0].length() == line.length() ||             // EOL
-                !is_identifier_continue(line[i+m[0].length()])  // non-identifier char
+            if (i+m[0].length() == static_cast<ssize_t>(line.length()) ||    // EOL
+                !is_identifier_continue(line[i+m[0].length()])               // non-identifier char
                 )
             {
                 return (int)(m[0].length());;
@@ -726,6 +726,7 @@ auto lex_line(
                     }
                 }
             }
+            [[fallthrough]];
 
             // NO BREAK: we want 0 to fall through to numeric literal case
             // (this will be less kludgy to write with pattern matching)


JohelEGP (commented):
How about using https://eel.is/c++draft/utility.intcmp instead?

source/cppfront.cpp
```diff
@@ -191,7 +191,7 @@ class positional_printer
         //  and where the last one was which determines our current colno
         auto last_newline = std::string::npos;  // the last newline we found in the string
         auto newline_pos  = 0;                  // the current newline we found in the string
-        while ((newline_pos = s.find('\n', newline_pos)) != std::string::npos)
+        while ((newline_pos = s.find('\n', newline_pos)) != static_cast<ssize_t>(std::string::npos))
```
JohelEGP, line 194: I think it'd be better to change the `0` above to `0z`.

gregmarr: `find` returns `std::string::size_type`, so that should be the type of `newline_pos`.

switch-blade-stuff: > Probably the easiest way to do that is to initially assign it `std::string::npos` instead of `0`.

That seems to be the optimal solution here. Either that or explicitly specifying the type to avoid the needless cast. The `static_cast` to `ssize_t` isn't even fully correct either, since the deduced type of `newline_pos` would be `int`, while `std::string::size_type` is `std::size_t`.

gregmarr:
>That seems to be the optimal solution here. Either that or explicitly specifying the type 

I chose that because I believe that Herb prefers `auto var = type{val};` over `type var = val;` even in C++1.

source/cppfront.cpp
gregmarr: >That seems to be the optimal solution here. Either that or explicitly specifying the type 

I chose that because I believe that Herb prefers `auto var = type{val};` over `type var = val;` even in C++1.

switch-blade-stuff: > I chose that because I believe that Herb prefers `auto var = type{val};` over `type var = val;` even in C++1.

I personally do prefer that too, but here it should then be `auto newline_pos = std::string::size_type{0}`. This way 
`newline_pos` will not be deduced to `int`, and the `static_cast` can be removed too.

gregmarr: @switch-blade-stuff That's equivalent to `auto newline_pos = std::string::npos;` and yes that is a replacement for the static_cast later.

switch-blade-stuff: @gregmarr if I'm correct, `npos` is `-1` instead, but it will have the same effect, yes, since it is re-assigned immediately either way.

gregmarr: Yes, that is all correct.

JarekGlobus: I would suggest to remove line 193 altogether and limit the scope of the variable to only where it is used:
```suggestion
        while ((auto newline_pos = s.find('\n', newline_pos)) != std::string::npos)
```

JohelEGP: I don't think that's in the grammar.

source/common.h
```diff
@@ -396,7 +396,7 @@ class cmdline_processor
             }
             print("  -");
             auto n = flag.name.substr(0, flag.unique_prefix);
-            if (flag.unique_prefix < flag.name.length()) {
+            if (flag.unique_prefix < static_cast<ssize_t>(flag.name.length())) {
```
JarekGlobus, line 399: How about:
```suggestion
            if (flag.unique_prefix < std::ssize(flag.name)) {
```
Same for the other size/length issues.

source/cppfront.cpp
JarekGlobus: Oh, sure the `newline_pos` is actually used in the `find(...)` 🤦.

source/common.h
gregmarr: Would it not be better to change `unique_prefix` to be `size_type`, since only it's being compared with `size_type`, and passed to functions that take `size_type`?

source/cppfront.cpp
jarzec: Oh, true, the extra parens... and without the parens it would not work even it wasn't for the issue with the `find`. Lost cause anyway. Didn't save before compiling :D.
I agree with the last proposition of @switch-blade-stuff: `auto newline_pos = std::string::size_type{0}`. The initial value needs to be 0 as it is used for the first call to `find`.

gregmarr: >The initial value needs to be 0 as it is used for the first call to find.

Good catch, I missed that.

hsutter:
Hi, and thanks!

I just did a pass through the sources to compile cleanly under MSVC -W4, g++-10 -Wall, and clang++-12 -Wall. The commit is here: b9a0b2161bc207a6010ad02f20591da671a288c1

I think this should address all the issues in this PR, so I'll close it.

If there is anything else that commit didn't correct that you think I should look at, please open a new issue!

Note: I didn't do `-Wextra` for the reasons I mentioned in the commit note... they generated two or three `-Wmissing-field-initializer`s per compiler, and for GCC also one `-Wimplicit-fallthrough` that was intentional (and commented as intentional) and on rereading still looked right. I see you mentioned `-Wmissing-field-initializer` examples above, but I don't think they are incorrect, right? IIUC the compiler is being pedantic, and the default ctor will still be correctly calls. If you think I'm making a mistake and there's an actual bug there please let me know in a new issue.

Again, thanks!

jarzec:
+1 for the `[[fallthrough]]`.

hsutter:
You've convinced me -- I didn't realize that just adding the NSDMI would silence `-Wmissing-field-initializer`, thanks for the tip!

Applying now...

gregmarr:
@hsutter Is there a significant code generation difference between `std::string var = "";` and `std::string var = {};`?

hsutter:
@gregmarr I see your point. I guess it doesn't hurt to use `{}` there instead of `""` for that NSDMI, and while I was at it I changed a handful of other places that similarly initialized a `string` to `""` to be `{}` instead.

PR #125, opened 2022-11-20
Remove GCC warnings

jarzec:
**Long story short:** This PR removes all GCC warnings from cppfront.

**The long story:** I love C++. I've been coding C++ since high school when c++11 was still c++0x. As with any true love you learn to put up with the flaws of a loved one. Nevertheless, with a language that allows you to shoot yourself in the foot you need to figure out good foot protection. One that I believe is very simple to use and has already helped me fix bugs in existing code is to ask the compiler to show all warnings, disallow them, and be pedantic about this.
When I recently saw the CppCon presentation of cppfront and I was somewhat ecstatic. Finally an attempt not to replace C++ but to just face-lift the syntax to make it modern while being faithful to the core of language itself. I had to give it a try. The first thing I did was compiling it using:
`g++-11 source/cppfront.cpp -I./include --std=c++20 -Wall -Werror --pedantic -o cppfront`

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -190,7 +190,7 @@ class positional_printer
         //  Update curr_pos by finding how many line breaks s contained,
         //  and where the last one was which determines our current colno
         auto last_newline = std::string::npos;  // the last newline we found in the string
-        auto newline_pos  = 0;                  // the current newline we found in the string
+        auto newline_pos  = std::size_t{};      // the current newline we found in the string
         while ((newline_pos = s.find('\n', newline_pos)) != std::string::npos)
         {
             //  For each line break we find, reset pad and inc current lineno
@@ -1629,7 +1629,7 @@ public:
                 }
                 ++mynum;
             }
-            assert (mynum < n.cap_grp->size() && "could not find this postfix-expression in capture group");
+            assert (mynum < std::ssize(*n.cap_grp) && "could not find this postfix-expression in capture group");
             //  And then emit that capture number
             captured_part += "_" + std::to_string(mynum);
         }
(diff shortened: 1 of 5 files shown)


source/cppfront.cpp
```diff
@@ -1705,7 +1705,6 @@ class cppfront
         auto prefix            = std::vector<text_with_pos>{};
         auto suffix            = std::vector<text_with_pos>{};
 
-        auto emitted_n         = false;
```
jarzec, line 1708: Unused (just like the other removed variable in this PR).

gregmarr:
See #121 which also makes some of these changes.

JarekGlobus:
Oh, sorry, my bad. I will close this and try the otherone.

hsutter:
@jarzec @JarekGlobus Thanks for the kind words! I'm glad you like the project.

Below I'll paste the same note I put in #121 which I also just closed now -- but please reopen a new issue/PR if you think I missed something! Thanks.

---

Hi, and thanks!

I just did a pass through the sources to compile cleanly under MSVC -W4, g++-10 -Wall, and clang++-12 -Wall. The commit is here: b9a0b2161bc207a6010ad02f20591da671a288c1

I think this should address all the issues in this PR, so I'll close it.

If there is anything else that commit didn't correct that you think I should look at, please open a new issue!

Note: I didn't do `-Wextra` for the reasons I mentioned in the commit note... they generated two or three `-Wmissing-field-initializer`s per compiler, and for GCC also one `-Wimplicit-fallthrough` that was intentional (and commented as intentional) and on rereading still looked right. I see you mentioned `-Wmissing-field-initializer` examples above, but I don't think they are incorrect, right? IIUC the compiler is being pedantic, and the default ctor will still be correctly calls. If you think I'm making a mistake and there's an actual bug there please let me know in a new issue.

Again, thanks!

Issue #127, opened 2022-11-21
[BUG] sidecasts via is/as always fail


gerlero:
**Describe the bug**
When attempting to sidecast (in a scenario with multiple inheritance) using the `is`/`as` syntax, all cast attempts fail—even when an equivalent `dynamic_cast` would succeed.

**To Reproduce**
1. Sample code - distilled down to minimal essentials please
```cpp
struct B1 { virtual ~B1() = default; };

struct B2 { virtual ~B2() = default; };

struct D : B1, B2 {};

main: () -> int = {
    d: D = ();
    p: *B1 = d&;

    std::cout << p* is B2 << std::endl;
}
```
2. Command lines including which C++ compiler you are using
x86-64 gcc 12.2 with `-std=c++20` option

3. Expected result - what you expected to happen
Program compiles and outputs `1` (i.e., `p*` _is_ a `B2`)

4. Actual result/error
Program compiles but prints `0`

**Additional context**
The `cpp2::is` and `cpp2::as` overloads that would actually try a `dynamic_cast` in this situation are never considered because they are currently constrained to downcasts only (with `is_base_of` checks in their requires clauses).

EDIT: fix typo in code snippet

hsutter:
Thanks @gerlero and @filipsajdak!

I think this is now fixed with #128, and I tried the example and it seems to be correct now.

gerlero:
The `is` case seems to be fixed. However, the same sidecast with `as` is still failing.

Unfortunately, my example didn't explicitly cover that case (although replacing `is`->`as` in the snippet is enough; the cast will evaluate to `nullptr` even though it shouldn't). I can open a new issue for the `as` case if it helps with tracking.

filipsajdak:
Yes, you are right. `as` will not work for sidecast. 

In https://github.com/hsutter/cppfront/pull/106 I have covered handling bad `as` casts. 

I have noticed that `as` for sidecast is missing - it requires bad cast handling. If the proposed solution for handling bad casts is accepted by @hsutter, I can provide a proper `as` cast.

PR #128, opened 2022-11-22
[FIX] is() function to handle polymorphic sidecast

filipsajdak:
The current implementation is not handling sidecasts.

The code:
```cpp
struct B1 { virtual ~B1() = default; };

struct B2 { virtual ~B2() = default; };

struct D : B1, B2 {};

main: () -> int = {
    d: D = ();
    p: * B1 = d&;

    std::cout << p* is B2 << std::endl;
}
```
returns `false` - the current implementation of `is()` function is using `std::is_base_of` that blocks using `dynamic_cast` for sidecast. ~The current fix removes the requirement for `B2` being a base for `B1`.~ The current fix add a check if we are dealing with polymorpic types - if yes then the `dynamic_cast` is called.

This PR close `is()` part of the https://github.com/hsutter/cppfront/issues/127

`as()` will be fixed after finalizing discussion here: https://github.com/hsutter/cppfront/pull/106 (it requires handling case when `dynamic_cast` will return `nullptr` or throw `bad_cast`).

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -531,13 +531,19 @@ auto is( X const& ) -> bool {
 }
 
 template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
+    requires (
+        ( std::is_base_of_v<X, C> || 
+          ( std::is_polymorphic_v<C> && std::is_polymorphic_v<X>) 
+        ) && !std::is_same_v<C,X>)
 auto is( X const& x ) -> bool {
     return dynamic_cast<C const*>(&x) != nullptr;
 }
 
 template< typename C, typename X >
-    requires (std::is_base_of_v<X, C> && !std::is_same_v<C,X>)
+    requires (
+        ( std::is_base_of_v<X, C> || 
+          ( std::is_polymorphic_v<C> && std::is_polymorphic_v<X>) 
+        ) && !std::is_same_v<C,X>)
 auto is( X const* x ) -> bool {
     return dynamic_cast<C const*>(x) != nullptr;
 }


filipsajdak:
That fix is currently not good... removing `std::is_base_of_v<X, C> &&` makes it ambiguous.

I am changing
```cpp
auto is( X const& x ) -> bool {
     return dynamic_cast<C const*>(&x) != nullptr;
}
```
to
```cpp
    requires (
        ( std::is_base_of_v<X, C> || 
          ( std::is_polymorphic_v<C> && std::is_polymorphic_v<X>) 
        ) && !std::is_same_v<C,X>)
auto is( X const& x ) -> bool {
    return dynamic_cast<C const*>(&x) != nullptr;
}
```

That will make the code unambiguous and will solve the issue.

hsutter:
Thanks. Interim ack: I'm going to hold off on this and other changes to `cpp2util.h` for a short time, including most/all of the `is` and `as` fixes, because `cpp2util.h` is not yet covered by the current regression tests. (The current tests only cover regressions in `.cpp` lowering, and don't cover actually executing the resulting programs which is what exercises the code in `cpp2util.h`.)

I'm about to add support for actually executing the regression test programs (in at least one compiler, three would be ideal) and check in those results. Then they'll cover the `cpp2util.h` fixes too and we can work through those with some automated support for knowing we aren't getting regressions.

filipsajdak:
OK, that sounds good. I am also struggling with ensuring what is working or not. I am supporting myself with a compiler explorer to ensure if new ideas are working on gcc, clang, and MSVC.

hsutter:
Looks good! Thanks.

PR #129, opened 2022-11-22
[FIX] Fix declarations of cpp2_new

gerlero:
Small fixes in the declarations of the `cpp2_new` functions:
* Take the argument packs by forwarding reference so that the arguments are actually forwarded
* Remove the `... Args` template parameter that is not used

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -338,20 +338,20 @@ auto assert_in_bounds(auto&& x, auto&& arg CPP2_SOURCE_LOCATION_PARAM_WITH_DEFAU
 //
 struct {
     template<typename T>
-    [[nodiscard]] auto cpp2_new(auto ...args) const -> std::unique_ptr<T> {
+    [[nodiscard]] auto cpp2_new(auto&& ...args) const -> std::unique_ptr<T> {
         return std::make_unique<T>(std::forward<decltype(args)>(args)...);
     }
 } unique;
 
 struct {
     template<typename T>
-    [[nodiscard]] auto cpp2_new(auto ...args) const -> std::shared_ptr<T> {
+    [[nodiscard]] auto cpp2_new(auto&& ...args) const -> std::shared_ptr<T> {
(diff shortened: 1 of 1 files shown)


hsutter:
Ah, good catch! That totally was a silly error on my part as I was modifying the code, and got into a not-fully-updated-but-happened-to-be-compilable state (sigh).

But I'm glad the core mistake here wouldn't have been possible in Cpp2: In Cpp2 it isn't possible to try to forward a parameter that isn't declared as forwardable, you have to correctly declare `forward` parameters and then the forwarding is automatic.

Aside: I'm looking forward to when I can start self-hosting parts of cppfront in the coming months, meaning not only writing some of the compiler's own code in Cpp2, but also parts of `cpp2util.h`. That is, I aim to write that code first as a `cpp2util.h2` mixed Cpp1/Cpp2 file with some parts like these three `cpp2_new` functions written in Cpp2, then generating `cpp2util.h` from that. That will get rid of some latent bugs like this and make some of the code simpler.

gerlero:
> But I'm glad the core mistake here wouldn't have been possible in Cpp2: In Cpp2 it isn't possible to try to forward a parameter that isn't declared as forwardable, you have to correctly declare `forward` parameters and then the forwarding is automatic.

That was my exact thought as I saw it. Hopefully we can start to forget about the whole "perfect forwarder" idiom soon enough.

> Aside: I'm looking forward to when I can start self-hosting parts of cppfront in the coming months, meaning not only writing some of the compiler's own code in Cpp2, but also parts of `cpp2util.h`. That is, I aim to write that code first as a `cpp2util.h2` mixed Cpp1/Cpp2 file with some parts like these three `cpp2_new` functions written in Cpp2, then generating `cpp2util.h` from that. That will get rid of some latent bugs like this and make some of the code simpler.

Looking forward to that (as well as eventual support for classes, etc.)

Thanks for making this available! I really like your ideas for C++'s future, and I hope this project succeeds.

PR #130, opened 2022-11-22
[REFACTOR] Define global new in terms of unique.new

gerlero:
More DRY and IMO makes the intent (i.e., plain `new` is a synonym of `unique.new`) clearer.


hsutter:
@gerlero Thanks Gabriel! I don't have a CLA on file for you so please send me email and I'll send you a DocuSign link.

hsutter:
Looks good. Thanks!

hsutter:
GitHub is having trouble seeing that this doesn't really conflict with #129 so I'll just apply this as a standalone commit.

Issue #133, opened 2022-11-22
[SUGGESTION] Use std::format for string formatting if possible


switch-blade-stuff:
Since we have `std::format` as part of STL now since C++20, I would expect people to be implementing specializations for `std::formatter` in Cpp1. As such, in my opinion it would be beneficial to use `std::format` for to-string conversion (or string formatting itself) if possible.

It would avoid the need to duplicate string conversion code (only a specialization of `std::formatter` will be needed), and could improve performance, since `std::format` is more efficient than string concat. **It would also prevent creation of yet another string formatting standard**.

switch-blade-stuff:
What to do with `char8_t`, `char16_t` & `char32_t`?

With `char8_t`

switch-blade-stuff:
Missclick, terribly sorry.

With `char8_t`, `char16_t`, `char32_t` the solution could be to either use the `to_string` and string concat approach, or use `std::format` and string conversion. The second approach would be more universal, however involves locales. Personally, it seems like an acceptable compromise, since the use of `char8_t`, `char16_t` and `char32_t` for string literals and other general-purpose string operations is limited at best due to usability and compatibility issues (`char8_t` is addressed with C++23), as such string conversion would not have a big impact.

hsutter:
I like this suggestion and I'm all for using the best standard facilities and requiring C++20 as a baseline... but this means the reality of "C++20 as she is spoke," meaning the (large) subset of C++20 that is implemented and supported in at least the three major compilers.

As a quick test to see how `<format>` was doing, I tried pasting cppreference's short `<format>` example and as of this writing only 2 of 3 major implementations can handle it... Godbolt with GCC/Clang/MSVC "trunk/latest": https://godbolt.org/z/sxhMc4sPW

So I'm interested in using `<format>` but I can't take a hard dependency on it yet, because I'll generate code that won't work on libc++. Does that make sense?

---

That said, this leads to a good question: What level of C++ support should cppfront take a hard dependency on? Informally I've been targeting the following, because I want Cpp2 code to be compilable with at least the latest version of the three major C++ compilers:

> all of C++20 supported by the current released versions of MSVC, GCC, and Clang

But perhaps I should raise the bar a little because not everyone can pick up the very latest version right away... maybe it should be

> all of C++20 supported by the current and previous major versions of MSVC, GCC, and Clang

[...]

PR #134, opened 2022-11-23
[SUGGESTION] Use std::format for string formatting and to_string when available

switch-blade-stuff:
Implementation of the suggestion in #133


hsutter:
Thanks! Unfortunately not all of the major implementations support `<format>` well yet.

For details and further  discussion see #133's comment thread.

PR #139, opened 2022-11-25
[FIX] Fix return value of main() if program failed

knst:
If *.cpp2 file contains any syntax error or in case if wrong arguments are specified, the `cppfront` will return success (main returns 0).

This patch propagates error code as a result of `cppfront` in case of wrong arguments or in case of syntax error in at least one input file.

It may allow to write scripts that runs build of resulting C++ output or any other steps such as [example for Bash]:
```cppfront main.cpp && echo OK```

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -2670,10 +2670,11 @@ auto main(int argc, char* argv[]) -> int
 
     if (cmdline.arguments().empty()) {
         std::cout << "cppfront: error: no input files\n";
-        return 0;
+        return EXIT_FAILURE;
     }
 
     //  For each .cpp2 source file
+    int exit_status = EXIT_SUCCESS;
     for (auto const& arg : cmdline.arguments())
     {
         std::cout << arg.text << "...";
@@ -2701,6 +2702,7 @@ auto main(int argc, char* argv[]) -> int
             std::cout << "\n";
             c.print_errors();
             std::cout << "\n";
+            exit_status = EXIT_FAILURE;
         }
 
         //  In any case, emit the debug information (during early development this is
@@ -2709,4 +2711,5 @@ auto main(int argc, char* argv[]) -> int
             c.debug_print();
         }
     }
+    return exit_status;
 }
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -2665,7 +2665,7 @@ auto main(int argc, char* argv[]) -> int
     cmdline.process_flags();
 
     if (cmdline.help_was_requested()) {
-        return 0;
+        return EXIT_SUCCESS;
     }
 
     if (cmdline.arguments().empty()) {

### CODE REVIEW

File: source/cppfront.cpp [hunk 1]
```diff
@@ -2709,4 +2711,5 @@ auto main(int argc, char* argv[]) -> int
             c.debug_print();
         }
     }
+    return ret_status;
```
MichaelCook, line 2714: I understand the C++ standards only allow for 0 (EXIT_SUCCESS) and 1 (EXIT_FAILURE).
In POSIX, only the least significant byte of ret_status would be used.
So, -1 and -2 are potentially problematic for those reasons.

https://en.cppreference.com/w/cpp/utility/program/exit
https://pubs.opengroup.org/onlinepubs/007904875/functions/exit.html

hsutter:
Thanks -- this seems useful, but `-1` and `-2` seem to be nonportable values... should it just be `EXIT_FAILURE`?

The C and C++ standards talk about `0`, `EXIT_FAILURE`, or `EXIT_SUCCESS`, which are implementation-defined but a quick Godbolt test shows they appear to often be `0` and `1` respectively on the major compilers on at least the OSes being used here: https://godbolt.org/z/EchfErqYs

You can also `return` from `main`, or call `exit`, with your own value but it may be translated...  See: https://en.cppreference.com/w/c/program/exit. Pasting related sample doc material from cppreference since I have them open:

> If `exit_code` is `0` or [`EXIT_SUCCESS`](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating successful termination is returned. If exit_code is [`EXIT_FAILURE`](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating unsuccessful termination is returned. In other cases implementation-defined status value is returned.

> Both `EXIT_SUCCESS` and the value zero indicate successful program execution status (see [std::exit](https://en.cppreference.com/w/cpp/utility/program/exit)), although it is not required that `EXIT_SUCCESS` equals zero.

See this Godbolt example for `exit(-2)` and `return -2;`... https://godbolt.org/z/f66zPKnjW

- **MSVC on Windows:** MSVC generates `mov  eax, 0xfffffffe`, the OS reports `Program returned: 4294967294`
- **Gcc and Clang:** Compilers both generate `mov eax, -2`, the OS reports `Program returned: 254`

So perhaps just return `EXIT_FAILURE` on failure... thanks!

source/cppfront.cpp
knst: thanks for notice, updated to EXIT_SUCCESS and EXIT_FAILURE

filipsajdak:
I have used `EXIT_SUCCESS` and `EXIT_FAILURE` as it is described on cppreference.

wingunder (changes requested):
To be complete, change line 2668 to `return EXIT_SUCCESS;`

hsutter:
Looks good, thanks!

PR #140, opened 2022-11-25
[FIX] fix test 'mixed-intro-example-three-loops.cpp2' for gcc 12

knst:
Gcc 12 requires to specify header `<memory>` for using `std::make_unique`


hsutter:
@knst Thanks Konstantin! I don't have a CLA on file for you so please send me email and I'll send you a DocuSign link.

Issue #147, opened 2022-11-28
[BUG] using template arguments in inspect cause compiler to fail


filipsajdak:
The current implementation of cppfront is using constexpr ifs with requires expressions and `CPP2_TYPEOF` macro to eliminate compilation of alternatives that did not match in inspect expression. It suffers from issues with using macros and commas.

The below code:
```cpp
template <int i, int j>
auto calc() {
    return i + j;
}

fun: (v : _) -> int = {
    return inspect v -> int {
        is int  = calc<1,2>();
        is _ = 0;
    };
}

main: () -> int = {
    return fun(42);
}
```
generates to (skipping boilerplate):
```cpp
[[nodiscard]] auto fun(auto const& v) -> int{
    return [&] () -> int { auto&& __expr = v;
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
        else return 0; }
    ()
; }

[[nodiscard]] auto main() -> int{
    return fun(42); 
}
```

Problematic part is `CPP2_TYPEOF(calc<1,2>())` that fail to compile (because of the coma) with error:
```
tests/inspect_return_template.cpp2:12:132: error: too many arguments provided to function-like macro invocation
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
                                                                                                                                   ^
include/cpp2util.h:209:9: note: macro 'CPP2_TYPEOF' defined here
#define CPP2_TYPEOF(x)  std::remove_cvref_t<decltype(x)>
        ^
tests/inspect_return_template.cpp2:12:113: error: use of undeclared identifier 'CPP2_TYPEOF'
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
                                                                                                                ^
error: constexpr if condition is not a constant expression
tests/inspect_return_template.cpp2:18:12: note: in instantiation of function template specialization 'fun<int>' requested here
    return fun(42); 
           ^
3 errors generated.
```

[...]

PR #148, opened 2022-11-28
[FIX] add additional parens for CPP2_TYPEOF macro used in inspect

filipsajdak:
The current implementation of cppfront is using constexpr ifs with requires expressions and the `CPP2_TYPEOF` macro to eliminate the compilation of alternatives that did not match in inspect expression. It suffers from issues with using macros and commas.

The below code:
```cpp
template <int i, int j>
auto calc() {
    return i + j;
}

fun: (v : _) -> int = {
    return inspect v -> int {
        is int  = calc<1,2>();
        is _ = 0;
    };
}

main: () -> int = {
    return fun(42);
}
```
generates to (skipping boilerplate):
```cpp
[[nodiscard]] auto fun(auto const& v) -> int{
    return [&] () -> int { auto&& __expr = v;
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
        else return 0; }
    ()
; }

    return fun(42); 
}
```

The problematic part is `CPP2_TYPEOF(calc<1,2>())` that fails to compile (because of the coma) with the error:
```
tests/inspect_return_template.cpp2:12:132: error: too many arguments provided to function-like macro invocation
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
                                                                                                                                   ^
include/cpp2util.h:209:9: note: macro 'CPP2_TYPEOF' defined here
#define CPP2_TYPEOF(x)  std::remove_cvref_t<decltype(x)>
        ^
tests/inspect_return_template.cpp2:12:113: error: use of undeclared identifier 'CPP2_TYPEOF'
        if (cpp2::is<int>(__expr)) { if constexpr( requires{calc<1,2>();} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF(calc<1,2>()),int> ) return calc<1,2>(); else return int{}; else return int{}; }
                                                                                                                ^
error: constexpr if condition is not a constant expression
tests/inspect_return_template.cpp2:18:12: note: in instantiation of function template specialization 'fun<int>' requested here
    return fun(42); 
           ^
3 errors generated.
```

This fix adds extra parens and produce:
```cpp
CPP2_TYPEOF((calc<1,2>()))
```

Closes: https://github.com/hsutter/cppfront/issues/147

### DIFF
--- a/regression-tests/test-results/pure2-inspect-expression-in-generic-function-multiple-types.cpp
+++ b/regression-tests/test-results/pure2-inspect-expression-in-generic-function-multiple-types.cpp
@@ -35,8 +35,8 @@ auto test_generic(auto const& x) -> void{
         << std::setw(30) << typeid(x).name() 
         << " value is " 
         << [&] () -> std::string { auto&& __expr = x;
-            if (cpp2::is<int>(__expr)) { if constexpr( requires{"integer " + std::to_string(cpp2::as<int>(x));} ) if constexpr( std::is_convertible_v<CPP2_TYPEOF("integer " + std::to_string(cpp2::as<int>(x))),std::string> ) return "integer " + std::to_string(cpp2::as<int>(x)); else return std::string{}; else return std::string{}; }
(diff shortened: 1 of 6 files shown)


hsutter:
Looks good! I'll merge this then add the new test case above to `/regression-tests`.

Issue #150, opened 2022-11-29
[SUGGESTION] Basic Continuous Integration


jarzec:
**The statement**
This is a kind of meta-suggestion. The goal is not to provide a language feature request for the experimental new syntax of C++ itself but to streamline the development of all of them.

I propose to add basic scripts that would automatize running cppfront through all the existing tests and would not complicate addition of new test cases. As this should only serve utility purposes not the main goal of the experiment, it should be kept to the minimum. To allow automated execution of existing and future tests following their original idea it is enough to add just basic scripts (one for Windows and one for Linux). The scripts would allow to run the tests for all the compilers targeted by the project. An example of such script for Linux can be seen in [this commit](https://github.com/jarzec/cppfront/commit/e684ed8f25e54dc9fa558751b96905f4e7da4e6b). This is a variation on the existing `regression-tests/run-tests.bat` that already exists in the project. The proposition takes it a step further and performs comparisons against the expected outputs.

Such scripts would allow for quick and easy verification of any proposed change in the project's source code. With the addition of checks against available reference code/output a trivial Github workflow can be defined to automatically execute tests for any PR. Such automated workflow can be added to the project without any extra configuration effort, just by adding a single file.  This should greatly facilitate evaluation of change/fix propositions.
An example can be seen in [the commit mentioned above](https://github.com/jarzec/cppfront/commit/e684ed8f25e54dc9fa558751b96905f4e7da4e6b).
The following PRs shows the effect of including the proposition in the project:
- [Here](https://github.com/jarzec/cppfront/pull/2) is an example of a PR that introduces breaking changes both in the code generated by cppfront and in its output. The goal is to show how test failures can be reported on the PR website as well as [how failure details can be presented](https://github.com/jarzec/cppfront/actions/runs/3569949342/jobs/6000440881).
- [Here](https://github.com/jarzec/cppfront/pull/3) is an example of a PR that introduces non-breaking changes.

To add new test with the proposed approach it is enough to add a `.cpp2` source file in the `regression-test` directory and the respective `.cpp` and `.output` files in `regression-test/test-results`. Updating exisiting tests and expected results is equally trivial.
 
**Alternatives**
Test automation is clearly not essential for the goals of the experiment. As such this suggestion can be simply disregarded. However, few programmers (if any) would question the added value of CI tools (even the most basic ones) in terms of developability of any software and cppfront happens to be software.
[...]

PR #151, opened 2022-11-29
Add testing script and github workflow

jarzec:
A proposed implementation of https://github.com/hsutter/cppfront/issues/150.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,21 @@
+name: CI
+
+on: [pull_request]
+
+jobs:
+  verify-pr-linux:
+    name: 'Verify that merging PR #${{ github.event.number }} causes no regression (Linux)'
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks! I am definitely interested in CI, but not just yet because I'm trying to keep dependencies a minimal as possible at this stage, plus honestly I don't understand this yet and at this early partial incubation stage I want to keep the contents of this repro to things I can maintain myself. Sorry! but in the future if this goes past the personal incubation stage it will be time to add things like CI for sure. Thanks for understanding!

stefanofiorentino:
@jarzec here is my [fork](https://github.com/stefanofiorentino/cppfront) with both cmake (only for regression-tests for now) and basic CI (only for ubuntu-latest now). As this is not the main target of @hsutter, I'll keep it on my fork and not opening any PR for this. Feel free to have a look and judge anyway :)

PR #169, opened 2022-12-20
Add support for ufcs chaining

filipsajdak:
The chaining starts from the function call `fun().gun().bun()` or the method call `o.fun().gun()`.

It integrates chaining to part of the code that rewrites Cpp2 postfix operators to Cpp1 prefix operators. That allows the use of recurrency to go through the entire expression and insert UFCS in parts that need it.

After this change the Cpp2 code (taken from `mixed-postfix-expression-custom-formatting.cpp2`)
```cpp
test: (a:_) -> std::string = {
    return call( a,
        a.b(a.c)*++, "hello", /* polite
                          greeting
                          goes here */ " there",
        a.d.e( a.f*.g()++, // because f is foobar
              a.h.i(),
              a.j(a.k,a.l) )
        );
}
```
will generate
```cpp
[[nodiscard]] auto test(auto const& a) -> std::string{
    return call( a,
        ++*cpp2::assert_not_null(CPP2_UFCS(b, a, a.c)), "hello", /* polite
                          greeting
                          goes here */ " there",
        CPP2_UFCS(e, a.d, ++CPP2_UFCS_0(g, *cpp2::assert_not_null(a.f)), // because f is foobar
              CPP2_UFCS_0(i, a.h),
              CPP2_UFCS(j, a, a.k, a.l)));

}
```

~~The algorithm skips methods with more than one template argument (even when it happens inside chaining - it is done in a way that compiles fine).~~
This change is handling methods with more than one template arguments - already pushed to the PR.

This Cpp2 code:
```cpp
fun0()*.fun1(1)++.fun2<1,2>()++;
```
will generate:
```cpp
++CPP2_UFCS_TEMPLATE_0(fun2, (<1,2>), ++CPP2_UFCS(fun1, *cpp2::assert_not_null(fun0()), 1));
```

This change is not using UFCS for function calls with at least one argument. So, the functionality presented below is missing. 
```
fun() -> fun()
fun(a) -> CPP2_UFCS_0(fun, a)
fun(a, b) -> CPP2_UFCS(fun, a, b)
```

This change replaces: #17, #18

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -536,21 +536,23 @@ public:
     } \
 }(PARAM1)
 
-#define CPP2_UFCS_TEMPLATE(FUNCNAME,PARAM1,...) \
+#define CPP2_UFCS_REMPARENS(...) __VA_ARGS__
+
+#define CPP2_UFCS_TEMPLATE(FUNCNAME,TEMPARGS,PARAM1,...) \
 [](auto&& obj, auto&& ...params) { \
-    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME(std::forward<decltype(params)>(params)...); }) { \
-        return std::forward<decltype(obj)>(obj).template FUNCNAME(std::forward<decltype(params)>(params)...); \
+    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(params)>(params)...); }) { \
+        return std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(params)>(params)...); \
     } else { \
-        return FUNCNAME(std::forward<decltype(obj)>(obj), std::forward<decltype(params)>(params)...); \
+        return FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(obj)>(obj), std::forward<decltype(params)>(params)...); \
     } \
 }(PARAM1, __VA_ARGS__)
 
-#define CPP2_UFCS_TEMPLATE_0(FUNCNAME,PARAM1) \
+#define CPP2_UFCS_TEMPLATE_0(FUNCNAME,TEMPARGS,PARAM1) \
 [](auto&& obj) { \
-    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME(); }) { \
-        return std::forward<decltype(obj)>(obj).template FUNCNAME(); \
+    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (); }) { \
+        return std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (); \
     } else { \
-        return FUNCNAME(std::forward<decltype(obj)>(obj)); \
+        return FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(obj)>(obj)); \
     } \
 }(PARAM1)
 //--------------------------------------------------------------------
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1789,10 +1789,7 @@ public:
             }
             // Going backwards if we found Dot and there is args variable
             // it means that it should be handled by UFCS
-            else if( i->op->type() == lexeme::Dot && args
-                        // don't use UFCS for methods with more than one template argument
-                        && i->id_expr && i->id_expr->template_args_count() < 2
-                    )
+            else if( i->op->type() == lexeme::Dot && args )
             {
                 auto funcname = print_to_string(*i->id_expr);
 
@@ -1805,8 +1802,14 @@ public:
                 //printer.print_cpp2("CPP2_UFCS(", n.position());
 
                 auto ufcs_string = std::string("CPP2_UFCS");
+
                 if (i->id_expr->template_args_count() > 0) {
                     ufcs_string += "_TEMPLATE";
+                    // we need to replace "fun<int,long,double>" to "fun, (<int,long,double>)" to be able to generate
+                    // from obj.fun<int, long, double>(1,2) this CPP2_UFCS_TEMPLATE(fun, (<int,long, double>), obj, 1, 2)
+                    auto split = funcname.find('<'); assert(split != std::string::npos);
+                    funcname.insert(split, ", (");
+                    funcname += ')';
                 }
                 //  If there are no additional arguments, use the _0 version
                 if (args.value().empty()) {
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1726,77 +1726,6 @@ public:
             captured_part += "_" + std::to_string(mynum);
         }
 
-        //  Check to see if it's just a function call with "." syntax,
-        //  and if so use this path to convert it to UFCS
-        if (// there's a single-token expression followed by . and (
-            n.expr->get_token() &&                      //  if the base expression is a single token
-            std::ssize(n.ops) >= 2 &&                   //  and we're of the form:
-            n.ops[0].op->type() == lexeme::Dot &&       //       token . id-expr ( expr-list )
-            n.ops[1].op->type() == lexeme::LeftParen &&
-            // alpha limitation: if it's a function call with more than one template argument (e.g., x.f<1,2>())
-            //                   the UFCS* macros can't handle that right now, so don't UFCS-size it
-            n.ops[0].id_expr->template_args_count() < 2 &&
-            // and either there's nothing after that, or there's just a $ after that
-            (
-                std::ssize(n.ops) == 2 ||
-                (std::ssize(n.ops) == 3 && n.ops[2].op->type() == lexeme::Dollar)
-            )
-            )
-        {
-            //  If we already replaced this with a capture (which contains the UFCS
-            //  work already done when the capture was computed), emit the capture
-            if (!captured_part.empty()) {
-                printer.print_cpp2(captured_part, n.position());
-                return;
-            }
-
-            //  Otherwise, do the UFCS work...
-
-            //  The . has its id_expr
-            assert (n.ops[0].id_expr);
-
-            //  The ( has its expr_list and op_close
-            assert (n.ops[1].expr_list && n.ops[1].op_close);
-
-            //--------------------------------------------------------------------
-            //  TODO: When MSVC supports __VA_OPT__ in standard mode without the
-            //        experimental /Zc:preprocessor switch, use this single line
-            //        instead of the dual lines below that special-case _0 args
-            //  AND:  Make the similarly noted change in cpp2util.h
-            //
-            //printer.print_cpp2("CPP2_UFCS(", n.position());
-
-            auto ufcs_string = std::string("CPP2_UFCS");
-            if (n.ops[0].id_expr->template_args_count() > 0) {
-                ufcs_string += "_TEMPLATE";
-            }
-            //  If there are no additional arguments, use the _0 version
-            if (n.ops[1].expr_list->expressions.empty()) {
-                ufcs_string += "_0";
-            }
-            printer.print_cpp2(ufcs_string+"(", n.position());
-            //--------------------------------------------------------------------
-
-            //  Make the "funcname" the first argument to CPP2_UFCS
-            emit(*n.ops[0].id_expr);
-            printer.print_cpp2(", ", n.position());
-
-            //  Then make the base expression the second argument
-            emit(*n.expr);
-
-            //  Then tack on any additional arguments
-            if (!n.ops[1].expr_list->expressions.empty()) {
-                printer.print_cpp2(", ", n.position());
-                push_need_expression_list_parens(false);
-                emit(*n.ops[1].expr_list);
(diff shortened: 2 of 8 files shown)


filipsajdak:
@hsutter I merged the previous two PRs and merged this into your other solution. I didn't update the tests yet. If the change is good from your point of view, I will update everything.

UFCS is not complete after this change. There is a lack of UFCS support for cases like:
```
fun(a) -> CPP2_UFCS_0(fun, a)
fun(a, b) -> CPP2_UFCS(fun, a, b)
```

one of the issues that blocked me at the moment is the code:
```cpp
std::to_string(x as int);
```
generates
```cpp
CPP2_UFCS_0(std::to_string, cpp2::as<int>(x))
```
that, for some reason, fail to compile.

filipsajdak:
@hsutter I have added support for handling methods and functions with more than one template arguments.

Now instead of
```cpp
CPP2_UFCS_TEMPLATE(FUNCNAME,PARAM1,...)
```
UFCS macro for templates looks:
```cpp
CPP2_UFCS_TEMPLATE(FUNCNAME,TEMPARGS,PARAM1,...)
```
That means that calling
```cpp
obj.fun<int,long,double>(1,2);
```
will be generated to:
```cpp
CPP2_UFCS_TEMPLATE(fun, (<int,long,double>), obj, 1, 2);
```

Working prototype: https://godbolt.org/z/nnoEqz814

hsutter:
Thanks! I'm going to start looking at this next.

> It integrates chaining to part of the code that rewrites Cpp2 postfix operators to Cpp2 prefix operators. 

That sounds really promising, it's exactly what I had in mind.

hsutter:
@filipsajdak Thanks again, I have some clarifying questions...

> UFCS is not complete after this change. There is a lack of UFCS support for cases like:
> `fun(a) -> CPP2_UFCS_0(fun, a)`
> `fun(a, b) -> CPP2_UFCS(fun, a, b)`

Those should not be transformed though, right? The current intent was for UFCS to be only for `x.f(...)` memberlike call notation.

Are you proposing to make the `f(x,...)` nonmember call notation also UFCS, so that it could invoke a member function? I'd rather stick with UFCS-ing member call notation only for now, and it will probably make this PR's implementation easier. How does that sound?

P.S.: In the diff view, I loved seeing the current UFCS extra branch at the top of the function show in "bulk deleted" all-red!. It really should be rolled into the general inside-out transformation loop that's the main workhorse in the function, which is what you're doing.

filipsajdak:
It is just a comment. I was trying to clarify what is working and what is not. 

I have seen both approaches proposed.

I have checked https://isocpp.org/files/papers/N4165.pdf and realized that what is currently implemented is what you want (for sure, it supports auto-completion).

I was trying to implement the other style as well. It is possible but the current change supports your proposed style and is fine for me.

If it is not needed e.g. for generic code let's leave it as is.

filipsajdak:
TLDR: it supports only `x.f(...)`

hsutter:
Ack: Great!

filipsajdak:
Rebased to main.

hsutter:
In-progress review ack: Looking good.

I've now completed the first two review steps:
- A first cursory code review: The basic approach and structure seems right.
- Ran regressions as a quick smoke test: No final execution output changed, so that's looking good. But there were some small Cpp1 whitespace output changes.

Next on my list:
- Investigate those whitespace changes: They might be fine, or might just need passing a different `.position()` somewhere.
- Deeper code review: Think through the new code line by line to make sure I grok each step.

source/cppfront.cpp
```diff
@@ -1928,6 +1933,18 @@ class cppfront
...
 
+        if (args) {
+            // if after printing core expression args is defined
+            // it means that the chaining started by function call
+            // we need to print its arguments
+            suffix.emplace_back(")", n.position());
+            for (auto&& e: args.value()) {
+                suffix.push_back(e);
+            }
+            suffix.emplace_back("(", n.position());
```
hsutter, line 1944: I think these two lines that print `"("` and `")"` may be what's causing the whitespace changes I noticed in some of the regression tests. Should we extend `args` to store the positions of the left and right parens, and then use those positions here? Just using `n.position()` is using the position of the whole expression and we're past that... the parens may even be on other lines, as they are in `mixed-function-expression-and-std-for-each.cpp2` and `mixed-postfix-expression-custom-formatting.cpp2`... those are two examples where I saw the `)` get moved up/left instead of keeping its original position.

hsutter (commented):
Looks pretty good, mostly minor comments including that I think I found the reason the `(` `)` were changing position in some of the regression test cases. I have a suggestion that should resolve that.

include/cpp2util.h
```diff
@@ -536,21 +536,23 @@ class out {
...
 
-#define CPP2_UFCS_TEMPLATE(FUNCNAME,PARAM1,...) \
+#define CPP2_UFCS_REMPARENS(...) __VA_ARGS__
+
+#define CPP2_UFCS_TEMPLATE(FUNCNAME,TEMPARGS,PARAM1,...) \
 [](auto&& obj, auto&& ...params) { \
-    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME(std::forward<decltype(params)>(params)...); }) { \
-        return std::forward<decltype(obj)>(obj).template FUNCNAME(std::forward<decltype(params)>(params)...); \
+    if constexpr (requires{ std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(params)>(params)...); }) { \
+        return std::forward<decltype(obj)>(obj).template FUNCNAME CPP2_UFCS_REMPARENS TEMPARGS (std::forward<decltype(params)>(params)...); \
```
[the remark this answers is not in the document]
filipsajdak: I have learned that here: https://www.scs.stanford.edu/~dm/blog/va-opt.html while working on https://github.com/hsutter/cppfront/pull/47

I am checking the link you have sent and will check the proposed name.

source/cppfront.cpp
```diff
@@ -1828,9 +1776,60 @@ class cppfront
...
+
+                auto ufcs_string = std::string("CPP2_UFCS");
+
+                if (i->id_expr->template_args_count() > 0) {
+                    ufcs_string += "_TEMPLATE";
+                    // we need to replace "fun<int,long,double>" to "fun, (<int,long,double>)" to be able to generate
+                    // from obj.fun<int, long, double>(1,2) this CPP2_UFCS_TEMPLATE(fun, (<int,long, double>), obj, 1, 2)
+                    auto split = funcname.find('<'); assert(split != std::string::npos);
+                    funcname.insert(split, ", (");
+                    funcname += ')';
```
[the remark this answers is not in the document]
filipsajdak: At this point we know that it is a template so it should have `<` and `>` - assertion is the right thing to do here.

I will add it.

source/cppfront.cpp
```diff
@@ -1874,21 +1873,25 @@ class cppfront
...
 
                 if (i->id_expr) {
-                    auto print = std::string{};
-                    printer.emit_to_string(&print);
-                    emit(*i->id_expr, false /*not a local name*/);
-                    printer.emit_to_string();
+
+                    if (args) {
+                        // if args are stored it means that this is function or method
+                        // that is not handled by UFCS e.g. that has more than one template argument
```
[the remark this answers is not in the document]
filipsajdak: That comment was for you to emphasise that it will work also for more template arguments.

source/cppfront.cpp
```diff
@@ -1805,6 +1734,25 @@ class cppfront
...
 
+        auto args = std::optional<std::vector<text_with_pos>>{};
+
+        auto print_to_string = [&](auto& i, auto... args) {
+            auto print = std::string{};
+            printer.emit_to_string(&print);
+            emit(i, args...);
+            printer.emit_to_string();
+            return print;
+        };
```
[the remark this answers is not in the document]
filipsajdak: I am glad that you like it.

source/cppfront.cpp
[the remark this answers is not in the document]
filipsajdak: Ah... Now I see that I did not update the comment - will correct that.

source/cppfront.cpp
filipsajdak: Yes, you are right. I struggled to find a solution as I know it is essential for you.

I tried to use `push_need_expression_list_parens(true)` at the generation of args, but something was wrong. I will test it once again.

include/cpp2util.h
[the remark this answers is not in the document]
filipsajdak: I don't like the `CPP2_VA_ARGS_IF_PARENS` name. `COMMA_IF_PARENS` is used when there are parentheses, then it will expand to the comma. In contrast, `CPP2_VA_ARGS_IF_PARENS` suggests that it will expand to the `__VA_ARGS__`, which is misleading. In the context of our macro, we use `__VA_ARGS__` but from the perspective of the caller it is just removing parentheses or there is also used the name `EXPAND` that might replace `REMPARENS`.

From my perspective, `REMPARENS` is more descriptive.

source/cppfront.cpp
hsutter: You shouldn't have to push the parens state... The basic idea would be to store the open and close paren positions alongside `args` itself, either as `source_position`s or as `token const*`s to the paren tokens so we can ask them their positions.

But I can also go do this post-merge if that's the only thing holding up the merge, don't let this alone block the PR.

hsutter: OK, I've found and tested the solution to correct the paren positions. I'll post it here as another review, it's just two more lines to remember the `(` and `)` position and then using them in the three places.

source/cppfront.cpp
```diff
@@ -1805,6 +1734,25 @@ class cppfront
         auto last_was_prefixed = false;
         auto saw_dollar        = false;
 
+        auto args = std::optional<std::vector<text_with_pos>>{};
```
hsutter, line 1737: Here, store two more values (we could bundle these three in a struct but I'm fine with three similarly named locals):

```
        auto args_open_pos  = source_position{};
        auto args_close_pos = source_position{};
```

source/cppfront.cpp
```diff
@@ -1828,9 +1776,60 @@ class cppfront
                 }
             }
 
+            // Going backwards if we found LeftParen it might be UFCS
+            // expr_list is emited to args variable for future use
+            if (i->op->type() == lexeme::LeftParen) {
+
```
hsutter, line 1782: Here, add:

```
                assert(i->op_close);
                args_open_pos  = i->op->position();
                args_close_pos = i->op_close->position();
```

source/cppfront.cpp
```diff
@@ -1828,9 +1776,60 @@ class cppfront
...
+                    funcname.insert(split, ", (");
+                    funcname += ')';
+                }
+                //  If there are no additional arguments, use the _0 version
+                if (args.value().empty()) {
+                    ufcs_string += "_0";
+                }
+
+                prefix.emplace_back(ufcs_string + "(" + funcname + ", ", i->op->position() );
+                suffix.emplace_back(")", i->op->position() );
```
hsutter, line 1820: Change these two lines to:

```
                prefix.emplace_back(ufcs_string + "(" + funcname + ", ", args_open_pos );
                suffix.emplace_back(")", args_close_pos );
```

source/cppfront.cpp
```diff
@@ -1828,9 +1776,60 @@ class cppfront
...
+                if (!args.value().empty()) {
+                    for (auto&& e: args.value()) {
+                        suffix.push_back(e);
+                    }
+                    suffix.emplace_back(", ", i->op->position());
```
hsutter, line 1825: (Aside: This position isn't precise but is probably fine because we're rewriting so much, and this position is probably earlier than we're currently at which means it won't move anything. There are other places, particularly in this function, where I don't use a precise position for similar reasons.)

source/cppfront.cpp
hsutter, line 1944: These two `suffix.` lines should also use the positions:

```
            suffix.emplace_back(")", args_close_pos);
            ...
            suffix.emplace_back("(", args_open_pos);
```

That should do it!

source/cppfront.cpp
```diff
@@ -1874,21 +1873,25 @@ class cppfront
...
+                        // that is not handled by UFCS e.g. that has more than one template argument
+                        suffix.emplace_back(")", n.position());
+                        for (auto&& e: args.value()) {
+                            suffix.push_back(e);
+                        }
+                        suffix.emplace_back("(", n.position());
+                        args.reset();
+                    }
+
+                    auto print = print_to_string(*i->id_expr, false /*not a local name*/);
```
hsutter, line 1888: These two `suffix.` lines should also use the positions:

```
                        suffix.emplace_back(")", args_close_pos);
                        ...
                        suffix.emplace_back("(", args_open_pos);
```

hsutter (commented):
This review shows the fix for preserving the `(` `)` positions.

source/cppfront.cpp
filipsajdak: That was also one of my ideas that I will try.

source/cppfront.cpp
filipsajdak: Ok. Thank you for all your comments. I will introduce these changes.

filipsajdak:
@hsutter I have sent an updated code.

hsutter:
Looks good, merging... thanks, this is a big improvement.

filipsajdak:
Thank you. If you see any other area I might help with, please let me know.

I will do my best to help.

PR #170, opened 2022-12-20
[ADD] Add test-results for apple-clang-14

filipsajdak:
As suggested (here https://github.com/hsutter/cppfront/issues/133#issuecomment-1356944284) I have created `apple-clang-14` in `test-results` that compiles regression tests with:
```
Target: arm64-apple-darwin22.1.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

The main issue is that it does not support `<ranges>` - that means that there are no correct results for:

* mixed-captures-in-expressions-and-postconditions.cpp2
* mixed-function-expression-and-std-ranges-for-each-with-capture.cpp2
* mixed-function-expression-and-std-ranges-for-each.cpp2
* mixed-function-expression-with-pointer-capture.cpp2
* mixed-function-expression-with-repeated-capture.cpp2

I have added failed `.output` files for them (with error):
```
mixed-function-expression-and-std-ranges-for-each-with-capture.cpp2:14:18: error: no member named 'for_each' in namespace 'std::ranges'
    std::ranges::for_each
    ~~~~~~~~~~~~~^
mixed-function-expression-and-std-ranges-for-each-with-capture.cpp2:18:18: error: no member named 'for_each' in namespace 'std::ranges'
    std::ranges::for_each( view, std::move(callback));
    ~~~~~~~~~~~~~^
2 errors generated.
```


hsutter:
Thanks!

No `<ranges>`, then? Interesting. Looking at the [cppreference C++20 conformance table](https://en.cppreference.com/w/cpp/compiler_support/20), I see Apple Clang is also missing `<=>` (spaceship), `jthread`, and of course `<format>` (the last one doesn't surprise me but the others do). 

Of these, the one that concerns me most is `<=>`, because I was planning to soon take a hard dependency on it in emitted Cpp1 code as I add support for user-defined types... and not having that is going to be interesting because I don't want to have to emit all that pre-C++20 comparison boilerplate just to the emitted Cpp1 source works on Apple Clang. :-( The three-way comparison `<=>` is the first Cpp2-derived feature to be standardized in ISO C++, and I was expecting to be able to rely on its being there to generate code. Well, I know actual Clang's support for `<=>` is "partial" but it seemed to work as expected the last time I happened to try it in passing... maybe Apple Clang will add it in the next release.

filipsajdak:
You already use `operator<=>` in cppfront in common.h file:

```cpp
struct source_position
{
    lineno_t    lineno;     // one-based offset into program source
    colno_t     colno;      // one-based offset into line

    source_position(lineno_t l = 1, colno_t  c = 1 )
        : lineno{ l }, colno{ c }
    {
    }

    auto operator<=>(source_position const&) const = default;

    auto to_string() const -> std::string
    {
        return "(" + std::to_string(lineno) + "," + std::to_string(colno) + ")";
    }
};
```

It compiles and works... does it means it is working?

hsutter:
Right, that's in cppfront's own source. So if Apple Clang doesn't support `<=>` then presumably it won't be able to compile cppfront itself.

But someone using an already-compiled cppfront executable to compile some Cpp2 code to Cpp1 and using Apple Clang as the second (Cpp1) compiler will be fine, because executing cppfront doesn't currently generate `<=>` in the Cpp1 output (including that there's no `<=>` in `cpp2util.h` that the Cpp1 output will include).

filipsajdak:
Taking a look at https://en.cppreference.com/w/cpp/compiler_support and I see that [`P1185R2`](https://wg21.link/P1185R2) is supported by apple clang 12 (I am currently using 14).

hsutter:
Strange that it would have P1185 supported but P0515 not. Oh well, maybe the table is not consistent or up to date -- if you try to compile `cppfront.cpp` with Apple Clang that should tell you pretty quickly whether it supports `<=>`. :)

filipsajdak:
This is my main compiler, and I have been compiling cppfront with Apple clang since my first contribution to cppfront.

That is why I was surprised to see that it is not supported.

filipsajdak:
That means apple clang has support for `operator <=>` and you don't have to change your plans.

hsutter:
Thanks. I'll ping Sergey and let him know he might want to update cppreference.

PR #181, opened 2022-12-28
[SUGGESTION] Replace compiler-dependent test code with modern alterntives

jarzec:
This PR proposes to replace some platform/compiler-dependent `<cstdlb>` and `<ctime>` features in test code with their modern, fully standardized counterparts.

The rationale here is that test code provides good introduction to a software project and a sort of reference. That was indeed the case with `cppfront` for myself. I believe that test code should be of "production-quality" as much as possible.

Over the years I developed a bit of allergy against platform/implementation-dependent features in C++ (most of those inherited from ANSI C). I was very pleased that subsequent versions of C++ standardized pseudo-random number generators, time measurement etc. The determinism provided by standardization allows for easier reasoning about the code. It allows e.g. for using deterministic branching that is difficult for compilers to optimize away.

The proposed changes replace `std::rand()` with `std::mt19937()()` and `std::time()` with the features of `std::chrono`. Apart from the determinism these changes eliminate the use of raw pointers that cause even stronger allergic reactions...

Note that the execution outputs of the translated + compiled code did not change. Thanks to the changes it was also possible to reestablish some branching while preserving determinism.

### DIFF
--- a/regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
+++ b/regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
@@ -26,5 +26,8 @@ fill: (
 print_decorated: (x:_) = std::cout << ">> [" << x << "]\n";
 
 // for test determinism, force "fill" branch
-// the standard mandates that std::mt19937()() == 3499211612
-flip_a_coin: () -> bool = std::mt19937()() % 2 == 1;
+bool flip_a_coin() {
+    // Change std::mt19937 to std::random_device for non-deterministic PRNG
+    static std::mt19937 rand; 
+    return rand() % 2 == 1;
+}
--- a/regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
+++ b/regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
@@ -1,6 +1,6 @@
+#include <random>
 #include <string>
 #include <vector>
-#include <ctime>
 
 main: () -> int = {
     x: std::string;         // note: uninitialized!
@@ -26,4 +26,5 @@ fill: (
 print_decorated: (x:_) = std::cout << ">> [" << x << "]\n";
 
 // for test determinism, force "fill" branch
-flip_a_coin: () -> bool = false;
+// the standard mandates that std::mt19937()() == 3499211612
+flip_a_coin: () -> bool = std::mt19937()() % 2 == 1;
--- a/regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
+++ b/regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
@@ -1,13 +1,14 @@
 
-#include <cstdlib>
 #include <iostream>
+#include <random>
 
 main: () -> int = {
     x: int = 42;
     p: *int;
 
     // ... more code ...
-    if std::rand()%2 {
+    // the standard mandates that std::mt19937()() == 3499211612
+    if std::mt19937()()%2 {
         // p = nullptr;
     }
     else {
--- a/regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
+++ b/regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
@@ -1,14 +1,13 @@
 
+#include <cstdlib>
 #include <iostream>
-#include <random>
 
 main: () -> int = {
     x: int = 42;
     p: *int;
 
     // ... more code ...
-    // the standard mandates that std::mt19937()() == 3499211612
-    if std::mt19937()()%2 {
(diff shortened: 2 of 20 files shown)


regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
```diff
@@ -26,4 +26,5 @@ fill: (
 print_decorated: (x:_) = std::cout << ">> [" << x << "]\n";
 
 // for test determinism, force "fill" branch
-flip_a_coin: () -> bool = false;
+// the standard mandates that std::mt19937()() == 3499211612
```
jarzec, line 29: [Explanation] This is true because the standard defines the default seed.

regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
```diff
@@ -1,13 +1,14 @@
...
+#include <random>
 
 main: () -> int = {
     x: int = 42;
     p: *int;
 
     // ... more code ...
-    if std::rand()%2 {
+    // the standard mandates that std::mt19937()() == 3499211612
+    if std::mt19937()()%2 {
```
jarzec, line 11: [Question] What do you think about disallowing implicit conversions of integer types to `bool`?

regression-tests/mixed-initialization-safety-3-contract-violation.cpp2
hsutter: Calling `std::rand()` multiple times will return different results, but calling `std::mt19937()()` will always return the same result, so this changes meaning (if the flip function were called multiple times)?

regression-tests/mixed-lifetime-safety-pointer-init-1-error.cpp2
hsutter: I plan to add that sometime down the road, yes.

hsutter:
Thanks. Hmm, I'm not sure about this change... some thoughts:

Re compiler-dependent: The headers/facilities you're removing are portable standard facilities, even if some do date back to C. The purpose of this repo isn't to show the most modern code using today's C++ syntax (Cpp1), and I actually deliberately use C headers sometimes to show the 100% compatibility cppfront gives. I'm definitely interested in clean modern code, but the clean modern code I'm most interested in exploring is the experimental Cpp2 syntax...

Re `std::rand`: One reason I generally avoid `<random>` in tests and demos is because it's actually quite hard to use... and I say this after directly asking the author of `<random>` the easiest way to use it that is usability-competitive with `std::rand()` which we should generally discourage :) but it's just not usability-competitive so I don't use it unless I really need high-quality random numbers which I don't in demos and tests.

Re `std::mt19937()()`: Doesn't that give the same answer every time, unlike `std::rand()`? So changing `std::rand()` to `std::mt19937()()` changes the meaning (if `flip_a_coin` were actually called multiple times), right?

jarzec:
I understand your concerns. Let me try to address them.

When I spoke of compiler-dependence I didn't mean issues with portability but determinism. I did understand you used the `<cxyz>` headers on purpose and I did not remove all of them. I did go a bit too far with some tests where the determinism of execution is not impacted by the call to `std::rand` - I will revert those.

As to the second point I noticed that in fact some of the tests recently became less attractive as examples as branches in them had to be made identical to ensure consistent output. My initial idea for this PR was to try to restore the non-identical branches by using deterministic PRNGs to control which branch will actually be taken.

I agree that the use of `std::mt19937()()` is not great. I started using that in cases that needed a single-shot randomness and it ended up in `flip_a_coin`s as well. To defend myself I can only say that the current implementations of `flip_a_coin` constantly return `true` or `false`, depending on the test.

What would you say about using the following to allow non-identical branching to be restored:
```
flip_a_coin: () -> bool = {
    // change std::mt19937 to std::random_device for non-deterministic PRNG
    static std::mt19937 rand; 
    return rand() % 2 == 1;
}
```
I see a number of advantages of this "poor man's coroutine":
- The intent is well expressed by the name and return type, as it used to be.
- It is not bad in terms of usability - just a single extra line of actual code compared to the original.
- The generated sequence of bool values is of decent quality for the purposes.
- The sequence is deterministic so the requested effective branching in the test can be easily ensured.
- The coin flip can be trivially made non-deterministic if need be, e.g. for a demo - as described in the comment.

As stated before: in cases where determinism of the PRNG is not important I will revert to `std::rand()`.

jarzec:
I reverted all the changes where there were no issues with determinism. I also modified the `flip_a_coin()` implementations to make them deterministically random.
@hsutter If you still prefer the original version you can close the PR. I could also improve it further.

hsutter:
Looks, good, thanks!

PR #183, opened 2022-12-29
Add option to emit source code to stdout

s2x:
This simple change allows you to build files directly without any intermediate steps. Previously, to compile a file, you had to first create a `*.cpp` file and then use a compiler. If we add the ability to print `*.cpp` file content to `stdout`, then we can use `g++/clang++` options to make `stdin` the source.

example commands:

for `g++`
`./cppfront regression-tests/pure2-hello.cpp2 -e | g++-10 -std=c++2a -x c++ -I include/ -O3 -o pure2-hello -`

same for `clang`
`./cppfront regression-tests/pure2-hello.cpp2 -e | clang++ -std=c++2a -x c++ -I include/ -O3 -o pure2-hello -`


source/cppfront.cpp
```diff
@@ -346,6 +353,8 @@ class positional_printer
             //  remain diff-identical
             print_extra("\n");
         }
+
+        cpp1_filename.empty() ? std::cout<<out.str() : file_out<<out.str();
```
prenaux, line 357: I would use a regular if/else statement here. None of the current code use the ternary operator.

source/cppfront.cpp
```diff
@@ -2853,14 +2867,15 @@ auto main(int argc, char* argv[]) -> int
     }
 
     if (cmdline.arguments().empty()) {
-        std::cout << "cppfront: error: no input files\n";
+        std::cerr << "cppfront: error: no input files\n";
```
prenaux, line 2870: Outputting all errors in stderr is better regardless of whether this feature merges.

source/cppfront.cpp
```diff
@@ -361,17 +370,17 @@ class positional_printer
...
     }
 
     auto is_open() -> bool {
-        if (out.is_open()) {
+        if (file_out.is_open() || cpp1_filename.empty()) {
             assert (pcomments && "ICE: if out.is_open, pcomments should also be set");
         }
-        return out.is_open();
+        return file_out.is_open() || cpp1_filename.empty();
     }
```
prenaux, line 384: I'd suggest moving the `cpp1_filename.empty()` tests to a `is_stdout_output()` utility function for clarity.

prenaux (commented):
This is useful to remove the need for a temporary file in build systems on at least Linux and macOS.

I wrote some comments in the spirit of improving code clarity.

Regardless of whether this merge outputting all errors to stderr would be better.

hsutter:
Thanks! Why not use the current `-o` flag and recognize the special string `stdout` as not being a file?

filipsajdak:
What would be beneficial is to have the possibility to run cppfront that writes to `stringstream` that will allow to write in-memory tests.

source/cppfront.cpp
hsutter: Good catch, most of them were going to cerr but I missed these ones, thanks!

hsutter:
Thanks! I've decided to take this but tweak the implementation a little, so I've applied it as a separate commit instead of asking you to make a bunch of small tweaks. Thanks again.

PR #187, opened 2022-12-30
cpp2 compiles and run without exceptions and rtti

prenaux:
Some of my code disables exceptions & rtti. This PR makes minor changes in cpp2util.h to allow cpp2 compiled code to compile & run with both exceptions and RTTI disabled.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -20,6 +20,26 @@
 #ifndef __CPP2_UTIL
 #define __CPP2_UTIL
 
+#if __EXCEPTIONS
+    #define CPP2_EXCEPTIONS
+#endif
+
+#ifdef __RTTI
+    #define CPP2_RTTI
+#endif
+
+#if defined CPP2_EXCEPTIONS
+    #if defined CPP2_THROW
+        #error "Do not redefine CPP2_THROW when using cpp2 with exceptions enabled."
+    #endif
+    #define CPP2_THROW(DESC,EXCEPTION) throw EXCEPTION
+#else
+    #if !defined CPP2_THROW
+        // TODO: This should call a utility function that prints DESC and a backtrace.
+        #define CPP2_THROW(DESC,EXCEPTION) std::terminate()
+    #endif
+#endif
+
 //  If this implementation doesn't support source_location yet, disable it
 //  TODO: technically this test should have <version> included first, but GEFN
 #if !defined(_MSC_VER) && !defined(__cpp_lib_source_location)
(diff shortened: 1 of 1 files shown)


include/cpp2util.h
```diff
@@ -20,6 +20,26 @@
...
+
+#ifdef __RTTI
+    #define CPP2_RTTI
+#endif
+
+#if defined CPP2_EXCEPTIONS
+    #if defined CPP2_THROW
+        #error "Do not redefine CPP2_THROW when using cpp2 with exceptions enabled."
+    #endif
+    #define CPP2_THROW(DESC,EXCEPTION) throw EXCEPTION
```
prenaux, line 35: DESC & EXCEPTION are separated because #EXCEPTION only work for literals.

hsutter:
Thanks for your pull request! It looks like this may be your first contribution to cppfront. Before I can look at your pull request, you'll need to sign a Contributor License Agreement (CLA). I think I sent it a few days ago, but it appears to have not been deliverable so please send me email to coordinate... thanks.

prenaux:
@hsutter I just realized I used a new domain name for these commits, which I hadn't linked to an actual email server. I've done that now, so your emails should go through if you resend the CLA. Alternatively, I sent you an email from my main email.

prenaux:
Rebased on main.

hsutter:
Thanks, I appreciate this input and suggestion.

How will the `dynamic_cast` based `as` implementation work without RTTI?

I'm generally sympathetic to the costs of EH and that a lot of projects disable it, and in Cpp2 I do aim to eventually implement value-based exceptions (P0709). And I'm supportive of efforts to make RTTI costs better, especially on Windows.

But I'm not sure whether I can yet take the dependency on running in environments that don't support EH and RTTI. For example, IIUC the `.DS_Store` works only on macOS... what will I do on Linux? Windows? I don't want to create and support a parallel mechanism for reporting errors like `std::filesystem` chose to do, I think that bifurcation is a symptom of the problem (as I explain in P0709) rather than a solution.

I do appreciate this and I do understand the prevalence of environments that disable EH and RTTI. I'm just not sure I'm yet in a state where I can support compatibility with those environments...? Perhaps a better solution than `.DS_Store` would be to just brute-force disable the parts of `cpp2util.h` that depend on EH and RTTI in environments that can't handle EH and RTTI... but then that would appear to mean disabling large parts of `is` and `as` which are pretty fundamental?

I'll think about this some more. Thanks for the input.

prenaux:
`.DS_Store` is noise from the macOS Finder app which it adds as soon as you look at a folder, similar to `Thumbs.db` & `desktop.ini` with Windows Explorer. Its not related to this PR so I removed it.

> How will the `dynamic_cast` based `as` implementation work without RTTI?
It wont, the change I made is just to make the code compile. That code doesn't use `as` so it does compile and run. If a no-RTTI mode is supported it should disable features that depends on it. Cpp1 doesn't allow you to use dynamic_cast, typeid or throw with rtti/exceptions disabled either.

RTTI and exceptions are disabled in our case on wasm targets for the web (compiled with emscripten) because it adds about ~10-20% to the wasm output and it runs a bit slower even though the code doesn't use it. A chunk of our code and nice to have features of our libraries are also disabled on wasm to keep the final wasm output as small as possible since download size is a very significant factor when we release the application.

At the end of the day if the language is so much better with those features that cost might be acceptable, but in the case of Cpp1 its just not.

This could change with Cpp2, but atm trying out bits of cpp2 in various places its simpler if it supports a no-rtti/no-exceptions mode in our case.

hsutter:
OK, after thinking on this more I've merged this via another commit using a little bit of a different strategy:
- add a "Support for constrained target environment" compiler flags section with `-fno-exceptions` and `-fno-rtti`, which if set get passed through to `cpp2util.h` via a `#define` (similar to the way `-add-source-info` causes `CPP2_USE_SOURCE_LOCATION` to be set, and `-p` causes `CPP2_USE_MODULES` to be set)
- wrap `Throw`, `Uncaught_exceptions`, `Dynamic_cast`, and `Typeid`, which test those settings in their bodies, and try to give nice error messages if users try to exercise the (few) features each switch disables (much like your PR does, plus `final_action` will still run if an exception actually does happen)
- minor other housekeeping, like also don't try to include `<exception>` and/or `<type_info>` if disabled

@prenaux Would you please give this a try in your WASM environment, and let me know if it addresses your needs?

PR #188, opened 2022-12-30
[Suggestion] Add a compile time flag to control the line error output format.

prenaux:
Controls whether we print `FILE:line:col: error:` or `FILE(line,col): error:`.

### DIFF
--- a/source/common.h
+++ b/source/common.h
@@ -104,15 +104,6 @@ struct comment
 //
 struct error
 {
-    // Controls whether we print `FILE:line:col: error:` or `FILE(line,col):
-    // error:`.  Need to integrate with editors that jump to errors, etc.
-    // TODO: This should become a compiler flag.
-#ifdef _MSC_VER
-    static constexpr bool print_colon_errors = false;
-#else
-    static constexpr bool print_colon_errors = true;
-#endif
-
     source_position where;
     std::string     msg;
     bool            internal = false;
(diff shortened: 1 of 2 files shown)


source/common.h
```diff
@@ -104,6 +104,15 @@ struct comment
...
 struct error
 {
+    // Controls whether we print `FILE:line:col: error:` or `FILE(line,col):
+    // error:`.  Need to integrate with editors that jump to errors, etc.
+    // TODO: This should become a compiler flag.
+#ifdef _MSC_VER
+    static constexpr bool print_colon_errors = false;
+#else
+    static constexpr bool print_colon_errors = true;
+#endif
```
hsutter, line 114: I don't think we need two defaults... the default should be the same regardless of the compiler we used to build cppfront itself, right?

hsutter: As you say in the comment, we would want to add a switch to cppfront itself, something like this (consistent style with the other flags):

```
static auto flag_print_colon_errors = false;
static cmdline_processor::register_flag cmd_enable_source_info(
    2,
    "format-colon-errors",
    "Emit ':line:col:' format for error messages",
    []{ flag_print_colon_errors = true; }
);
```

Thanks!

hsutter (changes requested):

source/common.h
prenaux: I updated the code.

hsutter:
Thanks!

PR #94, opened 2022-10-25
[SUGGESTION][FIX] Add pointer deduce type support

filipsajdak:
Current implementation is not handling deduced pointer types when analysing safety checks.

It means that this code will pass safety checks
```cpp
i : int = 42;
pi := i&;
pi = 0; // this should trigger error
```

This change introduce support for deduced pointer types and makes safety checks work with below code:
```cpp
main: (argc : int, argv : **char) -> int = {
    a:     int = 2;
    pa:   *int = a&;
    ppa: **int = pa&;

    pa = 0; // error

    pa2:= ppa*;
    pa2 = 0; // error

    pa3 := a&;
    pa3 = 0;  // error
    pa3 += 2; // error

    ppa2 := pa2&;
    pa4 := ppa2*;

    pa4 = 0; // error

    pppa := ppa&;
    pa5 := pppa**;
    pa5 = 0; // error

    return a*pa**ppa**; // 8
}
```
and will cause errors
```
tests/pointer-to-pointer.cpp2...
pointer-to-pointer.cpp2(6,8): error: = - pointer assignment from null or integer is illegal
pointer-to-pointer.cpp2(9,9): error: = - pointer assignment from null or integer is illegal
pointer-to-pointer.cpp2(12,9): error: = - pointer assignment from null or integer is illegal
pointer-to-pointer.cpp2(13,9): error: += - pointer assignment from null or integer is illegal
pointer-to-pointer.cpp2(18,9): error: = - pointer assignment from null or integer is illegal
pointer-to-pointer.cpp2(22,9): error: = - pointer assignment from null or integer is illegal
  ==> program violates lifetime safety guarantee - see previous errors
```

Based on https://github.com/hsutter/cppfront/pull/93

PR #196, opened 2023-01-03
Add pointers deduced types

filipsajdak:
The current implementation has safety guarantees only for explicit pointers declaration. If the pointer type is deduced, cppfront does not provide a safety check.

This change introduces pointer type deduction for types defined in cpp2. After this change cppfront will process below code
```cpp
fun: (inout i:int) -> *int = {
    return i&;
}

fun2: (inout i:int) -> (result : *int) = {
    result = i&;
}

main: (argc : int, argv : **char) -> int = {
    a:     int = 2;
    pa:   *int = a&;
    ppa: **int = pa&;

    pa = 0;       // caught

    pa2:= ppa*;
    pa2 = 0;      // caught

    pa3 := a&;
    pa3 = 0;      // caught
    pa3 += 2;     // caught

    ppa2 := pa2&;
    pa4 := ppa2*;
    pa4 = 0;      // caught

    pppa := ppa&;
    pa5 := pppa**;
    pa5 = 0;      // caught

    fun(a)++;     // caught
    fp := fun(a);
    fp = 0;       // caught

    f := fun(a)*;

    fp2 := fun2(a).result;
    fp2--;        // not caught :(

    return a * pa* * ppa**; // 8
}
```

And will generate errors:
```
pure2-deducing-pointers.cpp2...
pure2-deducing-pointers.cpp2(14,8): error: = - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(17,9): error: = - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(20,9): error: = - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(21,9): error: += - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(25,9): error: = - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(29,9): error: = - pointer assignment from null or integer is illegal
pure2-deducing-pointers.cpp2(31,11): error: ++ - pointer arithmetic is illegal - use std::span or gsl::span instead
pure2-deducing-pointers.cpp2(33,8): error: = - pointer assignment from null or integer is illegal
  ==> program violates lifetime safety guarantee - see previous errors
  ==> program violates bounds safety guarantee - see previous errors
```

Unfortunately, this implementation does not deduce pointer types from functions declared after the analyzed code. That means that defining `fun` and `fun2` functions after the `main` will block deducing pointer type for lines
```cpp
fun(a)++;     // not caught anymore
fp := fun(a);
fp = 0;       // not caught anymore

f := fun(a)*; // not deduced as pointer type anymore
```

This implementation also does not deduce pointer types returned as `parameter_declaration_list_node`
```cpp
fp2 := fun2(a).result; // not deduced as pointer
fp2--;        // not caught :(
```

This replaces: https://github.com/hsutter/cppfront/pull/93 & https://github.com/hsutter/cppfront/pull/94

### DIFF
--- a/regression-tests/test-results/pure2-deducing-pointers.cpp2.output
+++ b/regression-tests/test-results/pure2-deducing-pointers.cpp2.output
@@ -1,0 +1,12 @@
+pure2-deducing-pointers.cpp2...
+pure2-deducing-pointers.cpp2(14,8): error: = - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(17,9): error: = - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(20,9): error: = - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(21,9): error: += - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(25,9): error: = - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(29,9): error: = - pointer assignment from null or integer is illegal
+pure2-deducing-pointers.cpp2(31,11): error: ++ - pointer arithmetic is illegal - use std::span or gsl::span instead
+pure2-deducing-pointers.cpp2(33,8): error: = - pointer assignment from null or integer is illegal
+  ==> program violates lifetime safety guarantee - see previous errors
+  ==> program violates bounds safety guarantee - see previous errors
+
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1211,7 +1211,7 @@ public:
 
         in_definite_init = is_definite_initialization(n.identifier);
         if (!in_definite_init && !in_parameter_list) {
-            if (auto decl = sema.get_local_declaration_of(*n.identifier);
+            if (auto decl = sema.get_declaration_of(*n.identifier);
                 is_local_name &&
                 decl &&
                 //  note pointer equality: if we're not in the actual declaration of n.identifier
@@ -1735,6 +1735,86 @@ public:
         }
     }
 
+    // Don't work yet, TODO: finalize deducing pointer types from parameter lists
+    auto is_pointer_declaration(parameter_declaration_list_node const* decl_node, int deref_cnt, int addr_cnt) -> bool {
+        return false;
+    }
+
+    auto is_pointer_declaration(declaration_node const* decl_node, int deref_cnt, int addr_cnt) -> bool {
+        if (!decl_node) {
+            return false;
+        }
+        if (addr_cnt > deref_cnt) {
+            return true;
+        }
+
+        return std::visit([&](auto const& type){
+            return is_pointer_declaration(type.get(), deref_cnt, addr_cnt);
+        }, decl_node->type);
+    }
+
+    auto is_pointer_declaration(function_type_node const* fun_node, int deref_cnt, int addr_cnt) -> bool {
+        if (!fun_node) { 
+            return false;
+        }
+        if (addr_cnt > deref_cnt) {
+            return true;
+        }
+
+        return std::visit([&]<typename T>(T const& type){
+            if constexpr (std::is_same_v<T, std::monostate>) {
+                return false;
+            } else {
+                return is_pointer_declaration(type.get(),  deref_cnt, addr_cnt);
+            }
+        }, fun_node->returns);
+    }
+
+    auto is_pointer_declaration(type_id_node const* type_node, int deref_cnt, int addr_cnt) -> bool {
+        if (!type_node) { 
+            return false;
+        }
+        if (addr_cnt > deref_cnt) {
+            return true; 
+        }
+
+        if ( type_node->dereference_of ) {
+            return is_pointer_declaration(type_node->dereference_of, deref_cnt + type_node->dereference_cnt, addr_cnt);
+        } else if ( type_node->address_of ) {
+            return is_pointer_declaration(type_node->address_of, deref_cnt, addr_cnt + 1);
+        }
+
+        int pointer_declarators_cnt = std::count_if(std::cbegin(type_node->pc_qualifiers), std::cend(type_node->pc_qualifiers), [](auto* q) {
+            return q->type() == lexeme::Multiply;
+        });
+
+        if (pointer_declarators_cnt == 0 && type_node->suspicious_initialization) {
+            return is_pointer_declaration(type_node->suspicious_initialization, deref_cnt, addr_cnt);
+        }
+
+        return (pointer_declarators_cnt + addr_cnt - deref_cnt) > 0;
+    }
+
+    auto is_pointer_declaration(declaration_sym const* decl, int deref_cnt, int addr_cnt) -> bool {
+        if (!decl) {
+            return false;
+        }
+        if (addr_cnt > deref_cnt) {
+            return true;
+        }
+        return is_pointer_declaration(decl->declaration, deref_cnt, addr_cnt);
+    }
+
+    auto is_pointer_declaration(token const* t, int deref_cnt = 0, int addr_cnt = 0) -> bool {
+        if (!t) {
+            return false;
+        }
+        if (addr_cnt > deref_cnt) {
+            return true;
+        }
+        auto decl = sema.get_declaration_of(*t, true);
+        return is_pointer_declaration(decl, deref_cnt, addr_cnt);
+    }
 
     //-----------------------------------------------------------------------
     //
@@ -1754,7 +1834,7 @@ public:
             assert (n.expr->get_token());
             assert (!current_args.back().ptoken);
             current_args.back().ptoken = n.expr->get_token();
-            auto decl = sema.get_local_declaration_of(*current_args.back().ptoken);
+            auto decl = sema.get_declaration_of(*current_args.back().ptoken);
             if (!(decl && decl->parameter && decl->parameter->pass == passing_style::forward)) {
                 errors.emplace_back(
                     n.position(),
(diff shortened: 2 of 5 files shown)


source/cppfront.cpp
```diff
@@ -1664,6 +1664,66 @@ class cppfront
         }
     }
 
+    // Don't work yet, TODO: finalize deducing pointer types from parameter lists
+    auto is_pointer_declaration(parameter_declaration_list_node const* decl_node, int deref_cnt, int addr_cnt) -> bool {
+        return false;
+    }
+
+    auto is_pointer_declaration(declaration_node const* decl_node, int deref_cnt, int addr_cnt) -> bool {
+        if (!decl_node) return false;
```
hsutter, line 1673: Style: I always enclose my if/else bodies in `{` `}`. So please do the same here and in various other places below. (The reason I do that is mainly to keep myself honest because I require this in Cpp2 (it's the price of not requiring `(` `)` around the conditional-expression) and so I'm forcing myself to use `{` `}` always in Cpp1 too to validate that it's not an onerous requirement, and indeed it is working very well for me.)

source/cppfront.cpp
```diff
@@ -1664,6 +1664,66 @@ class cppfront
...
+        if (addr_cnt > deref_cnt) return true;
+
+        return std::visit([&](auto const& type){
+            return is_pointer_declaration(type.get(), deref_cnt, addr_cnt);
+        }, decl_node->type);
+    }
```
hsutter, line 1679: Can you elaborate on what the `std::visit` is doing?

source/cppfront.cpp
```diff
@@ -1702,32 +1762,38 @@ class cppfront
...
+                                op->position(),
+                                op->to_string(true) + " - pointer arithmetic is illegal - use std::span or gsl::span instead"
+                            );
+                            violates_bounds_safety = true;
+                        }
+                        else if (op->type() == lexeme::Tilde) {
+                            errors.emplace_back(
+                                op->position(),
+                                op->to_string(true) + " - pointer bitwise manipulation is illegal - use std::bit_cast to convert to raw bytes first"
+                            );
```
hsutter, line 1796: I haven't had the cycles to really grok this, and the only reason I'm reading through the PR is because (a) the test case at top is clear and (b) the code change is mostly localized here. That limits the downside because it should be very local.

I'm not yet sure whether I should merge this... In the long term I aim to implement my C++ Core Guidelines Lifetime analysis, and that will likely replace at least parts of this work. But if we do merge this now, I still view it as an interim '[expanded] initial partial implementation' and don't think we would want to go further in this direction in the near term. First we would want to implement the Lifetime analysis and then see where we are, and the Lifetime implementation itself needs to be farther in the future after classes and metaclass functions.

source/parse.h
```diff
@@ -3318,6 +3322,24 @@ class parser
                 }
             }
 
+            if (deduced_type) {
+                auto& type = std::get<declaration_node::object>(n->type);
+                if (peek(1)->type() == lexeme::Ampersand) {
+                    type->address_of = &curr();
+                }
```
hsutter, line 3329: So we peeking at the first character after the `n->equal_sign`, at the second token of the initializer? Looking for initializers of the form `var&` only if they're unqualfied -- correct?

source/parse.h
```diff
@@ -3318,6 +3322,24 @@ class parser
...
+                else if (peek(1)->type() == lexeme::Multiply) {
+                    type->dereference_of = &curr();
+                    for (int i = 1; peek(i)->type() == lexeme::Multiply; ++i)
+                        type->dereference_cnt += 1;
+                }
+                else if (
+                    (peek(1)->type() == lexeme::LeftParen && curr().type() != lexeme::Colon)
```
hsutter, line 3336: This is trying to avoid unnamed function expressions?

This makes me wonder whether this is happening too early... should we be doing this later, in `sema.h`, not during parsing? In sema we know whether it's an unnamed function declaration, we don't have to guess with a heuristic that can break.

source/parse.h
```diff
@@ -3318,6 +3322,24 @@ class parser
...
+                    || curr().type() == lexeme::Identifier 
+                ) {
+                    type->suspicious_initialization = &curr();
```
hsutter, line 3339: It's not clear to me what is suspicious exactly... this probably needs a comment to define what is suspicious. Or maybe a better name?

source/parse.h
```diff
@@ -520,6 +520,10 @@ struct type_id_node
     source_position pos;
 
     std::vector<token const*> pc_qualifiers;
+    token const* address_of                 = {};
+    token const* dereference_of             = {};
+    int dereference_cnt                     = {};
+    token const* suspicious_initialization  = {};
```
hsutter, line 526: This additional data also is localized. Have you checked to make sure there are no lifetime issues (e.g., we always point to parents that will be destroyed after we are, or else don't use the pointers after the targets might be destroyed)?

hsutter (changes requested):
Overall, I think if we merge this the work maybe should happen later in `sema.h` (see comment in the review). But I'd much rather leave this until the [Lifetime static analysis](https://github.com/hsutter/cppfront#2015-lifetime-safety) is implemented, which will do a lot of this Pointer type deduction and state tracking... and that won't be in the near term (sorry!).

I appreciate all the work on this, but I think the best case is that this PR probably should be updated to move some of the analysis to sema, and at that it will be a placeholder that will be replaced... that may limit the amount of additional time you want to put into it at this stage.

Again, sorry, and thanks for understanding!

source/cppfront.cpp
filipsajdak: I found `std::visit()` a convenient way to extract the type from `decl_node->type` and pass a pointer to the appropriate overload.

My previous implementation consists of many `std::get_if()` with if statements that look like this (it is before moving to `type_id` node):
```cpp
// if it is a function check if it returns pointer type
if (auto* fun = std::get_if<declaration_node::function>(&decl->declaration->type)) {
    if (auto* ret_id_expr = std::get_if<function_type_node::id>(&(*fun)->returns)) {
        if (auto* unqual = std::get_if<id_expression_node::unqualified>(&(*ret_id_expr)->id)){
            return is_it_pointer_declaration(*unqual, deref_cnt, addr_cnt);
        }
    }
}

// if it is an object declaration check if we declare an pointer type
if (auto* obj_id_expr = std::get_if<declaration_node::object>(&decl->declaration->type)) {
    if (auto* unqual = std::get_if<id_expression_node::unqualified>(&(*obj_id_expr)->id)){
        return is_it_pointer_declaration(*unqual, deref_cnt, addr_cnt);
    }
}

// if we initialize an object with a function check if it is initialized with a pointer type
if (decl->initializer && decl->initializer->suspicious_initialization) {
    auto init_decl = sema.get_declaration_of(*decl->initializer->suspicious_initialization, true);
    if (init_decl && init_decl->declaration) {
        if (auto* fun = std::get_if<declaration_node::function>(&init_decl->declaration->type)) {
            if (auto* ret_id_expr = std::get_if<function_type_node::id>(&(*fun)->returns)) {
                if (auto* unqual = std::get_if<id_expression_node::unqualified>(&(*ret_id_expr)->id)){
                    return is_it_pointer_declaration(*unqual, deref_cnt, addr_cnt);
                }
            }
        }
    }
}
```

source/parse.h
filipsajdak: I have not checked that. I assumed that the lifetime will be the same as `pc_qualifiers` - need to double check that.

hsutter: Ah, I made a thinko. You're fine here for lifetime -- you're pointing to `token`s which outlive everything. For some reason I was thinking you were pointing to grammar nodes, but you aren't -- no worries then, sorry for the distraction.

source/parse.h
filipsajdak: I will check if I can move that to sema.

filipsajdak:
@hsutter Do you think I can try to start implementing a Lifetime static analysis? Or maybe I should try to move pointer reduction to sema? Or do you see anything else I can work on that will support you?

source/parse.h
filipsajdak: Yes, on the unqualified.

source/parse.h
filipsajdak: I have added comments to the code.

`suspicious_initialization` is initialization by a function call or variable that must be checked. Variable or function might be defined on the cpp1 side, which is why I have called it `suspicious`. On the other hand, it means that the type needs to be deduced, so maybe there is a better name.

source/cppfront.cpp
filipsajdak: The main change here is about calling a new function `is_pointer_declaration(unqual->identifier)` and to check not only the operator that is called on the node but also the results of the function call (e.g. `n()++`)

source/cppfront.cpp
jarzec: I personally think that `std::visit` is what allows to fully unleash the power of `std::variant`. In particular the idiom presented at the end of [the cppreference example](https://en.cppreference.com/w/cpp/utility/variant/visit#Example) leaves all the type checks to the compiler (i.e. no explicit pointers checked for nullity) and provides very nice syntactic sugar.
In fact `inspect` from Cpp2 made me wonder about a twist of that idiom that would be even nicer syntactic sugar with a more left-to-right taste. [Here is a quick example](https://github.com/hsutter/cppfront/compare/main...jarzec:cppfront:inspect-in-cpp1) of a version of `inspect` for Cpp1 with an example application in cppfront that passes all regression tests. I know from experience that this kind of syntax for handling `variant`s is used easily even by less experienced C++ users.

hsutter: Cool, thanks!

hsutter:
Thanks. On reviewing this, I think maybe it's okay not to move this to sema -- this would be a (useful) placeholder, but something that would be replaced in the future and is pretty localized now. So I would lean toward taking this pretty much as is, and not trying to go further on this path for now.

I appreciate the offer for implementing Lifetime, but that will go best after classes are done because parts of Lifetime depend on classes. Probably similar with most other things -- I think the most useful thing will be for me to focus on classes in the coming few months, and in the meantime if you can focus on letting me know about any bug reports and repros that you find that would be useful, and then after classes are in and stable it would be time to look at new features again.

source/parse.h
hsutter: On second thought this heuristic is probably fine for this point-in-time localized feature.

hsutter (approved):

regression-tests/test-results/pure2-deducing-pointers.cpp2.output
```diff
@@ -0,0 +1,12 @@
+pure2-deducing-pointers.cpp2...
+pure2-deducing-pointers.cpp2(14,8): error: = - pointer assignment from null or integer is illegal
```
hsutter, line 2: Thanks for including a new regression test.

Minor naming thing that I can clean up post-merge: Tests that are intended to fail cppfront compilation (not generate a .cpp file) should have a name that ends in `...error.cpp2`. That way my super-simple `run-tests.bat` test runner will count it as not intending to to create a `.cpp` and not flag a mismatch error. I'll take care of this post-merge though.

filipsajdak: Right... will correct that.

filipsajdak: Oh, I see that you have merged that. Should I prepare another PR with the name change?

filipsajdak: By the way. Thank you for merging that. Appreciate it!

hsutter: Yup, I did in another commit. All's well, thanks!

PR #203, opened 2023-01-06
Fix to_string(std::any) overload

### LINKED ISSUE #200: [BUG] to_string(...) hidden by to_string(std::any) overload
Having the code

```cpp
struct X {};

main: () -> int = {
    x: X = ();

    std::cout << "x  = '(x)$'" << std::endl;
}
```

### Expected

Output:
```
../tests/bug_cast_to_optional.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

x  = '(customize me - no cpp2::to_string overload exists for this type)'
```

### Actual

```
../tests/bug.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

x  = 'std::any'
```

### Cause

The `X` type matches:
```cpp
inline auto to_string(std::any const&) -> std::string {
  return "std::any";
}
```

### Possible fix

Replace the above overload to:
```cpp
template<typename T>
    requires std::is_same_v<T, std::any>
inline auto to_string(T const&) -> std::string {
  return "std::any";
}

```

filipsajdak:
Closes #200 

Having the code

```cpp
struct X {};

main: () -> int = {
    x: X = ();

    std::cout << "x  = '(x)$'" << std::endl;
}
```
Currently, end with:
```
../tests/bug.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

x  = 'std::any'
```
But should with:
```
../tests/bug.cpp2... ok (mixed Cpp1/Cpp2, Cpp2 code passes safety checks)

x  = '(customize me - no cpp2::to_string overload exists for this type)'
```

This change replace the problematic `cpp2::to_string(std::any)z overload to:
```cpp
template<typename T>
    requires std::is_same_v<T, std::any>
inline auto to_string(T const&) -> std::string {
  return "std::any";
}
```

And fix one regression test.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1114,7 +1114,7 @@ inline auto to_string(...) -> std::string {
     return "(customize me - no cpp2::to_string overload exists for this type)";
 }
 
-inline auto to_string(std::any const&) -> std::string {
+inline auto to_string(std::same_as<std::any> auto const&) -> std::string {
     return "std::any";
 }
 
--- a/regression-tests/test-results/clang-12/mixed-string-interpolation.cpp.execution
+++ b/regression-tests/test-results/clang-12/mixed-string-interpolation.cpp.execution
@@ -14,4 +14,4 @@ p = (first, (empty))
 t = (3.140000, (empty), (empty))
 vv = 0
 vv = (1, 2.300000)
-custom = std::any
+custom = (customize me - no cpp2::to_string overload exists for this type)


include/cpp2util.h
```diff
@@ -1114,7 +1114,9 @@ inline auto to_string(...) -> std::string {
     return "(customize me - no cpp2::to_string overload exists for this type)";
 }
 
-inline auto to_string(std::any const&) -> std::string {
+template<typename T>
+    requires std::is_same_v<T, std::any>
+inline auto to_string(T const&) -> std::string {
```
JohelEGP, line 1119:
```suggestion
inline auto to_string(std::same_as<std::any> auto const&) -> std::string {
```

JohelEGP (changes requested):

include/cpp2util.h
filipsajdak: Oh, I forgot that we could use concepts in this context. I need to use to it. I am correcting it.

filipsajdak: I pushed the corrected version. Thank you, @JohelEGP, for pointing that out!

hsutter:
Thanks!

PR #205, opened 2023-01-07
Update tests missed in PR #203

jarzec:
This PR updates tests that were missed in [PR #203](https://github.com/hsutter/cppfront/pull/203).
In the mentioned PR only the `clang-10` test was updated.


hsutter:
Thanks!

PR #206, opened 2023-01-07
Update regression tests for apple-clang-14 (23w01)

filipsajdak:
Update regression tests for apple-clang-14


hsutter:
Interesting... so Apple clang-14's libc++ recently got ranges?

filipsajdak:
@hsutter after you write that... I realized that my test script is treating files with ranges differently, and while syncing with other tests, they were not synced correctly.

I have fixed my script and pushed the fix for the regression tests -> #207

Apple Clang still has no ranges support. Sorry for the trouble.

PR #207, opened 2023-01-07
Fix regression tests for apple-clang-14 y23w01b

filipsajdak:
Apple clang 14 still has no ranges support - my test script error.

Sorry for trouble


filipsajdak:
Previously I had generated files that showed errors in compiling the files with ranges... I will restore them... wait a sec.

filipsajdak:
OK, now corrected. Ready to be merged.

Issue #208, opened 2023-01-07
[BUG] function with empty body but with a list of multiple/named return failed to generate


filipsajdak:
The code:
```cpp
fun: () -> (s : std::string = "Hello World!") = {}
```
fail to compile as it calls `body->statements.back()->position();` on empty `body->statements`.

hsutter:
Thanks. Yes, I just noticed (and fixed) this while investigating #210...

PR #209, opened 2023-01-07
Fix bug generating 'return;' for empty function

filipsajdak:
Closes https://github.com/hsutter/cppfront/issues/208

The current implementation of cppfront fails to compile
```cpp
fun: () -> (s : std::string = "Hello World!") = {}
```
as it calls `body->statements.back()->position();` on empty `body->statements`.

This change introduces a check if `body->statements` is empty and puts 'return' in the place of close braces.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -3347,8 +3347,14 @@ private:
 
             if (body->statements.empty() || body->statements.back()->statement.index() != statement_node::return_)
             {
-                auto last_pos = body->statements.back()->position();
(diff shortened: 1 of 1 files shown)


hsutter:
Sorry, just saw this and already fixed it here d747f86dd3742314843ede9886d5569af05e56fc (also closes #208 and #210). Thanks for suggesting a fix though!

filipsajdak:
@hsutter the most crucial thing is that it is fixed. 👍

Issue #212, opened 2023-01-08
[SUGGESTION] Add support for static variables


jarzec:
The title says it all, really - support for `static` variables is a simple Cpp1 feature that is not yet implemented in Cpp2.
I think that in order to preserve the beauty of the "left-to-rightness" the `static` qualifier should only be allowed as the left-most one, eps. that this qualifies the variable rather than its type.

hsutter:
There are global (namespace-scope) statics, class statics, and function-local statics.

For all three, there are potential concurrency issues, but sometimes you really do need a shared state so that part doesn't worry me unduly.

For the first two, though, it's important to avoid the ["static initialization order fiasco"](https://en.cppreference.com/w/cpp/language/siof) -- if they are allowed at all, they must not allow dynamic initialization. My current thinking is to either (a) allow them but make them always `constexpr` or optionally `constinit const`, or (b) don't allow them and tell programmers to instead wrap them in a function instead. 

For the third, I would either (a) support them, or (b) provide a `std::singleton<name, type>` type that does the work and is implemented in terms of `call_once` -- i.e., make it a library feature.

I've been resisting making any choices here yet because I feel good about dragging my feet on adding global shared state support until we see if it's really a must-have language feature (you can get the same effect with a library and I'd like to see the use cases that don't do well with a library), and also because I want to get a feel for classes first which are the next major feature to implement.

Cons-Cat:
> constinit const

Sorry if this is a noob question, but why would `constinit` without `const` not be okay? My understanding is that `consinit` calls a `constexpr` constructor so that there is no initialization at runtime. But modifying it afterwards should still be okay afaik, but I assume I'm missing something? And what if it has a `mutable` member, would that matter here?

hsutter:
@Cons-Cat Good question! Real quick answer: Because of concurrency safety and 'const by default.'

switch-blade-stuff:
> For the third, I would either (a) support them, or (b) provide a std::singleton<name, type> type that does the work and is implemented in terms of call_once -- i.e., make it a library feature.

To my knowledge, that is already what happens with function-local statics. At least on GCC/CLang there is an atomic sentinel flag that is used in conjunction with futex'es (or a mutex buffer if futex is unavailable) that is checked for dynamic initialization of local statics. Can't speak for MSVC though.

[...]

PR #213, opened 2023-01-08
Add support for static varaibles

jarzec:
This PR is my attempt at https://github.com/hsutter/cppfront/issues/212 with tests.


source/parse.h
```diff
@@ -519,6 +519,7 @@ struct type_id_node
 {
     source_position pos;
 
+    token const* static_qualifier = {};
```
jarzec, line 522: [Explanation] I could have easily included `static` in `pc_qualifiers` (that could be renamed). I decided against that as `static` is of somewhat different nature.

hsutter: Yes, that's the way I tend to handle these qualifier.

But I feel bad that I'm going to close this PR unmerged now, seeing all the work you've put into it -- as I mentioned on #212 I'm still thinking about what I want to do for `static` variables and I'd like to push them down the road, especially to get experience with classes first (and how the demands for static members unfolds there)... sorry, and thanks for understanding!

hsutter:
Closing for now per comment above and #212 -- thanks for understanding! I appreciate the work you put into this though and that you wrote the code thoughtfully and included lots of test cases.

jarzec:
I understand and agree with your remarks. The work is not lost - I got to understand better how the code works.
BTW. do you already have a time frame for simple classes? I think this is the last bit missing before cppfront can be rewritten in Cpp2 😉.

hsutter:
Now that I've caught up on issues and PRs (barring a new wave) I hope to be able to start looking at the next set of features... thanks again.

PR #218, opened 2023-01-10
Fix check of `typename C::value_type` where `C` is an `int`

filipsajdak:
https://github.com/hsutter/cppfront/commit/4693e52341cfcbd10b23ce05da9fc029667b62be introduce a check `std::is_convertible_v<X, typename C::value_type>` that failed when `C` is an `int`.

The issue was discovered when trying to compile the following code:
```cpp
main: () -> int = {
    x := 'a';
    u := x as int;

    std::cout << "u = '(u)$'" << std::endl;
}
```
cppfront succeeds but the cpp1 compiler fails with the error:
```
In file included from ../tests/bug_value_type_check.cpp:2:
include/cpp2util.h:883:92: error: type 'int' cannot be used prior to '::' because it has no members
    if constexpr( requires{ typename C::value_type; } && std::is_convertible_v<X, typename C::value_type> ) {
                                                                                           ^
include/cpp2util.h:1474:52: note: in instantiation of function template specialization 'cpp2::as<int, char>' requested here
    else if constexpr( std::is_same_v< CPP2_TYPEOF(as<C>(CPP2_FORWARD(x))), nonesuch_ > ) {
                                                   ^
../tests/bug_value_type_check.cpp2:3:19: note: in instantiation of function template specialization 'cpp2::as_<int, char &>' requested here
    auto u {cpp2::as_<int>(x)}; 
                  ^
1 error generated.
```

This change corrects this error and makes the code compile and run successfully.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -880,7 +880,7 @@ auto as( X const& x ) -> auto
     //  Experiment: Recognize the nested `::value_type` pattern for some dynamic library types
     //  like std::optional, and try to prevent accidental narrowing conversions even when
     //  those types themselves don't defend against them
-    if constexpr( requires{ typename C::value_type; } && std::is_convertible_v<X, typename C::value_type> ) {
+    if constexpr( requires { requires std::is_convertible_v<X, typename C::value_type>; } ) {
         if constexpr( is_narrowing_v<typename C::value_type, X>) {
             return nonesuch;
         }


include/cpp2util.h
```diff
@@ -880,7 +880,7 @@ auto as( X const& x ) -> auto
     //  Experiment: Recognize the nested `::value_type` pattern for some dynamic library types
     //  like std::optional, and try to prevent accidental narrowing conversions even when
     //  those types themselves don't defend against them
-    if constexpr( requires{ typename C::value_type; } && std::is_convertible_v<X, typename C::value_type> ) {
+    if constexpr( requires{ typename C::value_type; } && requires { std::is_convertible_v<X, typename C::value_type>; } ) {
```
hsutter, line 883: Thanks! I do see the need for that `requires`.

But doesn't it also still need the original `&& std::is_convertible_c<X, typename C::value_type>` test itself?

filipsajdak: I have some doubts about it. When I am adding it 
```cpp
    if constexpr( requires{ typename C::value_type; } && requires { std::is_convertible_v<X, typename C::value_type>; } && std::is_convertible_v<X, typename C::value_type> ) {
```
it fails again:
```
In file included from ../tests/bug_value_type_check.cpp:2:
include/cpp2util.h:883:158: error: type 'int' cannot be used prior to '::' because it has no members
    if constexpr( requires{ typename C::value_type; } && requires { std::is_convertible_v<X, typename C::value_type>; } && std::is_convertible_v<X, typename C::value_type> ) {
                                                                                                                                                             ^
include/cpp2util.h:1474:52: note: in instantiation of function template specialization 'cpp2::as<int, char>' requested here
    else if constexpr( std::is_same_v< CPP2_TYPEOF(as<C>(CPP2_FORWARD(x))), nonesuch_ > ) {
                                                   ^
../tests/bug_value_type_check.cpp2:3:19: note: in instantiation of function template specialization 'cpp2::as_<int, char &>' requested here
    auto u {cpp2::as_<int>(x)}; 
                  ^
1 error generated.
```

filipsajdak: I have run all the regression tests with proposed change and they pass.

JohelEGP: You need to have a single _requires-expression_ with multiple *requirement*s, so
```C++
requires{
  typename C::value_type;
  requires std::is_convertible_v<X, typename C::value_type>;
}
```

filipsajdak: It looks like it is working well as in the next line there is another use of `C::value_type` that is not causing trouble:
```cpp
        if constexpr( is_narrowing_v<typename C::value_type, X>) {
```

So, the proposed change is passing the right `C` types through.

JohelEGP: > You need to have a single _requires-expression_ with multiple _requirement_s, so
> 
> ```c++
> requires{
>   typename C::value_type;
>   requires std::is_convertible_v<X, typename C::value_type>;
> }
> ```

In fact, the first _requirement_ is redundant, so `requires { requires std::is_convertible_v<X, typename C::value_type>; }` should suffice. See https://compiler-explorer.com/z/cYoETj4K3.

```C++
#include <type_traits>
struct X {
  using value_type = int;
  operator int();
};
template<class T> constexpr bool f() {
  return requires { requires std::is_convertible_v<T, typename T::value_type>; };
}
static_assert(not f<int>());
static_assert(f<X>());
```

filipsajdak: That is true. Pushing this version.

filipsajdak:
@hsutter , I have just pushed the version proposed by @JohelEGP

include/cpp2util.h
filipsajdak: Thanks @JohelEGP !

hsutter: Thanks @JohelEGP and @filipsajdak for debugging this incantation!

PR #219, opened 2023-01-11
cpp2util.h: Fix typeid(T) -> Typeid<T> usage with -fno-rtti.

prenaux:
The no exception / no rtti modes seem to be working fine. Just this small fix needed.

Side note: I ended up defining CPP2_NO_EXCEPTIONS & CPP2_NO_RTTI in my build system. I did look at adding a preprocessor based auto-detection in cpp2util.h but it turns out that its compiler dependant so its a whole mess to do robustly, and it seems out of scope at least for now.


hsutter:
Looks good, thanks!

PR #229, opened 2023-01-16
Update regression tests for apple-clang-14 (23w02)

filipsajdak:
Weekly update of regression tests for apple-clang-14


hsutter:
Thanks!

PR #242, opened 2023-01-24
Fix parsing of hex and binary integer literals closes #240

### LINKED ISSUE #240: [BUG] Cppfront reports errors when initialising variables with hex & binary integer literals
**Describe the bug**
Cppfront reports errors when initialising variables with hex or binary integer literals.

**To Reproduce**
Run Cppfront with the following, based on code from [cppreference](https://en.cppreference.com/w/cpp/language/integer_literal):
```
main: () -> int = {
    d : i32 = 42;         // OK
    o : i32 = 052;        // OK
    x : i32 = 0x2a;       // Error
    X : i32 = 0X2A;       // Error
    b : i32 = 0b101010;   // Error
}
```

Cppfront reports these errors:
`error: ill-formed initializer (at 'a')` for x
`error: ill-formed initializer (at 'A')` for X
`error: ill-formed initializer (at '0')` for b

filipsajdak:
The current implementation was not skipping the last character of integer literal while doing lex. In the same loop, it was parsed as the identificator after successfully parsing it as literal.

### DIFF
--- a/source/lex.h
+++ b/source/lex.h
@@ -1053,6 +1053,7 @@ auto lex_line(
                     if (is_binary_digit(peek2)) {
                         while (is_separator_or(is_binary_digit,peek(j))) { ++j; }
                         store(j, lexeme::BinaryLiteral);
+                        continue;
                     }
                     else {
                         errors.emplace_back(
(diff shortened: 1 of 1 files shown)


hsutter:
Ah, good catch. Thanks!

PR #244, opened 2023-01-24
Add return forward passing style support

filipsajdak:
Based on chapter `1.5 Return values, D0708 R0` cppfront should support forward return passing style. The current implementeation makes it impossible to pass variable as a reference, or to forward value with preserving the return-expression’s cv-qualification and value category.

After this change it is possible to create the following function:
```cpp
first: (forward rng) -> forward _ = {
    [[assert Bounds: !std::empty(rng)]]
    return std::begin(rng)*;
}
```
that will (for lvalue rng) returns lvalue reference to the first element of the rng and will make the following code possible:
```cpp
main: () -> int = {
    v : std::vector = (1,2,3);
    first(v) = 4;
    std::cout << first(v) << std::endl; // prints: 4
}
```

Limitations:

* don't know if `passing_style::move` has a special way or is it current default one.
* the wildcar is used currently `forward _` to get `decltype(auto)`,
* wildcars make these function order dependent and need to be defined before it is used.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -2921,7 +2921,11 @@ public:
             printer.print_cpp2( " -> ", n.position() );
             auto& r = std::get<function_type_node::id>(n.returns);
             assert(r);
-            emit(*r);
+            if (r->is_wildcard() && n.returns_passing == passing_style::forward) {
+                printer.print_cpp2( "decltype(auto)", n.position() );
+            } else {
+                emit(*r);
+            }
         }
 
         else {
--- a/source/parse.h
+++ b/source/parse.h
@@ -1045,6 +1045,7 @@ struct function_type_node
         std::unique_ptr<type_id_node>,
         std::unique_ptr<parameter_declaration_list_node>
     > returns;
+    passing_style returns_passing = passing_style::move;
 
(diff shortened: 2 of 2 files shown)


filipsajdak:
The above code will generate the following cpp1 code:
```cpp
[[nodiscard]] auto first(auto&& rng) -> decltype(auto){
    cpp2::Bounds.expects(!(std::empty(rng)), "");
    return *cpp2::assert_not_null(std::begin(CPP2_FORWARD(rng))); 
}

    std::vector v {1, 2, 3}; 

    first(v) = 4;

    std::cout << first(std::move(v)) << std::endl;
}
```

hsutter:
Thanks! This has been on my list.

I'm taking your code and massaging it into a separate PR as that's likely faster than just asking for changes and iterating. Some additions I want are to support non-generic examples like this:

    f: () -> forward const int = { ... }

In my coming PR this will return `int const&`.

Since the `passing_style` should apply only to the single-type-id case, I'm bundling it up into that variant alternative instead of having it separate on its own with a comment that it's only used for that variant alternative.

filipsajdak:
OK. That is fine. If I can help with anything, just let me know.

hsutter:
This is great, thanks again!

PR #245, opened 2023-01-28
Fix capture of same variable twice - closes #237

### LINKED ISSUE #237: [BUG] one variable captured in lambda as two captures
After propose https://github.com/hsutter/cppfront/pull/236 (user-defined literals) I have played with `boost::ut` library.

I found that the following case:

```cpp
#include "boost/ut.hpp"
#include "html/html.h2"

using namespace boost::ut;
using namespace std::literals;

main: () -> int = {
    "html::element"_test = :() = {
        e: html::element = ("tag");

        "setting attribute"_test = :() = {
            mut(e$).set_attribute("name", "value");
            expect(eq(to_string(e$), "<tag name=\"value\"></tag>"sv));
        };
    };
}
```
generates (skipping boilerplate):
```cpp
    "html::element"_test = [](){
        html::element e {"tag"}; 

        "setting attribute"_test = [_0 = e, _1 = std::move(e)](){
            CPP2_UFCS(set_attribute, mut(_0), "name", "value");
            expect(eq(to_string(_1), "<tag name=\"value\"></tag>"sv));
        };
    };
}
```
That means that the `e` variable is captured twice in the capture list. I was trying to reproduce the bug without using of external library but currently did not manage to do it.

Moving `to_string()` call to the variable fix the problem - the following code:
```cpp
main: () -> int = {
    "html::element"_test = :() = {
        e: html::element = ("tag");

        "setting attribute"_test = :() = {
            mut(e$).set_attribute("name", "value");
            actual := to_string(e$);
            expect(eq(actual, "<tag name=\"value\"></tag>"sv));
        };
    };
}
```
generates:
```cpp
    "html::element"_test = [](){
        html::element e {"tag"}; 

        "setting attribute"_test = [_0 = e, _1 = std::move(e)](){
            CPP2_UFCS(set_attribute, mut(_0), "name", "value");
            auto actual {to_string(_0)}; 
            expect(eq(std::move(actual), "<tag name=\"value\"></tag>"sv));
        };
    };
}
```

It works, but there are still two elements on the capture list.

filipsajdak:
In the current implementation, when the move from the last use happens to a variable captured, it will be captured twice: one by copy and one moved.

That means that the following code:

```cpp
#include "boost/ut.hpp"
#include "html/html.h2"

using namespace boost::ut;
using namespace std::literals;

main: () -> int = {
    test("html::element") = :() = {
        e: html::element = ("tag");

        should("setting attribute") = :() = {
            mut(e$).set_attribute("name", "value");
            expect(eq(to_string(e$), std::string_view("<tag name=\"value\"></tag>")));
        };
    };
}
```
Will generate (skipping boilerplate):
```cpp
    test("html::element") = [](){
        html::element e {"tag"}; 

        should("setting attribute") = [_0 = e, _1 = std::move(e)](){
            CPP2_UFCS(set_attribute, mut(_0), "name", "value");
            expect(eq(to_string(_1), std::string_view("<tag name=\"value\"></tag>")));
        };
    };
}
```
This change makes cppfront identify that variable is moved, capture it once by the move, and use it multiple times. That means that the above code will generate the following:
```cpp
    test("html::element") = [](){
        html::element e {"tag"}; 

        should("setting attribute") = [_0 = std::move(e)](){
            CPP2_UFCS(set_attribute, mut(_0), "name", "value");
            expect(eq(to_string(_0), std::string_view("<tag name=\"value\"></tag>")));
        };
    };
}
```

### DIFF
--- a/regression-tests/test-results/mixed-function-expression-with-repeated-capture.cpp
+++ b/regression-tests/test-results/mixed-function-expression-with-repeated-capture.cpp
@@ -22,7 +22,7 @@
 
     auto y {"\n"}; 
     std::ranges::for_each
-        ( view, [_0 = y, _1 = std::move(y)](auto const& x){std::cout << _0 << x << _1;});
+        ( view, [_0 = std::move(y)](auto const& x){std::cout << _0 << x << _0;});
 
     auto callback {[](auto& x) { return x += "-ish"; }}; 
     std::ranges::for_each( view, std::move(callback));
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -1075,11 +1075,15 @@ public:
     //  is needed where Cpp1 and Cpp2 have different grammar orders
     //
 
-    auto print_to_string(auto& i, auto... more) -> std::string {
-        auto print = std::string{};
-        printer.emit_to_string(&print);
+    void print_to_string(std::string* str, auto& i, auto... more) {
+        printer.emit_to_string(str);
         emit(i, more...);
         printer.emit_to_string();
+    };
+
+    auto print_to_string(auto& i, auto... more) -> std::string {
+        auto print = std::string{};
+        print_to_string(&print, i, more...);
         return print;
     };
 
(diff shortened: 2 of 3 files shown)


filipsajdak:
There is only one change in regression tests - in `mixed-function-expression-with-repeated-capture.cpp` file. The following code:

```cpp
    y := "\n";
    std::ranges::for_each
        ( view, :(x:_) = { std::cout << y$ << x << y$; } );
```
Generates:
```cpp
     auto y {"\n"}; 
     std::ranges::for_each
         ( view, [_0 = std::move(y)](auto const& x){std::cout << _0 << x << _0;});
```
instead of the following:
```cpp
     auto y {"\n"}; 
     std::ranges::for_each
         ( view, [_0 = y, _1 = std::move(y)](auto const& x){std::cout << _0 << x << _1;});
```

filipsajdak:
I have added overload of the `print_to_string` function that takes a pointer to `std::string` that will make it possible to eliminate an additional bunch of calls to `printer.emit_to_string()` that need to match earlier call to `printer.emit_to_string(&str)`.

The new overload looks:
```cpp
    void print_to_string(std::string* str, auto& i, auto... more) {
         printer.emit_to_string(str);
         emit(i, more...);
         printer.emit_to_string();
     };
```

hsutter:
Looks good, thanks!

Issue #247, opened 2023-01-31
[QUESTION] Is `&$*` a new cpp2 idiom?


filipsajdak:
As there is no possibility of passing a reference to the anonymous function (aka. lambda), you need to capture a variable as a pointer. That makes a new idiom emerge from the code: `&$*` - capture variable by pointer (and use as a reference). 

The following code:
```cpp
main: () -> int = {
    i := 42;

    l := :() = {
        std::cout << i&$* << std::endl;
    };

    l();
}
```
Will generate (skipping boilerplate):
```cpp
[[nodiscard]] auto main() -> int{
    auto i {42}; 

    auto l {[_0 = (&i)](){                                    // capture by pointer
        std::cout << *cpp2::assert_not_null(_0) << std::endl; // dereference in place of use
    }}; 

    std::move(l)();
}
```

Or maybe I have missed something?

---

For all that wanders what `i&$*` means (please remember that cppfront uses postfix operators - https://github.com/hsutter/cppfront/wiki/Design-note%3A-Postfix-operators). I will add parentheses to explain:

`i&` - address of `i`,
`(i&)$` - `$` captures the value on the left (in that case, it is a pointer to `i`),
`( (i&)$ )*` - `*` dereference the value captured by `$` (in that case, it is a pointer to `i`)

So, effectively `i&$*` is a reference to `i` that is captured by a lambda.

filipsajdak:
The funny thing is that it looks like grawlix :D

AbhinavK00:
That makes me think, as there are no references in cpp2, how will we write functions that return a reference in cpp2? Like the [] operator for containers, or do we just start returning pointers in all cases?
Val also does not have references but they have projections which work by something called inversion of control, but I don't those can be implemented at a transpiler level or even if herb would like to implement them.

filipsajdak:
@AbhinavK00, that is an excellent question. It is already covered by the `forward` return type passing style.

```cpp
fun: (i : int) -> forward int = {
    return i;
}

fun2: (i) -> forward _ = {
    return i;
}
```
Generates (skipping boilerplate):
```cpp
[[nodiscard]] auto fun(cpp2::in<int> i) -> int&{ // this is bug should be const int&
    return i; 
}

[[nodiscard]] auto fun2(auto const& i) -> decltype(auto){
    return i; 
}
```

filipsajdak:
[...]

PR #251, opened 2023-02-08
Add cpp2 raw string literals support with interpolation

filipsajdak:
The current implementation of cppfront does not support raw string literals on the cpp2 side. The raw strings are supported only on the cpp1 side. This change introduces:

  * cpp2 raw string liteterals support,
  * support for all prefixes for string literals (`L`, `u8`, `u`, `U`),
  * string interpolation for cpp2 raw string literals (enable by adding `$` before raw-string-literal),

That makes the following code:
```cpp
main: () -> int = {
    i := 42;
    m : std::map<std::string, int> = ();
    m["one"] = 1;
    m["two"] = 2;

    str : std::string = "this is a string";

    raw_str : std::string = R"string(raw string without interpolation)string";

    raw_str_multi : std::string = R"test(this is raw string literal

that can last for multiple

lines)test";

    raw_str_inter : std::string = $R"test(this is raw string literal
that can last for multiple
lines
(i)$ R"(this can be added too)"
calculations like m["one"] + m["two"] = (m["one"] + m["two"])$ also works
("at the beginning of the line")$!!!)test";

    raw_str_inter_multi : std::string = $R"(

    )" + $R"((i)$)" + $R"((i)$)";

    std::cout << str << std::endl;
    std::cout << raw_str << std::endl;
    std::cout << raw_str_multi << std::endl;
    std::cout << raw_str_inter << std::endl;
    std::cout << raw_str_inter_multi << std::endl;
    std::cout << ($R"((m["one"])$.)" + $R"((m["two"])$.)" + $R"((m["three"])$.)" + $R"((i)$)") << std::endl;
}
```
Generates the following cpp1 code (skipping boilerplate):
```cpp
    auto i {42}; 
    std::map<std::string,int> m {}; 
    cpp2::assert_in_bounds(m, "one") = 1;
    cpp2::assert_in_bounds(m, "two") = 2;

    std::string str {"this is a string"}; 

    std::string raw_str {R"string(raw string without interpolation)string"}; 

    std::string raw_str_multi {R"test(this is raw string literal

that can last for multiple

lines)test"}; 

    std::string raw_str_inter {R"test(this is raw string literal
that can last for multiple
lines
)test" + cpp2::to_string(i) + R"test( R"(this can be added too)"
calculations like m["one"] + m["two"] = )test" + cpp2::to_string(cpp2::assert_in_bounds(m, "one") + cpp2::assert_in_bounds(m, "two")) + R"test( also works
)test" + cpp2::to_string("at the beginning of the line") +  R"test(!!!)test"}; 

    std::string raw_str_inter_multi {R"(

    )" + cpp2::to_string(i) + cpp2::to_string(i)}; 

    std::cout << std::move(str) << std::endl;
    std::cout << std::move(raw_str) << std::endl;
    std::cout << std::move(raw_str_multi) << std::endl;
    std::cout << std::move(raw_str_inter) << std::endl;
    std::cout << std::move(raw_str_inter_multi) << std::endl;
    std::cout << (cpp2::to_string(cpp2::assert_in_bounds(m, "one")) + R"(.)" + cpp2::to_string(cpp2::assert_in_bounds(m, "two")) + R"(.)" + cpp2::to_string(cpp2::assert_in_bounds(std::move(m), "three")) + R"(.)" + cpp2::to_string(std::move(i))) << std::endl;
}
```

All regression tests pass.

### Limitations

cppfront accepts all prefixes for string literals (regular string literals and raw string literals). Unfortunately, wide characters are not supported by string interpolation - the generated code is OK, but the `cpp2::to_string()` functions produce `std::string`, incompatible with strings that use other character types.

### DIFF
--- a/source/common.h
+++ b/source/common.h
@@ -210,6 +210,13 @@ private:
     };
 };
 
+struct raw_string
+{
+    source_position start;
+    source_position end;
+    std::string     text;
+};
+
 //-----------------------------------------------------------------------
 //
 //  error: represents a user-readable error message
--- a/source/common.h
+++ b/source/common.h
@@ -100,6 +100,116 @@ struct comment
     std::string     text;
 };
 
+struct string_parts {
+    struct cpp_code   { std::string text; };
+    struct raw_string { std::string text; };
+    enum adds_sequences { no_ends = 0, on_the_begining = 1, on_the_end = 2, on_both_ends = 3 };
+
+    string_parts(const std::string& beginseq,
+                 const std::string& endseq,
+                 adds_sequences     strateg) 
+     : begin_seq{beginseq}
+     , end_seq{endseq}
+     , strategy{strateg}
+    {
+        if (!(strategy & on_the_begining)) {
+            parts.push_back(raw_string{""});
+        }
+    }
+
+    void add_code(const std::string& text) { parts.push_back(cpp_code{text});}
+    void add_string(const std::string& text) { parts.push_back(raw_string{text});}
+    void add_string(const std::string_view& text) { parts.push_back(raw_string{std::string(text)});}
+
+    void clear() { parts.clear(); }
+
+
+    auto generate() const -> std::string {
+        
+        if (parts.empty()) { 
+            return (strategy & on_the_begining ? begin_seq : std::string{}) 
+                 + (strategy & on_the_end ? end_seq : std::string{}); 
+        }
+
+        auto result = std::visit(begin_visit{begin_seq, strategy}, 
+                                 parts.front());
+
+        if (std::ssize(parts) > 1) { 
+            auto it1 = parts.cbegin();
+            auto it2 = parts.cbegin()+1;
+            for(;it2 != parts.cend(); ++it1, ++it2) {
+                result += std::visit(generator_visit{begin_seq, end_seq}, *it1, *it2);
+            }
+        }
+
+        if (!(strategy & on_the_end)) {
+            result += std::visit([this](const auto& lhs) {
+                return generator_visit{begin_seq, end_seq}(lhs, raw_string{""});
+            }, parts.back());
+        }
+
+        result += std::visit(end_visit{end_seq, strategy}, parts.back());
+
+        return result;
+    }
+
+    auto is_expanded() const -> bool {
+        for (const auto& p : parts) {
+            if (std::holds_alternative<cpp_code>(p)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+private:
+    std::string     begin_seq;
+    std::string     end_seq;
+    adds_sequences  strategy;
+    std::vector<std::variant<raw_string, cpp_code>> parts;
+
+    struct begin_visit {
+        std::string begin_seq;
+        adds_sequences strategy;
+
+        auto operator()(const raw_string& part) const -> std::string {
+            return (strategy & on_the_begining ? begin_seq : "") + part.text;
+        }
+        auto operator()(const cpp_code& part) const -> std::string {
+            return part.text;
+        }
+    };
+
+    struct end_visit {
+        std::string end_seq;
+        adds_sequences strategy;
+        auto operator()(const raw_string&) const -> std::string {
+            return strategy & on_the_end ? end_seq : "";
+        }
+        auto operator()(const cpp_code&) const -> std::string {
+            return {};
+        }
+    };
+
+    struct generator_visit {
+        std::string begin_seq;
+        std::string end_seq;
+
+        auto operator()(const raw_string&, const cpp_code& part ) const -> std::string {
+            return end_seq + " + " + part.text;
+        }
+        auto operator()(const cpp_code&, const raw_string& part ) const -> std::string  {
+            return " + " + begin_seq + part.text;
+        }
+        auto operator()(const raw_string&, const raw_string& part ) const -> std::string  {
+            return part.text;
+        }
+        auto operator()(const cpp_code&, const cpp_code& part ) const -> std::string  {
+            return " + " + part.text;
+        }
+    };
+};
+
 //-----------------------------------------------------------------------
 //
 //  error: represents a user-readable error message
--- a/source/common.h
+++ b/source/common.h
@@ -216,6 +216,13 @@ struct raw_string
     std::string     text;
     std::string     opening_seq;
     std::string     closing_seq;
+    bool            should_interpolate = false;
+};
+
+struct multiline_raw_string
+{
+    std::string     text;
+    source_position end = {0, 0};
 };
 
 //-----------------------------------------------------------------------
--- a/source/common.h
+++ b/source/common.h
@@ -213,8 +213,9 @@ private:
 struct raw_string
 {
     source_position start;
-    source_position end;
     std::string     text;
+    std::string     opening_seq;
+    std::string     closing_seq;
 };
 
 //-----------------------------------------------------------------------
--- a/source/lex.h
+++ b/source/lex.h
@@ -435,6 +435,70 @@ auto expand_string_literal(
     return parts.generate();
 }
 
+auto expand_raw_string_literal(
+    const std::string& opening_seq,
+    const std::string& closing_seq,
+    string_parts::adds_sequences closing_strategy,
+    std::string_view text,
+    std::vector<error>& errors,
+    source_position src_pos) -> string_parts
+{
+    auto const length = std::ssize(text);
+    auto pos = 0;
+    auto current_start   = pos;   // the current offset before which the string has been added to ret
+    string_parts parts{opening_seq, closing_seq, closing_strategy};
+
+    //  Now we're on the first character of the string itself
+    for ( ; pos < length; ++pos )
+    {
+        //  Find the next )$
+        if (text[pos] == '$' && text[pos-1] == ')')
+        {
+            //  Scan back to find the matching (
+            auto paren_depth = 1;
+            auto open = pos - 2;
+
+            for( ; open > current_start; --open)
+            {
+                if (text[open] == ')') {
+                    ++paren_depth;
+                }
+                else if (text[open] == '(') {
+                    --paren_depth;
+                    if (paren_depth == 0) {
+                        break;
+                    }
+                }
+            }
+            if (text[open] != '(')
+            {
+                errors.emplace_back(
+                    source_position( src_pos.lineno, src_pos.colno + pos ),
+                    "no matching ( for string interpolation ending in )$"
+                );
+                return parts;
+            }
+
+            //  'open' is now at the matching (
+
+            //  Put the next non-empty non-interpolated chunk straight into ret
+            if (open != current_start) {
+                parts.add_string(text.substr(current_start, open - current_start));
+            }
+            //  Then put interpolated chunk into ret
+            parts.add_code("cpp2::to_string" + std::string{text.substr(open, pos - open)});
+
+            current_start = pos+1;
+        }
+    }
+
+    //  Put the final non-interpolated chunk straight into ret
+    if (current_start < std::ssize(text)) {
+        parts.add_string(text.substr(current_start));
+    }
+
+    return parts;
+}
 
 //-----------------------------------------------------------------------
 //  lex: Tokenize a single line while maintaining inter-line state
@@ -453,6 +517,8 @@ auto expand_string_literal(
 //  into a whitespace-containing token (to merge the Cpp1 multi-token keywords)
 //  -- this isn't about tokens generated later, that's tokens::generated_tokens
 static auto generated_text = std::deque<std::string>{};
+
+static auto multiline_raw_strings = std::deque<multiline_raw_string>{};
 
 auto lex_line(
     std::string&            mutable_line,
@@ -888,6 +954,49 @@ auto lex_line(
         return do_is_keyword(multi_keys);
     };
 
+    auto reset_processing_of_the_line = [&]() {
+        //  Redo processing of this whole line now that the string is expanded,
+        //  which may have moved it in memory... move i back to the line start
+        //  and discard any tokens we already tokenized for this line
+        i = colno_t{-1};
+        while (
+            !tokens.empty()
+            && tokens.back().position().lineno == lineno
+            )
+        {
+            tokens.pop_back();
+        }
+    };
+
+    auto interpolate_raw_string = [&](
+        const std::string& opening_seq,
+        const std::string& closing_seq,
+        string_parts::adds_sequences closing_strategy,
+        std::string_view part, 
+        int pos_to_replace, 
+        int size_to_replace
+    ) -> bool {
+        auto parts = expand_raw_string_literal(opening_seq, closing_seq, closing_strategy, part, errors, source_position(lineno, pos_to_replace + 1));
+        auto new_part = parts.generate();
+        mutable_line.replace( pos_to_replace, size_to_replace, new_part );
+        i += std::ssize(new_part)-1;
+
+        if (parts.is_expanded()) {
+            // raw string was expanded and we need to repeat the processing of this line
+            reset_processing_of_the_line();
+
+            // but skipping end of potential multiline raw string that ends on this line
+            if (!multiline_raw_strings.empty() && multiline_raw_strings.back().end.lineno == lineno) {
+                i = multiline_raw_strings.back().end.colno;
+                raw_string_multiline.reset();
+            } else if (raw_string_multiline && raw_string_multiline->start.lineno == lineno) {
+                raw_string_multiline.reset();
+            }
+            return true;
+        }
+        return false;
+    };
+
     //
     //-----------------------------------------------------
 
@@ -917,6 +1026,7 @@ auto lex_line(
                 else if (peek1 == 'R' && peek2 == next)                 { return 3; } // LR" 
             } 
             else if (line[i] == 'R' && peek1 == next)                   { return 2; } // R"
+            else if (line[i] == '$' && peek1 == 'R' && peek2 == next)   { return 3; } // $R"
             return 0;
         };
 
@@ -946,22 +1056,36 @@ auto lex_line(
             auto end_pos = line.find(raw_string_multiline.value().closing_seq, i);
             auto part = line.substr(i, end_pos-i);
 
+            if (const auto& rsm = raw_string_multiline.value(); rsm.should_interpolate) {
+                
+                auto closing_strategy = end_pos == line.npos ? string_parts::no_ends : string_parts::on_the_end;
+                auto size_to_replace  = end_pos == line.npos ? std::ssize(line) - i  : end_pos - i + std::ssize(rsm.closing_seq);
+
+                if (interpolate_raw_string(rsm.opening_seq, rsm.closing_seq, closing_strategy, part, i, size_to_replace ) ) {
+                    continue;
+                }
+            }
+            // raw string was not expanded
+
             raw_string_multiline.value().text += part;
             if (end_pos == std::string::npos) {
                 raw_string_multiline.value().text += '\n';
                 break;
             }
 
-            // here we know that we are dealing with multiline raw string literal
-            // token needs to use generated_text to store string that exists in multiple lines
-            i = end_pos+std::ssize(raw_string_multiline.value().closing_seq)-1;
+            // here we know that we are dealing with finalized multiline raw string literal
+            // token needs to use multiline_raw_strings to store string that exists in multiple lines
             raw_string_multiline.value().text += raw_string_multiline.value().closing_seq;
 
-            generated_text.push_back(raw_string_multiline.value().text);
+            // and position where multiline_raw_string ends (needed for reseting line parsing)
+            i = end_pos+std::ssize(raw_string_multiline.value().closing_seq)-1;
+
+            const auto& text = raw_string_multiline.value().should_interpolate ? raw_string_multiline.value().text.substr(1) : raw_string_multiline.value().text;
+            multiline_raw_strings.emplace_back(multiline_raw_string{ text, {lineno, i} });
 
             tokens.push_back({
-                &generated_text.back()[0],
-                std::ssize(generated_text.back()),
+                &multiline_raw_strings.back().text[0],
+                std::ssize(multiline_raw_strings.back().text),
                 raw_string_multiline.value().start,
                 lexeme::StringLiteral
             });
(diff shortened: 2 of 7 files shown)


filipsajdak:
I found some bugs with this patch. Working on a patch to fix it.

filipsajdak:
@hsutter, the code is fixed. All tests passed. I have added a test covering edge cases I found along the way.

I decided to add `raw_strings` static variable similar to `generated_text`. I was thinking of using `generated_text`, but I am also using `start` and `end` `source_position`s. When I recognize the raw string literal, I am adding new `raw_string` to the `raw_strings` container, and I use it in the next steps - that ensures that the `start` & `end` variables are set to proper values.

neumannt:
I don't think interpolation in raw string literals makes sense. After all, you use raw string literals to avoid quoting. If you start interpreting the content of a raw string that defeats the purpose of raw strings. cpp2 should support raw strings, of course, but it should not interpret the content but pass it along as is.

filipsajdak:
@neumannt you are right. This is misleading at a minimum. I like the possibility of writing string without escaping the quotation marks that will interpolate. But raw string should be a raw string 😄 

Point taken. Maybe I will propose a way to enable interpolation.

hsutter:
Just spitballing here, I want to think about this more:

I can see value in "fully raw" string literals, and "raw plus interpolation" string literals.

One option could be to have the usual `R"..."` syntax for raw strings, and to enable interpolation put a `$` immediately before the `"` without whitespace (e.g., `R$"..."`).

Of course, doing that would resurrect previous requests that normal `"..."` strings not be interpolated either, and for consistency now also be prefixed with `$"..."` to enable interpolation. I've resisted that, but I do realize that would follow the existing practice in other languages, which seems to be fine and usable, to enable interpolation in string literals via a prefix in C# (`$"..."`) and Python (`f'...'`).

Requiring the `$` up front to enable interpolation also has the benefit of declaring intent up front. That can be good for readability, so the code readers (who outnumber the code author) don't have to read through a long string to see if there are any interpolations. And it can be good for diagnostics, because we could emit a warning if a `$"` string doesn't contain an interpolation ("did you forget...?", probably there would be no false positives but some users might be annoyed at "having to add the `$` to shut up the compiler") and possibly also if a non-`$"` string contains what would be an interpolation ("did you mean...?", but this one could have false positives and I would expect be more likely to annoy users).

I know we've discussed the above prefix-to-enable-interpolation idea before in another issue, and I resisted it. But now that we have the additional use case of raw string literals, where there also seems to be a reason to let users distinguish string literals that allow interpolation and ones that don't, it's an additional data point that might point toward "string literals 'want to' be prefixed for interpolation"...

hsutter:
Also: Do we really need the entire 2x2 matrix of { raw, non-raw } x { with-interpolations, no-interpolations } ?

In particular, is there a need for { non-raw, with-interpolations } ? If not, then `$"..."` could imply raw as well as allow interpolations, if those two things should naturally go together. I don't have enough experience yet with using the feature to have a good feel for the answer.

filipsajdak:
I have worked a little bit more on that and currently prepared an implementation that uses `$` as a first character of d-char-sequence
```
prefix(optional) R"d-char-sequence(optional) (r-char-sequence(optional))d-char-sequence(optional)"
```

That means that 
```cpp
i := 42;
rs1 := R"(raw string: (i)$)";         // this will not interpolate, pure raw string
rs2 := R"seq(raw string: (i)$)seq";   // this will not interpolate, pure raw string

rs3 := R"$(raw string: (i)$)$";       // this will interpolate
rs4 := R"$seq(raw string: (i)$)$seq"; // this will interpolate
```

Currently, `$` is not allowed in `d-char-sequence` but probably will be in C++26 thanks to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2558r0.html (in my current code when I detect it I replace it to `S` to make it work with current compilers).

This is my current state of work.

@hsutter, as I understand you correctly, you propose adding a `$` prefix for all strings that need to be interpolated (even if they are regular string literals and raw string literals), right?

So, the above example can be rewritten to:

```cpp
i := 42;

s1  := "string: (i)$";               // this will not interpolate, pure raw string
rs1 := R"(raw string: (i)$)";        // this will not interpolate, pure raw string
rs2 := R"seq(raw string: (i)$)seq";  // this will not interpolate, pure raw string

s2  := $"string: (i)$";              // this will interpolate
rs3 := $R"(raw string: (i)$)";       // this will interpolate
rs4 := $R"seq(raw string: (i)$)seq"; // this will interpolate
```

Did I get it right? If yes, I can rework my code to adjust it to this logic.

I like that there will be one rule instead of two.

hsutter:
Yes, thanks. I like that you changed the order, you're right that `$` before `R` is clearer.

filipsajdak:
Ok, I will adjust the code.

switch-blade-stuff:
Resurrecting old topic, but since the "interpolation --enabled" literals in this proposal are different from the non-interpolating ones, why not consider the `$"{}"` syntax? This could also allow for (`fmt`/`std::format` compatible) formatting support to be added, something like this `$"Some int: {some_int} Some int (hex): {some_int:#x}`. Less characters to type, looks cleaner, etc.

This also makes it more consistent with other languages, python especially, since this formatting syntax is originally from python afaik.

Having to use `()$` *after* already specifying `$"` seems like too much dollars :)

switch-blade-stuff:
As for formatting, it could be enabled only if `std::format` is available (otherwise, only `{id}` would work). Considering that Cpp2 does not have to deal with backwards-compatibility and that the required compiler versions are currently the most "bleeding edge" ones, it seems like a reasonable thing to do.

Would also neatly integrate a C++20 feature and reduce potential fragmentation, otherwise you'll have another competing choice for string formatting: Cpp2 interpolation, `std::format`, iostreams and old school `printf`.
If Cpp2 interpolation uses the same mechanism as `std::format`, it can be reduced to just 3 (technically 2, since `printf` is discouraged).

hsutter:
Understood, but please remember the considerations in [Design note: Capture](https://github.com/hsutter/cppfront/wiki/Design-note:-Capture#q-why-use-postfix--for-capture-wouldnt---be-nicer-for-string-interpolation-like-python)

switch-blade-stuff:
> Understood, but please remember the considerations in [Design note: Capture](https://github.com/hsutter/cppfront/wiki/Design-note:-Capture#q-why-use-postfix--for-capture-wouldnt---be-nicer-for-string-interpolation-like-python)... having a single way to spell a thing consistently is important to me, and I'm currently exploring the path of making interpolation be the same syntax and meaning as all other captures, which is why it looks like a language operator (it is, it's writing an expression inside the string).

The thing is, if interpolated strings require a prefix, this already breaks the lambda parallel. And IMHO, it would be better to have 100% consistency with how other *strings* and formatting works (both in C++20 and pther languages), rather than lambdas that have nothing to do with strings (since lambdas capture state, strings just print it).

switch-blade-stuff:
Also creating competing standards is to me the opposite of Cpp2's goal of cleaning up and simplifying the language

hsutter:
@filipsajdak For now please do the `$` prefix only for raw strings (`$R"` or `R$"` perhaps), to designate a raw string with interpolation. I'm not yet comfortable with requiring `$"` to start every ordinary string literal that uses interpolation, and don't want to cross that bridge yet. Thanks again for the PR.

jarzec:
I have a very naïve question concerning string interpolation. This a new concept from a C++ point of view and one that can (at least to some extent) be replaced by existing `std::format`. If I am not mistaken there was an idea for cppfront to avoid adding things that can be achieved through a library. To paraphrase Sugarman: `std::format` is the answer that makes the questions disappear 😉: i.e. one thing less to teach, no escape characters (other than what already exists), no extra pre-/postfixes, no extra escape characters, no trouble with interoperability with e.g. with regexps, less trouble with parsing/transpilation, ...
Let alone that libraries are easier to fix/extend than the core syntax.

On the other hand you might tell me that string interpolation in cppfront can work as a proof of concept of a possible proposal for ISO C++.

JohelEGP:
IIRC, it's about generality. See https://github.com/hsutter/cppfront/wiki/Design-note%3A-Capture.

gregmarr:
>If I am not mistaken there was an idea for cppfront to avoid adding things that can be achieved through a library

You can't do the same thing with a library.  `std::format` gives you this, where you still need to make sure that your replacement fields in the string line up with your arguments.

```
  auto mystring = std::format("Good morning {}, the time is {}, and the weather today is {}.", name, time, weather);
```

String interpolation gives you this:

```
    auto mystring = "Good morning (name)$, the time is (time)$, and the weather today is (weather)$.";
```

Now, with language support, you can convert the second into the first, maybe even we can standardize localization hooks at some point, so you get the compiler converting it into something like this:

```
    auto mystring = std::format(std::localize("Good morning {0}, the time is {1}, and the weather today is {2}."), name, time, weather);
```

>you might tell me that string interpolation in cppfront can work as a proof of concept of a possible proposal for ISO C++.

That basically describes the mission of cppfront.

switch-blade-stuff:
> You can't do the same thing with a library.

Correct me if i am mistaken, but doesn't `std::format` alao support named arguments? Or is it an `fmt` feature that did not get carried over?

gregmarr:
> Correct me if i am mistaken, but doesn't std::format alao support named arguments?

No, just by index.

gregmarr:
```
    print("You clicked {button} at {x},{y}.", arg("button", "b1"), arg("x", 50), arg("y", 30));
```
I'm not quite sure that this qualifies, anyway.  You still need to make sure that your `arg()` strings match your format string, and it's a lot more boilerplate.

jarzec:
> > You can't do the same thing with a library.
> 
> Correct me if i am mistaken, but doesn't `std::format` alao support named arguments? Or is it an `fmt` feature that did not get carried over?

I think it is only `fmt`, at least for now 😉.

jarzec:
> ```
>     print("You clicked {button} at {x},{y}.", arg("button", "b1"), arg("x", 50), arg("y", 30));
> ```
> 
> I'm not quite sure that this qualifies, anyway. You still need to make sure that your `arg()` strings match your format string, and it's a lot more boilerplate.

It is true that compile-time checks are a strong plus.

hsutter:
> IIRC, it's about generality. See https://github.com/hsutter/cppfront/wiki/Design-note%3A-Capture.

Yes... in a nutshell, I really want to see if interpolation is (can naturally be) just another case of capture, rather than a special feature that works only in strings.

filipsajdak:
Sorry for delay on that. I spend my evening on that. One case left to be covered - leaving it for tomorrow. 

The good thing is I have integrated my code with `generated_code` not to duplicate `std::deque` container to store multi line raw string.

filipsajdak:
@hsutter I have finished correcting this one - sorry it took so long. 

TLDR:
1. The change brings support for raw string literals to cpp2.
```cpp
    raw_str_multi : std::string = R"test(this is raw string literal

("that can last for multiple")$

lines)test";
```
will generate:
```cpp
    std::string raw_str_multi {R"test(this is raw string literal

("that can last for multiple")$

lines)test"}; 
```

2. The interpolation of raw string can be triggered by adding `$` in front of it:

```cpp
    raw_str_multi : std::string = $R"test(this is raw string literal

("that can last for multiple")$

lines)test";
```
will generate
```cpp
    std::string raw_str_multi {R"test(this is raw string literal

)test + cpp2::to_string("that can last for multiple") + R"test(

lines)test"}; 
```

filipsajdak:
I choose `$R"` if you don't like it I can rework it to `R$"`

JohelEGP:
The last one is just one string literal. It's missing a `"` before ` + cpp::to_string`.

filipsajdak:
> The last one is just one string literal. It's missing a `"` before ` + cpp::to_string`.

copy & paste issue - Thank you @JohelEGP for spotting this!

hsutter:
@JohelEGP do you feel like doing a code review? I'll assign you just in case (I'm learning to delegate, in my old age), but just reply if you'd rather not be tagged and that's fine too.

JohelEGP:
I'm available. I can look for general problems with in the Cpp1 code of the compiler. But I'm not familiar with the code base or Cpp2 code, so I probably won't find issues specific to them, if that's what you're looking for.

filipsajdak:
I can split my commits in any way that will make it easy to understand (I already did that to some extent, but I can go even further).

gregmarr:
https://github.com/hsutter/cppfront/pull/251/commits/8ba4039b72fd49a359fc0e76e47f8c22fb6e95ae#diff-53c80455692db08228e06ab3a94959cc5e3b4cb5c8f325f2407707f3c8dcb5f0R901

Should probably update this comment:
        //G encoding-prefix: one of
        //G     'u8' 'u'

I'm having a hard time determining either `$u8R` or `u8$R` is valid.  It doesn't look like they are, but I might be missing something.

filipsajdak:
@gregmarr the only valid one is `$R"` just because interpolation use `cpp2::to_string()` function that produce `std::string` which is not compatible with `u8R`, `u`, `uR`, `u8`, `U`, `UR`, `L`, `LR` prefixes.

That is why the function matches only `$R` prefix for interpolation. The rest prefixes are used for raw-string literals without interpolation.

You are right about the comment it should be adjusted.

filipsajdak:
If other prefixes will be needed we can make `cpp2::to_string()` function a template e.g. `cpp2::to_string<CharT>()` that could be handled by interpolation for various of string literals.

filipsajdak:
* I have moved the check for `$R"` out of the `is_encoding_prefix_and()` function and moved it directly to a place where we match `$` in `lex_line()` (https://github.com/hsutter/cppfront/commit/a6a2348ae7e4f3a016858dfc25cb38350f6525cd#diff-53c80455692db08228e06ab3a94959cc5e3b4cb5c8f325f2407707f3c8dcb5f0R1167). 
* The comment to the function is updated.
* I have added this change as another commit, not to change previous commits that you might check already - after the review, I can rebase everything to make clear commit history.

I have rebased to the current main.

filipsajdak:
I usually preffer to rewrite the history of commits but please tell me what will make you life easier.

gregmarr:
>@gregmarr the only valid one is $R" just because interpolation use cpp2::to_string() function that produce std::string

Okay, that makes sense, probably enough for now.

filipsajdak:
I have found one issue that is already corrected (https://github.com/hsutter/cppfront/compare/a6a2348ae7e4f3a016858dfc25cb38350f6525cd..0455de13b658c3526bfc58a45af70349b495be3d) - in cases where raw-string literal starts in the middle of the line and the same line there was an interpolation, cppfront generated wrong cpp1 code.

msadeqhe:
In string literal the symbol `"` is already a special character and needs to be escaped. What if we just use `"` for string interpolation!? To do this, CPP2 has to support automatic joining `"..."` and `(...)$` expressions. For example, I suggest to change the syntax from:
```
x := "His name is (name)$!";
```
to:
```
x0 := "His name is "(name)$"!";
```
Now, the above line can be written as the following lines too (optionally all of them are the same):
```
x1 := "His name is "      (name)$               "!";
x2 := "His name is " name$ "!";
x3 := "His name is "name$"!";
```
For that to work, CPP2 has to just automatically join _capture expressions_ and _string literals_ together.
More examples:
```
y := "I saw "name$" yesterday!";
z := "The result of 2 * 2 is "(2 * 2)$", and you knew it".
```

By this way, for raw string literals we can write:
```
rawstr := R"I saw "name$R" yesterday!";
```
But if CPP2 supports automatically joining `"..."` and `...$` expressions, it may be better to change the raw string literal from `R"..."` notaion to `"(...)"` notation or something similar:
```
rawstr := "(I saw )"name$"( yesterday!)";
```

msadeqhe:
Also to avoid defining a new raw-sting literal, CPP2 can just accept escape sequences between or before of after a string literal, but an escape sequence cannot be used alone without string literal:
```
x := "First line\nSecond line"; // \n is not new-line
a := "First line"\n"Second line"; // \n is new-line
b := "First line"\n; // \n is new-line
c := \n"Second line"; // \n is new-line
d := \n; // ERROR
e := ""\n; // \n is new-line
```

msadeqhe:
Also string literals can be more integrated to the language for [reflection and generation](https://github.com/hsutter/cppfront/wiki/Design-note%3A-Capture) example:
```
(func.name()+"_wrapper")$: (forward func.params.first().name$: _) = { do_wrapped_extra_stuff(); func.name()$(func.params.first().name$); } }
```
If CPP2 allows directly using string literal for function name, then it can be changed to:
```
func.name()$"_wrapper": (forward func.params.first().name$: _) = { do_wrapped_extra_stuff(); func.name()$(func.params.first().name$); } }
```

msadeqhe:
Finally, combination of string literals can have prefix, the prefix of the first string literal determines the behavior of the combined string literal. For example if CPP2 supports `u8` prefix:
```
x := u8"First "second$" Last"\n;
```
then `u8` will be applied to all `First`, `second$`, `Last` and `\n`. Think about the middle `"` characters are just visual separators like digit grouping `1'234'258` or `1_234_258` which brain can ignore them.

msadeqhe:
I forgot to mention that all string literals will be raw strings and we can break them into several lines:
```
x := "First name: "first$\n"Last name: "last$\n"Age: "age$\n"Sex: "sex$;

y := "First name: "first$"
Last name: "last$"
Age: "age$"
Sex: "sex$;

z := "First name: "first$\n
     "Last name: "last$\n
     "Age: "age$\n
     "Sex: "sex$;
```
All `x`, `y`, `z` have the same value.

filipsajdak:
@msadeqhe Thank you for your sugestion. I am not eager to experiment with this more - I am trying to add features that are missing in cppfront that are present in cpp1 and I am trying to follow papers that Herb is mentioning (in the end is all about C++ and not the new thing).

If you'd like to play with it, please consider how it will impact the other features, e.g., captures in the lambda. (check here: https://github.com/hsutter/cppfront/issues/247) or capture in contracts:
(https://github.com/hsutter/cppfront/blob/fb4a9fc4bf43ec17e0c9d6c863221032c344677f/regression-tests/mixed-postexpression-with-capture.cpp2#L16). I think all use cases are collected here: https://github.com/hsutter/cppfront/wiki/Design-note:-Capture#q-why-use-postfix--for-capture-wouldnt---be-nicer-for-string-interpolation-like-python

Thanks!

msadeqhe:
Thank you for clarification and sorry if I shouldn't write the suggestion here. I'll open an issue for it with more detailed information, by the way, I think @hsutter won't accept it :cold_sweat:.

filipsajdak:
Your feedback is always welcome! I just want to fix all the bugs and align that code with Herb style to be merged with the main branch.

filipsajdak:
Rebased to newest changes - all regression tests passed.

source/common.h
```diff
@@ -100,6 +100,131 @@ struct comment
...
+        }
+        auto operator()(const cpp_code& part) const -> std::string {
+            return part.text;
+        }
+    };
+
+    struct end_visit {
+        std::string end_seq;
+        adds_sequences strategy;
+        auto operator()(const raw_string& part) const -> std::string {
```
hsutter, line 186: Thanks, I tried applying this manually and it passes regressions and the test case. The only thing I noticed it that it doesn't build without warnings...

Here's the first one: Unreferenced parameter? Did you mean to use `part`, or is it not needed?

source/common.h
```diff
@@ -100,6 +100,131 @@ struct comment
...
+            return strategy & on_the_end ? end_seq : "";
+        }
+        auto operator()(const cpp_code& part) const -> std::string {
```
hsutter, line 189: Ditto

source/lex.h
```diff
@@ -366,25 +365,8 @@ auto expand_string_literal(
     auto first_quote_pos = pos;
```
hsutter, line 365: This variable is no longer needed, right? Seems to be unused now

source/lex.h
```diff
@@ -451,18 +429,78 @@ auto expand_string_literal(
...
+    const std::string& opening_seq,
+    const std::string& closing_seq,
+    string_parts::adds_sequences closing_strategy,
+    std::string_view text,
+    std::vector<error>& errors,
+    source_position src_pos) -> string_parts
+{
+    auto const length = std::ssize(text);
+    auto pos = 0;
+    auto first_quote_pos = pos;
```
hsutter, line 449: Ditto, unused?

source/common.h
filipsajdak: Checking...

filipsajdak: OK. I have checked it. I only need the type of the part (`raw_string` or `cpp_code`) name of the variable is not required.

I am preparing fix.

source/lex.h
filipsajdak: Ah, that is an artifact from the previous implementation - I have corrected my compiler flags to spot unused variables.

Do you have a list of flags that you have enabled during compilation? I will make sure I have the same set of flags set not to send faulty changes.

I have prepared the fix and I am running local tests to ensure that I did not break anything.

source/lex.h
filipsajdak: Yes, true. Sorry for not paying attention to details.

source/lex.h
filipsajdak: Fixed

source/lex.h
filipsajdak: Fixed

source/common.h
filipsajdak: Fixed

source/common.h
filipsajdak: Fixed

filipsajdak:
Apply review comments. All regression tests passes.

hsutter:
Thanks Filip!

source/lex.h
```diff
@@ -1284,54 +1472,80 @@ auto lex_line(
 
                 //G string-literal:
                 //G     encoding-prefix? '"' s-char-seq? '"'
+                //G     encoding-prefix? 'R"' d-char-seq? '(' s-char-seq? ')' d-char-seq? '"'
```
JohelEGP, line 1475: Shouldn't this be prefixed with `'$'?`? Or is that somehow implied by the apparently unused `token` grammar?
```Cpp2
//G token:
//G     identifier
//G     keyword
//G     literal
//G     operator-or-punctuator
//G
//G operator-or-punctuator:
//G     operator
//G     punctuator

//G punctuator: one of
//G     '...' '.'
//G     '::' ':'
//G     '{' '}' '(' ')' '[' ']' ';' ',' '?' '$'
```

JohelEGP: Actually, see https://github.com/hsutter/cppfront/issues/387#string-literal.

PR #260, opened 2023-02-19
patch for -Wshadow-field-in-constructor

hypatia-of-sva:
This is a small patch that makes cpp2 code compile under -Wshadow, specifically -Wshadow-field-in-constructor. The only other remaining conflicts with clang's -Weverything (besides the obvious C++98 compatibilty options) are reserved identifieres with leading _ (which should be fine in a system level header) and -Wctad-maybe-unsupported, which says there should be a deduction guide for c_raii and fopen, which I would have to further look into how to implement

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -546,11 +546,11 @@ class out {
     bool called_construct_ = false;
 
 public:
-    out(T*                 t) noexcept :  t{ t}, has_t{true}      { Default.expects( t); }
-    out(deferred_init<T>* dt) noexcept : dt{dt}, has_t{false}     { Default.expects(dt); }
-    out(out<T>*           ot) noexcept : ot{ot}, has_t{ot->has_t} { Default.expects(ot);
-        if (has_t) {  t = ot->t;  }
-        else       { dt = ot->dt; }
+    out(T*                 t_object) noexcept :  t{ t_object}, has_t{true}      { Default.expects( t_object); }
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks for your pull request! It looks like this may be your first contribution to cppfront. Before I can look at your pull request, you'll need to sign a Contributor License Agreement (CLA). Please send me your email address (by private email is fine), and I can send you one.

hypatia-of-sva:
Hello Herb Sutter,

my current email address is hypatia.sva@posteo.eu.

Hypatia of Sva.

SebastianTroy:
This is publically visible!


On 19 February 2023 21:39:45 hypatia-of-sva ***@***.***> wrote:

Hello Herb Sutter,

my current email address is ***@***.******@***.***>.

Hypatia of Sva.

hypatia-of-sva:
Hello,

yes, I'm sorry, I didn't know that that's a problem for you. I have the 
e mail in my header file comments already so I thought it would be okay.
Regardless of that, if it works to answer to this e mail I will do so 
from now on, I just wasn't sure what email to use for you since I didn't 
find that field in the github profile.
Do you think its okay or would I have to use a different address because 
of it?

Hypatia of Sva.

Am 20.02.2023 00:43 schrieb Sebastian Troy:
> This is publically visible!
> 
>

hsutter:
Thanks! I deleted the comments in this thread with/about the email address as it didn't seem you intended them to be visible here. I'll follow up in private email.

hypatia-of-sva:
Okay, then thats fine

Issue #270, opened 2023-03-10
[BUG] `in` argument passing to function that uses cpp2 user defined type failed with error


filipsajdak:
258190ef6df8db8bc31d0eb5d94094da5d800623 introduces the possibility of defining user-defined types in cpp2. Unfortunately, it fails with a free functions that use UDT as `in` parameters.

The following code:
```cpp
X : type = {
    i : int = 42;
}

fun: (x: X) = {
    std::cout << x.i << std::endl;
}
```
generates:
```cpp

#define CPP2_USE_MODULES         Yes

#include "cpp2util.h"


#line 1 "../tests/bug_incomplete_type.cpp2"
class X;
#line 5 "../tests/bug_incomplete_type.cpp2"
auto fun(cpp2::in<X> x) -> void;

//=== Cpp2 definitions ==========================================================

#line 1 "../tests/bug_incomplete_type.cpp2"
class X   {
    private: int i {42}; 
};

auto fun(cpp2::in<X> x) -> void{
    std::cout << x.i << std::endl;
}
```
It fails on cpp1 compilation with the error:
```
In file included from ../tests/bug_incomplete_type.cpp:4:
include/cpp2util.h:496:5: error: invalid application of 'sizeof' to an incomplete type 'X'
    sizeof(T) < 2*sizeof(void*) && std::is_trivially_copy_constructible_v<T>;
    ^~~~~~~~~
include/cpp2util.h:501:9: note: in instantiation of variable template specialization 'cpp2::prefer_pass_by_value<X>' requested here
        prefer_pass_by_value<T>,
        ^
../tests/bug_incomplete_type.cpp2:5:16: note: in instantiation of template type alias 'in' requested here
auto fun(cpp2::in<X> x) -> void;
               ^
../tests/bug_incomplete_type.cpp2:1:7: note: forward declaration of 'X'
class X;
      ^
In file included from ../tests/bug_incomplete_type.cpp:4:
include/cpp2util.h:501:9: error: non-type template argument is not a constant expression
        prefer_pass_by_value<T>,
        ^
../tests/bug_incomplete_type.cpp2:5:16: note: in instantiation of template type alias 'in' requested here
auto fun(cpp2::in<X> x) -> void;
               ^
2 errors generated.
```

[...]

PR #276, opened 2023-03-12
Add check if type is complete to before checking sizeof type in prefer_pass_by_value check

filipsajdak:
In the current implementation of `prefer_pass_by_value`, we check the size of the type. It fails in case that type is incomplete. This change adds a check if there is a possibility to check the size of the type.

This might be a temporary solution but it can help avoid failed builds.


JohelEGP:
To help avoid failed builds, I think this should be a `requires` check.
In its current form, it seems like a source for ODR violations.

JohelEGP:
> To help avoid failed builds, I think this should be a `requires` check.

Or more like fail with a more meaningful error message.

The current formulation still fails because the use of `sizeof`, even if not evaluated due to the short-circuit, doesn't permit an incomplete type. See https://compiler-explorer.com/z/j9zs675YY.

SebastianTroy:
I get that practically this needs to build so needs must, but as cpp2 is intended to be order independant, would a type being declared but not defined be an issue for end users to worry about still?


On 12 March 2023 22:20:55 Johel Ernesto Guerrero Peña ***@***.***> wrote:

To help avoid failed builds, I think this should be a requires check.

Or more like fail with a more meaningful error message.

The current formulation still fails because the use of sizeof, even if not evaluated due to the short-circuit, doesn't permit an incomplete type. See https://compiler-explorer.com/z/j9zs675YY.

JohelEGP:
The error message seems shorter and clearer if `in` directly has the _requires-clause_. https://compiler-explorer.com/z/K4KcjeTqj.

You can see how the meaning of `in` changes depending on the completeness of `X` at https://compiler-explorer.com/z/57oK9jEhj by (de)commenting `void f(in<X>);`. That probably runs afoul of https://eel.is/c++draft/temp.res.general#6 or some other rule. As you can see, the meaning of `in<X>` is fixed to `const X&` if it's used while `X` is incomplete.

filipsajdak:
@JohelEGP Thank you for spotting that!  What about this: https://compiler-explorer.com/z/9b7bcEf8c

```cpp
template<typename T, typename = void>
constexpr bool prefer_pass_by_value = false;

template<typename T>
constexpr bool prefer_pass_by_value<T, std::void_t<decltype(sizeof(T))>> =
    sizeof(T) < 2*sizeof(void*)
    && std::is_trivially_copy_constructible_v<T>;
```

Can you tell me why it might be an ODR violation?

filipsajdak:
@JohelEGP  Just to clarify, the order of https://github.com/hsutter/cppfront/issues/270#issuecomment-1464660533 should look like following:

```cpp
// declaration of classes (UDT)
struct X {
    int i;
    double d;

    void fun();
    double gun();
};

// declarations of free functions
void f(in<X>);

// implementations of class functions
void X::fun() {
    // code
}

double X::gun() {
    return i * d;
}

// implementations of free functions
void f(in<X>) { }
```

filipsajdak:
My issue is that I have a project divided to several `.h2` files that compiles to `.h` and `.hpp` files. When I include the `.h2` file in `.cpp2` file it is later replaced with `.h` file in declaration section and `.hpp` in definitions sections.

My `.h2` file looks like the following:
```cpp
#include <map>

namespace execspec {
    using config_t = std::map<std::string, std::string>;
}

execspec: namespace = {

    process_stage : type = {
        public name: std::string;
        public cmd:  std::string;
    }

    spec: type = {
        name:        std::string;
        description: std::string = "";
        config:      execspec::config_t = ();
        stages:      std::vector<execspec::process_stage> = ();

        operator=: (out this, n: std::string) = {
            name = n;
        }

        get_name:        (this) -> forward _ = name;
        get_description: (this) -> forward _ = description;
        get_config:      (this) -> forward _ = config;
        get_stages:      (this) -> forward _ = stages;

// skipping some code
    }
}

debug: (inout o : std::ostream, spec : execspec::spec) -> forward std::ostream = {
    o << "# (spec.get_name())$\n";

// skipping some code

    return o;
}
```
it generates `.h` file (skipping boilerplate):
```cpp
#include <map>

namespace execspec {
    using config_t = std::map<std::string, std::string>;
}

#include "cpp2util.h"

#line 7 "spec.h2"
namespace execspec   {
    class process_stage;
    class spec;
};
#line 49 "spec.h2"
[[nodiscard]] auto debug(std::ostream& o, cpp2::in<execspec::spec> spec) -> std::ostream&;
```
`.hpp` file has all definitions that normally would be generated in `.cpp` file when compiling above code as `.cpp2` file.

I have an issue when I include `spec.h2` file in other `.cpp2` file - in my case I include it in `spec_test.cpp2` file that looks following:
```cpp
#include "boost/ut.hpp"
using namespace boost::ut;

#include <sstream>
using namespace std::literals;

#include "execspec/spec.h2"

main: () -> int = {
    execspec__spec : suite = :() = {
        "create a spec"_test = :() = {
            t := execspec::spec("spec name"s);

            o : std::ostringstream = ();
            o.debug(t);
            
            expect(eq(o.str(), "# spec name\n\n"sv));
        };
    };
}
```
it generates:
```cpp
#include "boost/ut.hpp"
using namespace boost::ut;

#include <sstream>
using namespace std::literals;

#include "execspec/spec.h"

#include "cpp2util.h"


#line 9 "spec_tests.cpp2"

#include "execspec/spec.hpp"
#line 8 "spec_tests.cpp2"

    suite execspec__spec {[]() -> void{
        "create a spec"_test = []() -> void{
            auto t {execspec::spec("spec name"s)}; 

            std::ostringstream o {}; 
            CPP2_UFCS(debug, o, t);

            expect(eq(CPP2_UFCS_0(str, o), "# spec name\n\n"sv));
        };
    };
}
```
And that fail with an error:
```
[build] In file included from spec_tests.cpp2:7:
[build] In file included from spec.h2:7:
[build] cppfront/include/cpp2util.h:496:5: error: invalid application of 'sizeof' to an incomplete type 'execspec::spec'
[build]     sizeof(T) < 2*sizeof(void*) && std::is_trivially_copy_constructible_v<T>;
[build]     ^~~~~~~~~
[build] cppfront/include/cpp2util.h:501:9: note: in instantiation of variable template specialization 'cpp2::prefer_pass_by_value<execspec::spec>' requested here
[build]         prefer_pass_by_value<T>,
[build]         ^
[build] spec.h2:49:49: note: in instantiation of template type alias 'in' requested here
[build] [[nodiscard]] auto debug(std::ostream& o, cpp2::in<execspec::spec> spec) -> std::ostream&;
[build]                                                 ^
[build] spec.h2:9:11: note: forward declaration of 'execspec::spec'
[build]     class spec;
[build]           ^
[build] In file included from spec_tests.cpp2:7:
[build] In file included from spec.h2:7:
[build] cppfront/include/cpp2util.h:501:9: error: non-type template argument is not a constant expression
[build]         prefer_pass_by_value<T>,
[build]         ^
[build] spec.h2:49:49: note: in instantiation of template type alias 'in' requested here
[build] [[nodiscard]] auto debug(std::ostream& o, cpp2::in<execspec::spec> spec) -> std::ostream&;
[build]                                                 ^
[build] 2 errors generated.
```

With my current patch it compiles fine.

filipsajdak:
> I get that practically this needs to build so needs must, but as cpp2 is intended to be order independant, would a type being declared but not defined be an issue for end users to worry about still? 

@SebastianTroy eventually, it will not be a problem - it is just a matter of generating the output files in a specific order - example here: https://github.com/hsutter/cppfront/pull/276#issuecomment-1465352371. The current issue is caused by the introduction of classes (which is an immature implementation).

The issue is more problematic when you split your project into multiple `.cpp2` and `.h2` files. Most of our current examples are simple `.cpp2` files.

filipsajdak:
Taking another look at this... I think that an ODR violation might be an issue here. Probably I will need to temporarily change the meaning of `in<T>` to always `T const&` unless someone has a better idea.

JohelEGP:
> Can you tell me why it might be an ODR violation?

I wish I could give a methodical answer with proof and references. But I used _might_ for a reason. The gist is usually that `T<U>` shouldn't change meaning. At https://compiler-explorer.com/z/3aGq5KzcG is a valid, non ODR violating (AFAIK) formulation that demonstrates the change in meaning. (See https://eel.is/c++draft/temp.arg.general#9, https://compiler-explorer.com/z/98o99r6v9 for https://eel.is/c++draft/temp.inst#7.sentence-2 [likely a GCC bug]).

JohelEGP:
> Taking another look at this... I think that an ODR violation might be an issue here. Probably I will need to temporarily change the meaning of `in<T>` to always `T const&` unless someone has a better idea.

https://compiler-explorer.com/z/ffo1E8ohf. Making an exposition-only type that's complete earlier.

PR #303, opened 2023-03-28
Fix string literal parsing when \\ is on the end, closes #301

### LINKED ISSUE #301: [BUG] Back-slash before capture doesn't work in string literals
This is the sample code:

```
main :() = {
    x := 0;
    y := "\\(x)$";
}
```

C++2 should generate the following C++1:

```
auto main() -> int{
    auto x {0}; 
    auto y {"\\" + cpp2::to_string(x)}; 
}
```

But the compiler fails with the following error message:

```
# Compilation provided by Compiler Explorer at https://godbolt.org/
<Compilation failed>
# Compiler exited with result code 139
Standard error:
cppfront: source/lex.h:409: std::string cpp2::expand_string_literal(std::string_view, std::vector<error>&, source_position): Assertion `pos == length-1 && text[pos] == '"'' failed.
```

filipsajdak:
The issue is that when there is '\\' on the end of the string `expand_string_literal()` function goes out of string range because for loop has a check
```cpp
pos < length && (text[pos] != '"' || text[pos-1] == '\\')
```
that is true when '\\' is before last '"' character that cause loop to go beyond it and cause assertion to fail.

The current change adds a check if the '\' character before last '"' is not escaped by the former '\' character.

Closes #301 - all regression tests pass

### DIFF
--- a/source/lex.h
+++ b/source/lex.h
@@ -380,7 +380,7 @@ auto expand_string_literal(
     //  Now we're on the first character of the string itself
     for (
         ;
-        pos < length && (text[pos] != '"' || text[pos-1] == '\\');
+        pos < length && (text[pos] != '"' || (text[pos-1] == '\\' && pos>=2 && text[pos-2] != '\\'));
         ++pos
         )
     {


hsutter:
Looks good, thanks!

filipsajdak:
I start to become a cppfront string specialist 😉

hsutter:
At this point you've probably written as much of the string literal handling as I have, maybe more. 😁 🙏

PR #309, opened 2023-03-29
Fix: Remove unused offset variable

filipsajdak:
After removing adjust_remaining_token_columns_on_this_line_visitor (396e33)
offset variable was left and was used only by this visitor.

The current change removes that variable.

All regression tests pass.


hsutter:
Thanks! I just noticed that earlier in the week and made this change as part of the next large commit I'm about to push today, so I'll close this and do it in that commit.

PR #331, opened 2023-04-08
Fix removing parentheses when using deducing assignement, closes #327

### LINKED ISSUE #327: [BUG] parentheses missing when emitting binary expression tree
**Describe the bug**
When cpp2 code uses parentheses to explicitly specify the precedence of operations, the order of operations is correctly parsed into the expression tree. But when the tree is lowered to cpp, the parentheses are omitted.

**To Reproduce**
1. Sample code - distilled down to minimal essentials please
```c++
main: () -> int = {
    d := (1 + 2) * 3;
}
```
produces
```c++
    auto d {1 + 2 * 3}; 
}
```
2. Command lines including which C++ compiler you are using
transpiling with `cppfront -p main.cpp` then compiling with gcc12
2. Expected result - what you expected to happen
I would expect the result to be
```c++
    auto d {(1 + 2) * 3}; 
}
```
where the parentheses from the original expression are preserved. Or possibly even:
```c++
    auto d {((1 + 2) * 3)}; 
}
```
where expressions are fully (redundantly) parenthesised.
3. Actual result/error
parentheses are omitted leading to a change in the order of execution of binary operations.

**Additional context**
Double parentheses in the cpp2 code work around the issue:
```c++
main: () -> int = {
    d := ((1 + 2)) * 3;
}
```
produces
```c++
    auto d {(1 + 2) * 3}; 
}
```

I doubt this is the ideal solution, but just as additional context, the following modifications appear to fully parenthesise binary expressions and I'm currently using cppfront with this "fix" applied:
```patch
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -3244,13 +3244,14 @@ public:

         //  Else if this is an assignment expression, don't add std::move on the lhs
         //  even if this is a definite last use (only do that when an rvalue is okay)
-        if (
-            !n.terms.empty()
-            && is_assignment_operator(n.terms.front().op->type())
-            )
+        if (!n.terms.empty())
         {
-            suppress_move_from_last_use = true;
+            if (is_assignment_operator(n.terms.front().op->type()))
+               suppress_move_from_last_use = true;
+            else
+               printer.print_cpp2("(", n.position());
         }
+
         emit(*n.expr);
         suppress_move_from_last_use = false;

@@ -3320,6 +3321,12 @@ public:
                 emit(*x.expr);
             }
         }
+
+        if ((!n.terms.empty())
+          && !is_assignment_operator(n.terms.front().op->type()))
+        {
+            printer.print_cpp2(")", n.position());
+        }
     }
```

filipsajdak:
The issue is caused by `emit(expression_list_node)` that skips parentheses when a node is inside the initializer - that serves cases like:
```cpp
v : std::vector<int> = (1,2,3);
```
Where it generates:
```cpp
std::vector<int> v{1,2,3};
```

When `:=` is used in the following cases:
```cpp
d := (1 + 2) * (3 + 4) * (5 + 6);
```
It removes the first parentheses, and we end up with the following:
```cpp
auto d = {1 + 2 * (3 + 4) * ( 5 + 6)};
```

This change corrects this behavior on the parsing side. After parsing
the expression list it checks if the next lexeme is Semicolon. If it is it
means that we are on the initializer of the form:
```cpp
d1 := ((2 + 1) * (4 - 1) * (8 - 3));
d3 := (move d2);
v : std::vector<int> = ();
```
And we can suppress printing of parentheses - as there will be braces:
```cpp
auto d1 {(2 + 1) * (4 - 1) * (8 - 3)};
auto d3 {std::move(d2)};
std::vector<int> v {};
```

When the next lexeme is not `Semicolon` it means that we are in an initializer
of the form:
```cpp
d2 := (2 + 1) * (4 - 1) * (8 - 3);
d4 : _ = (1 + 2) * (3 + 4) * (5 + 6);
d5 : int = (1 + 2) * (3 + 4) * (5 + 6);
```
And we need to keep all the parentheses, and it will be generated to:
```cpp
auto d2 {(2 + 1) * (4 - 1) * (8 - 3)};
auto d4 {(1 + 2) * (3 + 4) * (5 + 6)};
int d5 {(1 + 2) * (3 + 4) * (5 + 6)};
```

Closes #327. All regression tests pass.

There is only one difference in generated code in `pure2-ufcs-member-access-and-chaining.cpp2` test. The following line:
```cpp
    res := (42).ufcs();
```
Generates after this change:
```cpp
    auto res {CPP2_UFCS_0(ufcs, (42))}; // previously: auto res {CPP2_UFCS_0(ufcs, 42)}; 
```

This is correct behavior, as the parentheses are not required in this context so they should be left. I have checked the following code:
```cpp
    r1 := 42.ufcs();
    r2 := (42).ufcs();
```
That will generate after this change:
```cpp
    auto r1 {CPP2_UFCS_0(ufcs, 42)}; 
    auto r2 {CPP2_UFCS_0(ufcs, (42))}; 
```

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -2852,7 +2852,9 @@ private:
             //      || curr().type() == lexeme::LeftBrace
             )
         {
-            bool inside_initializer = (peek(-1)->type() == lexeme::Assignment);
+            bool inside_initializer = ( 
+                peek(-1) && peek(-1)->type() == lexeme::Assignment
+            );
             auto open_paren = &curr();
             auto close = close_paren_type(open_paren->type());
             auto close_text = [&] () -> std::string { if (close == lexeme::RightParen) { return ")"; } return "}"; }();
@@ -2870,6 +2872,9 @@ private:
             }
             expr_list->close_paren = &curr();
             next();
+            if (curr().type() != lexeme::Semicolon) {
+                expr_list->inside_initializer = false;
+            } 
             n->expr = std::move(expr_list);
             return n;
         }


hsutter:
Thanks! I'm not sure this is the right fix yet, two reasons:

1) I noticed there are also cases like

    x : int = (1+2) * 3;
    x : _ = (1+2) * 3;

that still have incorrect code gen with this fix.

2) The code gen difference you mention is that code gen like this today

    auto z {std::move(x)}; 

with this fix becomes instead

    auto z {(std::move(x))}; 

and we should be avoiding generating extra parens (they're often innocuous, but I'm concerned about ending up hitting the comma operator).

filipsajdak:
Will work on that.

filipsajdak:
@hsutter I found the solution, and the current proposed solution solves all the cases.

There is only one difference in generated code in `pure2-ufcs-member-access-and-chaining.cpp2` test. The following line:
```cpp
    res := (42).ufcs();
```
Generates after this change:
```cpp
    auto res {CPP2_UFCS_0(ufcs, (42))}; // previously: auto res {CPP2_UFCS_0(ufcs, 42)}; 
```

This is correct behavior, as the parentheses are not required in this context so they should be left. I have checked the following code:
```cpp
    r1 := 42.ufcs();
    r2 := (42).ufcs();
```
That will generate after this change:
```cpp
    auto r1 {CPP2_UFCS_0(ufcs, 42)}; 
    auto r2 {CPP2_UFCS_0(ufcs, (42))}; 
```

I have updated the description of this PR.

filipsajdak:
There is one issue left: shall we rename `inside_initializer`?

hsutter:
Thanks! Trying it out now...

BTW in case I wasn't clear in my comment about hitting the comma operator, I meant that when there are multiple parameters, I want to avoid accidentally hit a case where a list of multiple things like `{1, 2}` accidentally becomes a list of a single thing `{(1, 2)}` because the comma turns into a comma operator and the `1` gets thrown away.

hsutter:
OK, looks good -- including that now the one change in the code gen is actually a correction, that's always a good sign.

Thanks! Merging...



> There is one issue left: shall we rename `inside_initializer`?

What would be a better name, `inside_assignment`?

filipsajdak:
`inside_initializer` is fine but it only is used in cases when it is in the following form:
```cpp
a : type = ( /* any complex expression */ );
```

So, it is more about the initializer being surrounded entirely by parenthesis.

filipsajdak:
@hsutter I have just realized that we have anonymous objects that we can create in place e.g. as a function argument. Need to check but probably my change broke the following code:
```cpp
fun(:std::vector=(1,2,3));
```
Need to check it.

filipsajdak:
Sorry, this change brake:
```cpp
fun(:std::vector=(1,2,3));
```
Generates:
```cpp
fun(std::vector{(1, 2, 3)});
```

Looking for a fix... and we need to add a regression test for in-place creation of objects.

filipsajdak:
@hsutter OK, one way to fix is to replace (in `parse.h` line 2875):
```cpp
            if (curr().type() != lexeme::Semicolon) {
                expr_list->inside_initializer = false;
            } 
```
with
```cpp
            if (
                   curr().type() != lexeme::Semicolon
                && curr().type() != lexeme::RightParen 
                && curr().type() != lexeme::RightBracket 
                && curr().type() != lexeme::Comma
            ) {
                expr_list->inside_initializer = false;
            } 
```
That should cover cases:
```cpp
    f1(:std::vector=(1,2,3));
    f2(:std::vector=(1,2,3),:std::vector=(4,5));
    ar[:index=(1,2)];
    ar2[:index=(1,2), :index=(2,1)];
```
that will generate:
```cpp
    f1(std::vector{1, 2, 3});
    f2(std::vector{1, 2, 3}, std::vector{4, 5});
    cpp2::assert_in_bounds(ar, index{1, 2});
    cpp2::assert_in_bounds(ar2, index{1, 2}, index{2, 1});
```

Are there any other cases that we should cover in-placed created objects?

PR #342, opened 2023-04-09
Provide hint for `auto` declaration in Cpp2 code

ntrel:
AFAIK, no Cpp2 declaration starts with an `auto` keyword, so forbid it.
Like the hints for `namespace`/`class` keywords.

Without this:
```c++
f: () = {
    ...
    auto x = 1; // causes ill-formed initializer error but line is first line of `f`, not this line
    auto [x,y] = (1,2); // passes cppfront but errors in C++1
}
```
With this, both `auto` declarations are caught with correct line numbers:
```
test.cpp2(8,10): error: unknown declaration (use `name: type = initializer` instead?)
test.cpp2(9,10): error: unknown declaration (use `name: type = initializer` instead?)
```
BTW I have emailed Herb requesting a CLA form.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -5453,9 +5453,13 @@ private:
                 && *id->get_token() == "auto"
                 )
             {
+                auto name = std::string{"v"};
+                if (peek(0) && peek(0)->type() == lexeme::Identifier) {
+                    name = peek(0)->to_string(true);
+                }
                 errors.emplace_back(
                     curr().position(),
-                    "unknown declaration (use `name: type = initializer` instead?)"
+                    "to define a variable " + name + " of type T, write '" + name + ": T = /* initializer */'"
                 );
                 return {};
             }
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks, this looks good! CLA is on the way.

Could you please update it to follow the style in the immediately following check for `namespace`? That is, use the next token (if there is one) in the error message, and make the error message follow that pattern... something like `"to define an object " + name + ", write '" + name + " : /*type*/ = /*initializer*/;', or '" + name + " : /*type*/;' and initialize the object later"`. 🙏

ntrel:
Now done, it prints:
```
test.cpp2(8,10): error: to define a variable x of type T, write 'x: T = /* initializer */'
test.cpp2(9,10): error: to define a variable v of type T, write 'v: T = /* initializer */'
```
The simple case parses the identifier, the `auto [x, y]` case just defaults to `v`.

hsutter:
Looks good, thanks!

PR #348, opened 2023-04-10
Add diagnostics for using auto as a type of variable and as function return type

filipsajdak:
While parsing the following code:
```cpp
f: () -> auto = 42;
```
Rise an error with the following diagnostics:
```
error: to define a function f with deduced return type, write 'f : ( /* arguments / ) -> _ = { / function body */ }'
```

While parsing the following code:
```cpp
i : auto = 42;
```
Rise an error with the following diagnostics:
```
error: to define a variable i with deduced type, write 'i := /* initializer */;'
```

All regression tests pass. Continuation of #342 (increasing QoI)

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -4894,6 +4894,21 @@ private:
                 }
             }
             else if (auto t = type_id()) {
+                if (
+                    t->get_token()
+                    && t->get_token()->to_string(true) == "auto"
+                    )
+                {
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks! These are good helpful QoI diagnostics for common mistakes when coming from Cpp1 syntax, great stuff. 👍 🙏

Issue #346, opened 2023-04-10
[BUG] _expression-statement_ of unnamed templated function is a `template`


JohelEGP:
A `template` was generated, not a lambda expression.

**Minimal reproducer** (https://godbolt.org/z/56cecbYMh):
```Cpp2
f: () = :<T> () = {};
```

**Commands**:
```bash
cppfront x.cpp2
clang++17 -std=c++2b -stdlib=libc++ -I $CPPFRONT_INCLUDE_DIR x.cpp
```

**Expected result**: A lambda expression when lowering to C++.

**Actual result and error**:

<details><summary>Generated C++1.</summary>

```C++


//=== Cpp2 type declarations ====================================================


#include "cpp2util.h"


//=== Cpp2 type definitions and function declarations ===========================

#line 1 "x.cpp2"
auto f() -> void;

//=== Cpp2 function definitions =================================================

#line 1 "x.cpp2"
auto f() -> void { template<typename T> []() -> void{};  }

```

</details>

```output
x.cpp2:1:20: error: expected expression
auto f() -> void { template<typename T> []() -> void{};  }
                   ^
1 error generated.
```

JohelEGP:
Is the final `;` necessary? The grammar suggests it isn't, but it doesn't work otherwise.
```Cpp2
f: () = :<T> () = {};
```

filipsajdak:
@JohelEGP I am not sure if I get Your question right.

```cpp
f: () = :<T> () = {};
```

Is similar to the following code:

```cpp
f: () = 42;
```

Where `42` and `:<T> () = {}` are unnamed objects. So, I guess a semicolon is required. 

The `f` function does not return anything - is it intentional? I am asking as by default functions return `void` (if not specified otherwise).

JohelEGP:
Looking closer, both a `template` and a lambda were generated:
```C++
auto f() -> void { template<typename T> []() -> void{};  }
```

filipsajdak:
Generic lambdas with an explicit list of types are not yet supported.

But they should - to have symmetry between named and unnamed functions.

JohelEGP:
You're right. I thought it did work. Turns out I had used non-generic ones, and I tried a generic one as an _expression-statement_, which led me to believe that was the reason.

filipsajdak:
I have a PR that will enable support for them... I will run the last tests and will push it.

filipsajdak:
@JohelEGP I have pushed the PR: #349

PR #349, opened 2023-04-10
Add support for generic lambda with explicit type list

filipsajdak:
Currently cppfront process the following code:
```cpp
main: () = {
    g := :<T> () = { v := 0; };
}
```
And generates:
```cpp
auto main() -> int{
    auto g {template <typename T> []() -> void{auto v {0}; }};
}
```
After this change cppfront will generate:
```cpp
auto main() -> int{
    auto g {[]<typename T>() -> void{auto v {0}; }};
}
```

Closes: #346. All regression tests pass.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -2370,6 +2370,12 @@ public:
             //  Handle an anonymous function
             if (decl->is_function()) {
                 auto lambda_intro = build_capture_lambda_intro_for(decl->captures, n.position());
+
+                // Handle an anonymous generic function with explicit type list
+                if (decl->template_parameters) {
+                    print_to_string(&lambda_intro, *decl->template_parameters, false, true);
+                }
+
                 emit(*decl, lambda_intro);
             }
             //  Else an anonymous object as 'typeid { initializer }'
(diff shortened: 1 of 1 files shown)


JohelEGP:
Thank you. This fixes #346.

**Code** (https://godbolt.org/z/56cecbYMh):
```Cpp2
g: () = :<T> () = { };
f: () = {
  h := :<T> () = { };   // #346
  :<T> (x: T) -> _ = x; // #346
}
id_fn: () -> _ = :<T> (x: T) -> _ = x; // #346
```

<details><summary>Generated C++1.</summary>

```C++


#include "cpp2util.h"


#line 1 "x.cpp2"
auto g() -> void;
auto f() -> void;

#line 6 "x.cpp2"
[[nodiscard]] auto id_fn() -> auto;

#line 1 "x.cpp2"
auto g() -> void { []<typename T>() -> void{};  }
auto f() -> void{
  auto h {[]<typename T>() -> void{}}; // #346
  []<typename T>(T const& x) -> auto { return x;  };// #346
}
[[nodiscard]] auto id_fn() -> auto { return []<typename T>(T const& x) -> auto { return x;  };  }// #346


```

</details>

hsutter:
Sweet! 👏  Thanks @JohelEGP and @filipsajdak . I'm not a big user of explicit template parameter lists on lambdas, so I hadn't noticed yet that that case didn't code gen correctly. Good stuff, appreciated.

PR #350, opened 2023-04-10
Modify the scope_depth for variables in lambdas

### LINKED ISSUE #295: [BUG] Defining lambda that takes argument using forward passing style and calling it on function variable makes cppfront treat the variable as forwarded
In the current implementation of cppfront (fbf55ad9efcded58f550f05ede37b5b48399a7fd), the following code:

```cpp
f_inout: (inout x) -> _ = {
    x *= 2;
    return x;
}

main: () -> int = {
    x := 21;

    l1 := :(forward x) = {
        std::cout << f_inout(forward x) << std::endl;
    };

    l1(x);

    x++;
}
```
Generates:
```cpp
[[nodiscard]] auto f_inout(auto& x) -> auto{
    x *= 2;
    return x; 
}

    auto x {21}; 

    auto l1 {[](auto&& x) -> void{
        std::cout << f_inout(        CPP2_FORWARD(x)) << std::endl;
    }}; 

    std::move(l1)(x);

    ++CPP2_FORWARD(x);
}
```

And will fail to compile with the error:
```
../tests/bug_lambda_forward_iife.cpp2... ok (all Cpp2, passes safety checks)

../tests/bug_lambda_forward_iife.cpp2:15:5: error: expression is not assignable
    ++CPP2_FORWARD(x);
    ^ ~~~~~~~~~~~~~~~
1 error generated.
```

The variable x was passed to the lambda defined in the `main` function and is treated as forwarded, which is not in the scope of the `main` function.

If you replace `x++` with another call to the same lambda, it will generate the following:
```cpp
[[nodiscard]] auto f_inout(auto& x) -> auto{
    x *= 2;
    return x; 
}

    auto x {21}; 

    auto l1 {[](auto&& x) -> void{
        std::cout << f_inout(        CPP2_FORWARD(x)) << std::endl;
    }}; 

    l1(x);

    std::move(l1)(CPP2_FORWARD(x));
}
```
And will fail to compile with the error:
```
../tests/bug_lambda_forward_iife.cpp2... ok (all Cpp2, passes safety checks)

../tests/bug_lambda_forward_iife.cpp2:10:22: error: no matching function for call to 'f_inout'
        std::cout << f_inout(        CPP2_FORWARD(x)) << std::endl;
                     ^~~~~~~
../tests/bug_lambda_forward_iife.cpp2:15:18: note: in instantiation of function template specialization 'main()::(anonymous class)::operator()<int>' requested here
    std::move(l1)(CPP2_FORWARD(x));
                 ^
../tests/bug_lambda_forward_iife.cpp2:1:20: note: candidate function [with x:auto = int] not viable: expects an lvalue for 1st argument
[[nodiscard]] auto f_inout(auto& x) -> auto{
                   ^
1 error generated.
```

If you move the definition of lambda outside of `main`, it will work as expected and will generate the following code:
```cpp
[[nodiscard]] auto f_inout(auto& x) -> auto{
    x *= 2;
    return x; 
}

auto l1 {[](auto&& x) -> void{
    std::cout << f_inout(        CPP2_FORWARD(x)) << std::endl;
}}; 

    auto x {21}; 

    l1(x);

    ++x;
}
```

filipsajdak:
If a variable parent is a function without the name, the `scope_depth` is increased by `1`. Refrain from being mixed with variables from the scope where lambda was created.

After this change, the following code:
```cpp
f_inout: (inout x) -> _ = {
    x *= 2;
    return x;
}

main: () -> int = {
    x := 21;

    l1 := :(forward x) = {
        std::cout << f_inout(forward x) << std::endl;
    };

    l1(x);

    x++;
}
```
Will generate (skipping boilerplate):
```cpp
[[nodiscard]] auto f_inout(auto& x) -> auto{
    x *= 2;
    return x; 
}

    auto x {21}; 

    auto l1 {[](auto&& x) -> void{
        std::cout << f_inout(CPP2_FORWARD(x)) << std::endl;
    }}; 

    std::move(l1)(x);

    ++x;
}
```

Closes #295. All regression tests pass.

### DIFF
--- a/source/sema.h
+++ b/source/sema.h
@@ -971,7 +971,9 @@ public:
             || n.pass == passing_style::forward
             )
         {
-            symbols.emplace_back( scope_depth, declaration_sym( true, n.declaration.get(), n.declaration->name(), n.declaration->initializer.get(), &n));
+            // Handle variables in unnamed functions. For such cases scope_depth is increased by +1
+            auto depth = scope_depth + ((n.declaration->parent_is_function() && n.declaration->parent_declaration->name() == nullptr) ? 1 : 0 );
+            symbols.emplace_back( depth, declaration_sym( true, n.declaration.get(), n.declaration->name(), n.declaration->initializer.get(), &n));
         }
     }
 


hsutter:
Thanks! Hmm. I think this will do it... I'll think some more about whether we should actually be bumping the scope depth, I wonder if there are cases for more complex nested unnamed functions where they really want their own scope level.

PR #351, opened 2023-04-11
Fix unpublished comments ICE error

### LINKED ISSUE #344: [BUG] ICE: not all comments were printed
**Minimal reproducer** (https://godbolt.org/z/Y9sqcY1jG):
```Cpp2
a: type = { }
c: a = (); //
```

**Commands**:
```bash
cppfront x.cpp2
```

**Expected result**: A well-formed program.

**Actual result and error**:

```output
cppfront: $prefix/cppfront.cpp:1356: lower_to_cpp1_ret cpp2::cppfront::lower_to_cpp1(): Assertion `tokens.num_unprinted_comments() == 0 && "ICE: not all comments were printed"' failed.
Aborted (core dumped)
```

filipsajdak:
In the current implementation of cppfront (f83ca9b727e9c07f60a1933584a49a3d971936f5), the following code:
```cpp
a: type = { } // a
i : int = 42;
```
or
```cpp
a: namespace = { } // a
i : int = 42;
```
End with ICE `not all comments were printed` error.

This change removes extra new-line printed after namespace or UDT definition that solves this issue.

All regression tests pass (except for two files impacted by missing newlines - updated in this PR).

### DIFF
--- a/regression-tests/test-results/pure2-types-basics.cpp
+++ b/regression-tests/test-results/pure2-types-basics.cpp
@@ -14,7 +14,6 @@ class myclass;
 
 #line 60 "pure2-types-basics.cpp2"
 }
-
 //=== Cpp2 type definitions and function declarations ===========================
 
 
--- a/regression-tests/test-results/pure2-types-order-independence-and-nesting.cpp
+++ b/regression-tests/test-results/pure2-types-order-independence-and-nesting.cpp
@@ -26,7 +26,6 @@ template<typename T, typename U> class A;
 }
 
 }
-
 //=== Cpp2 type definitions and function declarations ===========================
 
 
--- a/regression-tests/test-results/pure2-types-order-independence-and-nesting.cpp
+++ b/regression-tests/test-results/pure2-types-order-independence-and-nesting.cpp
@@ -69,8 +69,8 @@ namespace M {
 template<typename T, typename U> class A {
     public: template<int I> class B {
         public: template<typename V, int J, typename W> static auto f(W const& w) -> void;
-public: B() = default; B(B const&) = delete; auto operator=(B const&) -> void = delete; };
-public: A() = default; A(A const&) = delete; auto operator=(A const&) -> void = delete; };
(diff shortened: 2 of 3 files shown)


JohelEGP:
Thank you. I still get it on this:

<details><summary>Still ICEs</summary>

```Cpp2
a: type = {
  b: int = 0;
}
c: a = (); // Directly initializes `b` from `0`.

// If we could make an aggregate:
// d: type = {
//   public e: int;
// }
// f: d = (0); // Directly initializes `e` from `0`.

g: type = {
  h: int;
  operator=: (out this, x: int) = h = x;
}
i: g = 0; // Initializes `x` from `0`, then `h` from `x`.

j: type = { }
k: type = {
  l: int = 0;
  this: j = ();
  m: int = 0;
  operator=: (out this) = { }
  operator=: (out this, x: int, y: int) = {
    l = x;
    m = y;
  }
}
// Initializes `QoI` from `0`, then `l` from `std::move(QoI)`.
n: k = (); // Directly initializes `m` from `0`.
// Initializes `x` from `0`, then `QoI` from `x`, then `l` from `std::move(QoI)`.
o: k = (0, 1); // Initializes `y` from `1`, then `m` from `y`.
```

</details>

filipsajdak:
@JohelEGP yes, I have noticed that. I am looking for a fix.

filipsajdak:
https://github.com/hsutter/cppfront/issues/344 is not fixed by this PR (I fixed two other issues) - more description about the original issue here: https://github.com/hsutter/cppfront/issues/344#issuecomment-1502566608

hsutter:
Thanks! BTW I made that comment check such a noisy ICE because it's super important to me to never drop anything the programmer wrote, including (or even especially?) comments. Thanks for looking into this.

hsutter:
Thanks for the bug reports and repros! These should all work now, fixed as a separate commit (see commit message), please check... and I did take the part of the suggested PRs which was to do the extra finalizing at the end of each section (not just at the very end of the file). 🙏

filipsajdak:
Checking...

filipsajdak:
I confirm that it works on my tests as well.

PR #354, opened 2023-04-11
Fix ICE when comment on last line on global object, closes #344

### LINKED ISSUE #344: [BUG] ICE: not all comments were printed

filipsajdak:
In the current implementation of cppfront (f83ca9), the following code
```cpp
o: k = (0, 1); // Initializes `y` from `1`, then `m` from `y`.
```
Cause ICE `not all comments were printed` error.

This change introduces a check if all comments were printed after phase1. If there are such comments, they are printed.

All regression tests pass. Closes #344

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -538,7 +538,7 @@ public:
     //-----------------------------------------------------------------------
     //  Finalize
     //
-    auto finalize()
+    auto finalize(phases phase = phases::phase2_func_defs)
     {
         flush_comments( {curr_pos.lineno+1, 1} );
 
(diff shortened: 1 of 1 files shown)


JohelEGP:
https://github.com/hsutter/cppfront/pull/351#issuecomment-1502498321 works now. Thank you.

PR #359, opened 2023-04-11
Correction for short function syntax, closes #356

### LINKED ISSUE #356: [BUG] Contract after simple expression function fails
It works with an extra `;`, if the expression function has a _compound-statement_, or with a statement in-between.

**Minimal reproducer** (https://godbolt.org/z/TfajzxMT1):
```Cpp2
main: () = {
  :() = 1;
  [[assert: 1]]
}
```

**Commands**:
```bash
cppfront x.cpp2
```

**Expected result**: The same as any of the first two workarounds above.

**Actual result and error**:

```output
main.cpp2(3,4): error: subscript expression [ ] must not be empty (if you were trying to name a C-style array type, use 'std::array' instead) (at '[')
```

filipsajdak:
The current implementation does not work for the following code:
```cpp
main: () = {
    :() = 1;
    [[assert: 1]]
}
```
It fails with the error:
```
error: subscript expression [ ] must not be empty (if you were trying to name a C-style array type, use 'std::array' instead) (at '[')
```

This change introduces a minor correction that moves back parsing to a semicolon (to simulate a double semicolon) for short syntax.

It is not done in the following cases:
```cpp
:() = 1;(); // immediately called lambda
f(a,b,:() = 1;); // last argument in function call
f(a,:() = 1;,c); // first or in the middle argument
```

After this change, the original issue is solved.
All regression tests pass. Closes #356

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -3015,6 +3015,17 @@ private:
                     next();
                     return {};
                 }
+                if (
+                    peek(-1) && peek(-1)->type() != lexeme::RightBrace  // it is short function syntax
+                    && curr().type() != lexeme::LeftParen               // not imediatelly called
+                    && curr().type() != lexeme::RightParen              // not as a last argument to function
+                    && curr().type() != lexeme::Comma                   // not as first or in-the-middle, function argument
+                ) {
+                    // this is a fix for a short function syntax that should have double semicolon used
+                    // (check comment in expression_statement(bool semicolon_required))
+                    // We simulate double semicolon by moving back to single semicolon.
+                    next(-1);
+                }
             }
             else {
                 error("(temporary alpha limitation) an unnamed declaration at expression scope must be a function or an object");


JohelEGP:
I can confirm that the original issue is solved with this. Thank you.

JohelEGP:
Let me add that I'm not sure of the implications of this solution. I hadn't thought of IIFE for simple function expressions.

I tested the following on branch `main`. The handling of IIFEs is inconsistent. Is that because it is a semantic exception that applies to a function scope? In particular, for `main` and the `type`, they are all in a _statement_ (as opposed to the `declaration-seq` of the top-level). See <https://godbolt.org/z/vYsqTe3Gq>.

<details><summary>Cpp2 code demonstrating the issue.</summary>

```Cpp2
a0: std::any = :() -> _ = { return 0; }();

// // error: unexpected end of source file
// b0: std::any = :() -> _ = { return 0; }(
// );

// // error: unexpected end of source file
// c0: std::any = :() -> _ = { return 0; }
// ();

d0: std::any = :() -> _ = 0;();

// // error: unexpected end of source file
// e0: std::any = :() -> _ = 0;(
// );

f0: std::any = :() -> _ = 0;
(); // mixed2.cpp2:18:2: error: expected unqualified-id

main: () = {
  a1: std::any = :() -> _ = { return 0; }();

  b1: std::any = :() -> _ = { return 0; }(
  );

  c1: std::any = :() -> _ = { return 0; }
  ();

  d1: std::any = :() -> _ = 0;();

  e1: std::any = :() -> _ = 0;(
  );

  f1: std::any = :() -> _ = 0;
  ();
}

x: type = {
  // Enabled individually, all generate:
  // ```C++
  //   private: std::any 𝘓2 {[]() -> auto;
  // ()}; public: x() = default; x(x const&) = delete; auto operator=(x const&) -> void = delete;
  // ```

  // a2: std::any = :() -> _ = { return 0; }();

  // b2: std::any = :() -> _ = { return 0; }(
  // );

  // c2: std::any = :() -> _ = { return 0; }
  // ();

  // d2: std::any = :() -> _ = 0;();

  // e2: std::any = :() -> _ = 0;(
  // );

  // f2: std::any = :() -> _ = 0;
  // ();
}
```

</details>

Should I open an issue for this?

filipsajdak:
No need for a new issue as this is not yet merged.

filipsajdak:
For the following code:
```cpp
c0: std::any = :() -> _ = { return 0; }
();
```
first line is cpp2, the second cpp1:
```
+/* 2 */ c0: std::any = :() -> _ = { return 0; }
+/* 1 */ ();
```

Nevertheless it should just inform that there is an issue with missing semicolon.

filipsajdak:
I have pushed the fix - the issue is with the `error()` method used to report errors.
The method uses `curr()` internally, which throws when cppfront has processed all tokens.
https://github.com/hsutter/cppfront/blob/f83ca9b727e9c07f60a1933584a49a3d971936f5/source/parse.h#L2760

https://github.com/hsutter/cppfront/blob/f83ca9b727e9c07f60a1933584a49a3d971936f5/source/parse.h#L2787-L2795

filipsajdak:
Now the error looks like the following:
```
bug_352_error_message_semicolon.cpp2...
bug_352_error_message_semicolon.cpp2(1,39): error: ill-formed initializer (after '}')
bug_352_error_message_semicolon.cpp2(1,1): error: unexpected text at end of Cpp2 code section (at 'c0')
bug_352_error_message_semicolon.cpp2(1,0): error: parse failed for section starting here
```

JohelEGP:
Yes, with the build from 2b730e846f0a31ba153e3d6c545493f024197655.

hsutter:
Looks good, thanks!

PR #360, opened 2023-04-11
Add diagnostics for type alias to a wildcard, closes #357

### LINKED ISSUE #357: [BUG] alias: type == _ generates invalid cpp1
**Minimal reproduce**
`alias5: type == _;`
passes succssfully cpp2, but generates invalid cpp1:
`using alias5 = auto;`

**Command lines including which C++ compiler you are using**
`cppfront example.cpp2 -p`
(build from latest mai. https://github.com/hsutter/cppfront/commit/f83ca9b727e9c07f60a1933584a49a3d971936f5 with clang++-15)
Then cpp1 compiled with
`clang++-15 -Icppfront/include $1.cpp -std=c++20 -o $1`

**Actual result/error**
Cppfront passes with
`example.cpp2... ok (all Cpp2, passes safety checks)`
But then clang fails with
`error: 'auto' not allowed in type alias using alias5 = auto;`

**Expected result**
Diagnostic at cpp2 side. (I guess that`s strategy, to catch whatever possible cpp2 relateed before relying on cpp1 compiler?)

**Additional context**
This is extreme case, because clang says `auto` is never allowed. 
There is other case
`alias6: type == foo` when `foo` is not defined generates correct
`using alias6 = foo;`
But I guess this can be separate issue, for should cppfront check if alias goes to something undefined.

filipsajdak:
In the current implementation of cppfront (f83ca9), the following code:
```cpp
alias5: type == _;
alias6: type == auto;
```
Generates succesfuly:
```cpp
using alias5 = auto;
using alias6 = auto;
```
Which is an invalid cpp1 code.

After this change, the alias to wildcard will generate the following error:
```
error: a 'type ==' alias declaration must be followed by a type name (not a wildcard _ nor auto)
```

All regression tests pass. Closes #357

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -5617,6 +5617,16 @@ private:
                 );
                 return {};
             }
+            if (
+                t->is_wildcard() 
+                || ( t->get_token() && t->get_token()->to_string(true) == "auto" )
+            ) {
+                errors.emplace_back(
+                    curr().position(),
+                    "a 'type ==' alias declaration must be followed by a type name (not a wildcard _ nor auto)"
+                );
+                return {};
+            }
             a->initializer = std::move(t);
         }
 


hsutter:
Thanks!

PR #362, opened 2023-04-12
Add diagnostics for bad semicolons errors, closes #352

### LINKED ISSUE #352: [QoI] Improve cryptic error message due to accidental semicolon
The source locations are nowhere near the semicolon. The implementation should know when a semicolon is not needed, so the error message can be more to the point.

**Minimal reproducer** (https://godbolt.org/z/5Eovbx4Yb):
```Cpp2
g: () = { }
f: () = {
  for v* | all next log() do :(e) = {
    foo();
    bar*;
  };
  g();
}
```

**Commands**:
```bash
cppfront x.cpp2
```

**Expected result**: A diagnostic indicating to remove accidental semicolon.

**Actual result and error**:

```output
x.cpp2(2,9): error: ill-formed initializer (at '{')
x.cpp2(2,1): error: unexpected text at end of Cpp2 code section (at 'f')
x.cpp2(1,0): error: parse failed for section starting here
```

filipsajdak:
The current cppfront (f83ca9) implementation provides unclear error messages - it is hard to find what causes the error.

This change introduces diagnostics for:
```cpp
f: () = {}; // error: a compound function body shall not end with a semicolon (at ';')
g: () -> int = 42;; // error: a short function syntax shall end with a single semicolon (at ';')

for v* | all next log() do :(e) = {
  foo();
  bar*;
}; // error: for..do loop body shall not end with a semicolon (at ';')

i := 0;
while i* < container.size() next i++ {
  std::cout << container[i] << std::endl;
}; // error: while loop body shall not end with a semicolon (at ';')

l2 := :() -> int = 24;; // error: a short function syntax shall end with a single semicolon (at ';')

; // error: empty statement is not allowed - remove extra semicolon (at ';')

i := 0;; // error: empty statement is not allowed - remove extra semicolon (at ';')
```

All regression tests pass. Closes #352

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -4485,6 +4485,11 @@ private:
                 n->for_with_in = true;
             }
 
+            if (!done() && curr().type() == lexeme::Semicolon) {
+                error("a loop body may not be followed by a semicolon (empty statements are not allowed)");
+                return {};
+            }
+
             return n;
         }
 
--- a/source/parse.h
+++ b/source/parse.h
@@ -5690,6 +5690,21 @@ private:
             return {};
         }
 
+        if (
+            n->is_function()
+            && n->initializer
+            && !done() && curr().type() == lexeme::Semicolon
+            )
+        {
+            if (n->initializer->is_compound() && n->has_name()) {
+                error("a braced function body may not be followed by a semicolon (empty statements are not allowed)");
+                return {};
+            } else if (n->initializer->is_expression()) {
+                error("a single-expression function should end with a single semicolon");
+                return {};
+            }
+        }
+
         //  If this is an implicit constructor, it must have two parameters
         if (n->is_constructor())
         {
--- a/source/parse.h
+++ b/source/parse.h
@@ -4722,6 +4722,11 @@ private:
     )
         -> std::unique_ptr<statement_node>
     {
+        if (!done() && curr().type() == lexeme::Semicolon) {
+            error("empty statement is not allowed - remove extra semicolon");
+            return {};
+        }
+
         auto n = std::make_unique<statement_node>();
 
         //  Now handle the rest of the statement
--- a/source/parse.h
+++ b/source/parse.h
@@ -4423,6 +4423,10 @@ private:
             if (!handle_logical_expression  ()) { return {}; }
             if (!handle_optional_next_clause()) { return {}; }
             if (!handle_compound_statement  ()) { return {}; }
+            if (!done() && curr().type() == lexeme::Semicolon) {
+                error("a loop body may not be followed by a semicolon (empty statements are not allowed)");
+                return {};
+            }
             return n;
         }
 
--- a/source/parse.h
+++ b/source/parse.h
@@ -3068,15 +3068,15 @@ private:
         -> void
     {
         auto m = std::string{msg};
+        auto i = done() ? -1 : 0;
+        assert (peek(i));
         if (include_curr_token) {
-            m += std::string(" (at '") + curr().to_string(true) + "')";
+            m += std::string(" (at '") + peek(i)->to_string(true) + "')";
         }
         if (
             err_pos == source_position{}
-            && peek(0)
-            )
-        {
-            err_pos = peek(0)->position();
+        ) {
+            err_pos = peek(i)->position();
         }
         errors.emplace_back( err_pos, m, false, fallback );
     }
@@ -5093,7 +5093,8 @@ private:
 
         //  If there's no [ [ then this isn't a contract
         if (
-            curr().type() != lexeme::LeftBracket
+            done()
+            || curr().type() != lexeme::LeftBracket
             || !peek(1)
             || peek(1)->type() != lexeme::LeftBracket
             )
--- a/source/parse.h
+++ b/source/parse.h
@@ -3225,6 +3225,12 @@ private:
                     next();
                     return {};
                 }
+                if ( // check if a single-expression function is followed by an extra second semicolon
+                    decl->initializer && decl->initializer->is_expression()
+                    && !done() && curr().type() == lexeme::Semicolon
+                ) {
+                    error("a single-expression function should end with a single semicolon");
+                }
                 if (!(*func)->contracts.empty()) {
                     error("an unnamed function at expression scope currently cannot have contracts");
                     next();


filipsajdak:
I have removed `error: bad statement!` - it is even better now (no impact on regression tests).

source/parse.h
```diff
@@ -3010,6 +3025,12 @@ class parser
                     next();
                     return {};
                 }
+                if ( // check if unnamed function ends with semicolon
```
hsutter, line 3028: Thanks! I'm mainly going to suggest comment/message rewordings please...

"check if a single-expression function is followed by an extra second semicolon"

source/parse.h
```diff
@@ -3010,6 +3025,12 @@ class parser
...
+                    decl->initializer && decl->initializer->is_expression()
+                    && !done() && curr().type() == lexeme::Semicolon
+                ) {
+                    error("an unnamed function shall end with a single semicolon");
```
hsutter, line 3032: "a single-expression function should end with a single semicolon"

source/parse.h
```diff
@@ -4197,6 +4218,10 @@ class parser
             if (!handle_logical_expression  ()) { return {}; }
             if (!handle_optional_next_clause()) { return {}; }
             if (!handle_compound_statement  ()) { return {}; }
+            if (!done() && curr().type() == lexeme::Semicolon) {
+                error("while loop body shall not end with a semicolon");
```
hsutter, line 4222: "a loop body may not be followed by a semicolon (empty statements are not allowed)"

source/parse.h
```diff
@@ -4259,6 +4284,11 @@ class parser
                 n->for_with_in = true;
             }
 
+            if (!done() && curr().type() == lexeme::Semicolon) {
+                error("for..do loop body shall not end with a semicolon");
+                return {};
```
hsutter, line 4289: "a loop body may not be followed by a semicolon (empty statements are not allowed)"

source/parse.h
```diff
@@ -5446,6 +5482,21 @@ class parser
...
         }
 
+        if (
+            n->is_function()
+            && n->initializer
+            && !done() && curr().type() == lexeme::Semicolon
+            )
+        {
+            if (n->initializer->is_compound() && n->has_name()) {
+                error("a compound function body shall not end with a semicolon");
```
hsutter, line 5492: "a braced function body may not be followed by a semicolon (empty statements are not allowed)"

source/parse.h
```diff
@@ -5446,6 +5482,21 @@ class parser
...
+                return {};
+            } else if (n->initializer->is_expression()) {
+                error("a short function syntax shall end with a single semicolon");
+                return {};
```
hsutter, line 5496: "a single-expression function should end with a single semicolon"

hsutter (changes requested):

source/parse.h
filipsajdak: working on it...

filipsajdak:
@hsutter all requested changes applied.

source/parse.h
```diff
@@ -3071,16 +3071,31 @@ class parser
...
-        if (
-            err_pos == source_position{}
-            && peek(0)
-            )
-        {
-            err_pos = peek(0)->position();
+
+        if (done()) {
+            int i = 0;
+            while(!peek(i) && pos + i > 0) { --i; };
```
hsutter, line 3077: I see this is making `error` more robust if we're `done()`, but I'm not sure what this loop is doing... is it assuming we might be more than one past the end? But that shouldn't happen -- see the `assert (pos <= std::ssize(*tokens));` in `done()`.

Instead of this diff, what if we changed this original code

```
        if (include_curr_token) {
            m += std::string(" (at '") + curr().to_string(true) + "')";
        }
        if (
            err_pos == source_position{}
            && peek(0)
            )
        {
            err_pos = peek(0)->position();
```

to this?

```
        auto i = done() ? -1 : 0;
        assert (pos+i < std::ssize(*tokens));
        if (include_curr_token) {
            m += std::string(" (at '") + peek(i)->to_string(true) + "')";
        }
        if (
            err_pos == source_position{}
            && peek(i)
            )
        {
            err_pos = peek(i)->position();
```

Does that accomplish what you were aiming for here?

filipsajdak: Your proposed code was the first I made. Later, I wondered if I could assume that going back by one would be enough. That is why I have introduced the loop to ensure to go back to the last good one. If you think that going back by one is enough, we shall add an assertion that will emphasize this assumption.

filipsajdak: I have the following assertion in mind:
```cpp
auto i = done() ? -1 : 0;
assert (peek(i)); // <---- this one
assert (pos+i < std::ssize(*tokens));
```

hsutter: Yes, that looks good.

filipsajdak: OK, I am correcting it.

hsutter: In fact asserting `peek(i)` is better, and you can remove the second assertion which is redundant (the peek checks it already, plus it checks `>0` so it's strictly better).

filipsajdak: As peek implementation is the following:
```cpp
    auto peek(int num) const
        -> token const*
    {
        assert (tokens);
        if (
            pos + num >= 0
            && pos + num < std::ssize(*tokens)
            )
        {
            return &(*tokens)[pos + num];
        }
        return {};
    }
```

It makes `assert (pos+i < std::ssize(*tokens));` redundant as it is already checked in `peek()`.

filipsajdak: Pushed corrected code.

source/parse.h
```diff
@@ -4707,6 +4722,11 @@ class parser
     )
         -> std::unique_ptr<statement_node>
     {
+        if (!done() && curr().type() == lexeme::Semicolon) {
+            error("empty statement is not allowed - remove extra semicolon");
```
hsutter, line 4726: I think this works, but just a note: Emitting an error here (not just returning null) could interfere with productions where *statement* is one alternative but where if it doesn't parse as a statement we would want to try other fallback production(s) instead. I don't think that actually happens today in the grammar though, so this should be okay.

hsutter: Basically, the scenario would be "is there any production where a statement could appear, where if it's not a statement something else that starts with a `;` could appear instead" -- I think there isn't today

PR #365, opened 2023-04-12
Add diagnostics for unused literal or identifier, closes #364

### LINKED ISSUE #364: [SUGGESTION] Don't allow unused value inside function body
I understand that it may be difficult to implement on cppfront side.

Consider the following, it's easy to write by mistake
`foo: ()->int = 42;`
`bar: ()->int = { 13; } //value discarded; non-void function returns nothing`

`main: () -> int = {`
`  std::cout << foo() << bar();`
`}`

In that case of course cpp1 will  give a warning about `bar` not returning value, but if it's not enforced that would be UB and crash. 
But may be it is possible to track in cppfront if function has return value, return paths etc.

filipsajdak:
In a compound statement, when there is a literal or identifier followed by a semicolon, it is treated as unused.
For the following code:
```cpp
    x;
```
The following error is reported:
```
error: unused literal or identifier (at 'x')
```

All tests pass (one test that has unused literals was updated). Closes #364


regression-tests/mixed-float-literals.cpp2
```diff
@@ -1,94 +1,94 @@
...
-  // 1e - 10.0f;
+  // (void)1e-10.0f;;
+  // 1.0e-10.0;;
+  // 1.0e-10.0f;;
+  // 1e-10.0;;
+  // 1e-10.0f;;
+  // 1e - 10;;
+  // 1e - 10.0;;
+  // 1e - 10f;;
+  // 1e - 10.0f;;
```
JohelEGP, line 54: Seems like you accidentally added `;`s to these.

filipsajdak: right... fixing.

filipsajdak: fixed.

hsutter:
Looks good, thanks!

PR #366, opened 2023-04-13
Add support for multiple `else if` and `if constexpr`

filipsajdak:
Add support for else if branches and constexpr ifs

* Added support for else if branches.
* Added support for constexpr ifs
* Added support for definite initialization checks
* Added support for debug symbols (if branch, if else branch, else branch)
* Add support for constructing variables in conditions
* Add support for nested ifs/else-ifs

The current implementation of cppfront (f83ca9) supports only `if` with one condition:
```cpp
if x > 1 {
  // true branch
} else {
  // false branch
}
```
 
This change makes it possible to construct ifs with multiple `else if` checks. That makes the following code:
```cpp
main: (args) = {
    p : *int;

    a := 1;
    b := 2;
    c := 3;
    d := 4;

    if args.cout == 3 {
        p = a&;
    } else if p = b& {
        if args.cout == 2 {
            p = c&;
        } else {
            if b > 0 {
                p = a&;
            } 
            else {
                p = d&;
            }
        }
    } else {
        p = c&;
    }

    std::cout << p* << std::endl;
}
```
Generates:
```cpp
auto main(int const argc_, char const* const* const argv_) -> int{
    auto args = cpp2::make_args(argc_, argv_); 
#line 2 "/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2"
    cpp2::deferred_init<int*> p; 

    auto a {1}; 
    auto b {2}; 
    auto c {3}; 
    auto d {4}; 

    if (args.cout==3) {
        p.construct(&a);
    } else if (p.construct(&b)) {
        if (args.cout==2) {
            p.construct(&c);
        }else {
            if (cpp2::cmp_greater(std::move(b),0)) {
                p.construct(&a);
            }
            else {
                p.construct(&d);
            }
        }
    }else {
        p.construct(&c);
    }

    std::cout << *cpp2::assert_not_null(std::move(p.value())) << std::endl;
}
```

The change also brings support for debug symbols:
```
  0 | function main
  1 |   scope
  2 |     var p *** UNINITIALIZED
  2 |     /var 
  2 |     var a
  2 |     /var 
  2 |     var b
  2 |     /var 
  2 |     var c
  2 |     /var 
  2 |     var d
  2 |     /var 
  2 |     selection
  3 |       if branch
  4 |         *** (10,9) DEFINITE INITIALIZATION OF p
  4 |         *** use of a
  4 |         /if branch
  3 |       *** (11,15) DEFINITE INITIALIZATION OF p
  3 |       *** use of b
  3 |       if else branch
  4 |         selection
  5 |           if branch
  6 |             *** (13,13) DEFINITE INITIALIZATION OF p
  6 |             *** (13,17) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of c
  6 |             /if branch
  5 |           else branch
  6 |             selection
  7 |               *** (15,16) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of b
  7 |               if branch
  8 |                 *** (16,17) DEFINITE INITIALIZATION OF p
  8 |                 *** (16,21) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of a
  8 |                 /if branch
  7 |               else branch
  8 |                 *** (19,17) DEFINITE INITIALIZATION OF p
  8 |                 *** (19,21) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of d
  8 |                 /else branch
  7 |               /selection
  6 |             /else branch
  5 |           /selection
  4 |         /if else branch
  3 |       else branch
  4 |         *** (23,9) DEFINITE INITIALIZATION OF p
  4 |         *** (23,13) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of c
  4 |         /else branch
  3 |       /selection
  2 |     *** (26,18) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of p
  2 |     /scope
  1 |   /function
```

All regression tests pass.


JohelEGP:
> Add support for constructing variables in conditions

Looking at <https://godbolt.org/z/vT8rz8brb>, which fails with
```output
main.cpp2:3:7: error: value of type 'void' is not contextually convertible to 'bool'
  if (b.construct(true)) {
      ^~~~~~~~~~~~~~~~~
1 error generated.
```
Wouldn't making that return an lvalue reference to the constructed object, simulating the conventional semantics of the assignment operator, have sufficed?

filipsajdak:
Thanks for the feedback! I did not check if that makes sense. This is how it works before my change - I just make it work in multiple `else if` conditions.

filipsajdak:
You are right - returning the lvalue reference will solve that issue.

JohelEGP:
> I asked myself what would @hsutter say. I think he might ask if is it a bug or is it a feature.

I do wonder. The meaning of the first `x = y` changes depending on whether `x` was uninitialized or not.

JohelEGP:
> Currently, it will save us from the mistakes like writing `i = 42` instead of `i == 42`.

That can happen if it's not a definite first assignment.

Opened #368 for the suggestion at https://github.com/hsutter/cppfront/pull/366#issuecomment-1506153367.

source/sema.h
```diff
@@ -695,6 +698,7 @@ class sema
             }
         };
         std::vector<stack_entry> selection_stack;
+        bool                     in_branch = false;
```
hsutter, line 701: Since this is a single value rather than a stack, just checking: Does the code handle nested loops correctly?

hsutter (commented):
Thanks! Definite initialization currently requires that an uninitialized variable be initialized on both branches or neither branch... did you test that this generalizes correctly to requiring that an uninitialized variable be initialized on all branches or no branch?

filipsajdak:
Yes. I have test that. That was the easy part. The thing I was strangling was the case where the initialization happened in the condition of some else-if (previously there was only one condition and it was handled as an exception).

I might add some tests to this PR as a check.

filipsajdak:
I am not sure about not "no branch case" - I will check it.

source/sema.h
filipsajdak: I have tested the previous implementation of if-else:

```cpp
    if args.cout == 3 {
        p = a&;
    } else {
        if args.cout == 2 {
            p = c&;
        } else {
            if p = b& {
                p = a&;
            } 
            else {
                p = d&;
            }
        }
    }
```
Generates:
```cpp
    if (args.cout==3) {
        p.construct(&a);
    }else {
        if (args.cout==2) {
            p.construct(&c);
        }else {
            if (p.value() = &b) { // this is not good
                p.value() = &a; // this is not good
            }
            else {
                p.value() = &d; // this is not good
            }
        }
    }
```

And:
```cpp
    if args.cout == 3 {
        while d > 0 next d-- {
            p = a&;
        }
    } else {
        for args do :(e) = {
            p = c&;
        }
    }
```
Generates:
```cpp
    if (args.cout==3) {
        for( ; cpp2::cmp_greater(d,0); --d ) {
            p.construct(&a); // bad due to Default.expects(!init);
        }
    }else {
        for ( auto const& cpp2_range = args;  auto const& e : cpp2_range ) {
            p.construct(&c); // bad due to Default.expects(!init);
        }
    }
```

filipsajdak: My current implementation behaves in the same way but handles multiple else-ifs.

filipsajdak: I will work on solving these issues.

filipsajdak:
I have fixed nested ifs/else-ifs. The new version of the code was pushed.

Now for the following code:
```cpp
main: (args) = {
    p : *int;

    a := 1;
    b := 2;
    c := 3;
    d := 4;

    if args.cout == 3 {
        p = a&;
    } else if p = b& {
        if args.cout == 2 {
            p = c&;
        } else {
            if b > 0 {
                p = a&;
            } 
            else {
                p = d&;
            }
        }
    } else {
        p = c&;
    }

    std::cout << p* << std::endl;
}
```
It generates:
```cpp


#include "cpp2util.h"



#line 1 "/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2"
auto main(int const argc_, char const* const* const argv_) -> int;
    

#line 1 "/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2"
auto main(int const argc_, char const* const* const argv_) -> int{
    auto args = cpp2::make_args(argc_, argv_); 
#line 2 "/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2"
    cpp2::deferred_init<int*> p; 

    auto a {1}; 
    auto b {2}; 
    auto c {3}; 
    auto d {4}; 

    if (args.cout==3) {
        p.construct(&a);
    } else if (p.construct(&b)) {
        if (args.cout==2) {
            p.construct(&c);
        }else {
            if (cpp2::cmp_greater(std::move(b),0)) {
                p.construct(&a);
            }
            else {
                p.construct(&d);
            }
        }
    }else {
        p.construct(&c);
    }

    std::cout << *cpp2::assert_not_null(std::move(p.value())) << std::endl;
}
```
And creates the following symbols:
```
  0 | function main
  1 |   scope
  2 |     var p *** UNINITIALIZED
  2 |     /var 
  2 |     var a
  2 |     /var 
  2 |     var b
  2 |     /var 
  2 |     var c
  2 |     /var 
  2 |     var d
  2 |     /var 
  2 |     selection
  3 |       if branch
  4 |         *** (10,9) DEFINITE INITIALIZATION OF p
  4 |         *** use of a
  4 |         /if branch
  3 |       *** (11,15) DEFINITE INITIALIZATION OF p
  3 |       *** use of b
  3 |       if else branch
  4 |         selection
  5 |           if branch
  6 |             *** (13,13) DEFINITE INITIALIZATION OF p
  6 |             *** (13,17) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of c
  6 |             /if branch
  5 |           else branch
  6 |             selection
  7 |               *** (15,16) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of b
  7 |               if branch
  8 |                 *** (16,17) DEFINITE INITIALIZATION OF p
  8 |                 *** (16,21) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of a
  8 |                 /if branch
  7 |               else branch
  8 |                 *** (19,17) DEFINITE INITIALIZATION OF p
  8 |                 *** (19,21) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of d
  8 |                 /else branch
  7 |               /selection
  6 |             /else branch
  5 |           /selection
  4 |         /if else branch
  3 |       else branch
  4 |         *** (23,9) DEFINITE INITIALIZATION OF p
  4 |         *** (23,13) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of c
  4 |         /else branch
  3 |       /selection
  2 |     *** (26,18) DEFINITE LAST POTENTIALLY MOVING USE OF *** use of p
  2 |     /scope
  1 |   /function
```

filipsajdak:
If one of the branches is not initializing the variable, it stops with the error on the if statements that do not provide initialization in all branches. E.g.:

```cpp
main: (args) = {
    p : *int;

    a := 1;
    b := 2;
    c := 3;
    d := 4;

    if args.cout == 3 {
        p = a&;
    } else if p = b& {
        if args.cout == 2 {
            p = c&;
        } else {
            if b > 0 {
                p = a&;
            } 
            else {
                // p = d&;
            }
        }
    } else {
        p = c&;
    }

    std::cout << p* << std::endl;
}
```
Gives the following error:
```
/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2...
pure2-nested-ifs.cpp2(2,5): error: local variable p must be initialized on both branches or neither branch
pure2-nested-ifs.cpp2(15,13): error: "if" initializes p on:
  branch starting at line 15
but not on:
  branch starting at line 18
```

And the following: 
```cpp
main: (args) = {
    p : *int;

    a := 1;
    b := 2;
    c := 3;
    d := 4;

    if args.cout == 3 {
        p = a&;
    } else if p = b& {
        if args.cout == 2 {
            p = c&;
        } else {
            if b > 0 {
                p = a&;
            } 
            else {
                p = d&;
            }
        }
    } else {
        // p = c&;
    }

    std::cout << p* << std::endl;
}
```
Gives:
```
/Users/filipsajdak/dev/execspec/external/tests/pure2-nested-ifs.cpp2...
pure2-nested-ifs.cpp2(2,5): error: local variable p must be initialized on both branches or neither branch
pure2-nested-ifs.cpp2(9,5): error: "if" initializes p on:
  branch starting at line 9
  branch starting at line 11
but not on:
  branch starting at line 22
```

JohelEGP:
> ```c++
>     } else if (p.construct(&b)) {
>         if (args.cout==2) {
>             p.construct(&c);
>         }else {
>             if (cpp2::cmp_greater(std::move(b),0)) {
>                 p.construct(&a);
>             }
>             else {
>                 p.construct(&d);
>             }
>         }
>     }else {
> ```
> -- https://github.com/hsutter/cppfront/pull/366#issuecomment-1516916735

This looks wrong. The `if` is entered if the condition is `true`, which happens when `p` is constructed. So the other `contruct` should be the `value` accessor to acces the constructed value.

filipsajdak:
> Thanks! Definite initialization currently requires that an uninitialized variable be initialized on both branches or neither branch... did you test that this generalizes correctly to requiring that an uninitialized variable be initialized on all branches or no branch?

@hsutter I can now confirm that this generalization correctly requires that an uninitialized variable be initialized on all branches or on no branch.

filipsajdak:
OK, tests added. PR ready

filipsajdak:
I need to fix a few cases... it is not yet ready.

hsutter:
Two comments after thinking about this more:

Re `if constexpr`: Yes, that was not getting emitted. I'll fix this on its own in a separate commit to get it in there. Thanks!

Re `else if`: I think everything already works correctly, including for guaranteed initialization, except only that `else if` has to be spelled `else { if` (and then another `}` at the end of that block)... correct? If so, I wonder if a less invasive approach would be just to drop the `{` `}` requirement if `else` is followed by `if`. That would happen during parsing, and wouldn't require any changes in `sema.h` I think?

filipsajdak:
@hsutter I have completed the change. Please check the tests that are included.

I have also covered "support for constructing variables in conditions" even though `p.construct(&c)` cannot be used as a condition as it returns `void` - cpp1 code generation is fine.

filipsajdak:
I see that I was late ~15min - nice sync. I was doing the last checks and tests. I was thinking about spelling `else if` as `else { if`. 

I drop that idea for some reason. I will think why as I don't recall it now.

hsutter:
Thanks again for this suggestion and all the work on this.

I'm sorry I have to be the bearer of sad news: I've decided to try out the simpler suggestion, and coded it up just now. It seems to work well so I'm going to go ahead and check it in. But your work on this has not gone to waste, you raised the issue that we needed `else if` to work (and `/*else*/ if constexpr` 😁), you explored and showed one way to do it, and you generated the test cases above that I'm using and will check in for this. Thanks!

Also, your example made me remember that I had forgotten to disallow `=` initialization inside an `if` conditional expression. It's already correctly disallowed in other conditional expressions, such as `while`, because those used `logical_expression` instead of `expression`. The `if` production now does that too.

PR #369, opened 2023-04-14
Add class template argument deduction (CTAD) for template user defined types (UDT)

filipsajdak:
Currently, after compiling the following code:
```cpp
value: <T: type, Y: type, Z: type> type = {
    public data1: T;
    public data2: Y;
    public data3: Z;

    operator=: (out this, t: T, inout y: Y, z: Z) = {
        data1 = t;
        data2 = y;
        data3 = z;
    }
    operator=: (out this, t: * const T, y: *Y, move z: Z) = {
        data1 = t*;
        data2 = y*;
        data3 = z;
    }
}

main: () = {
    s := std::string("lvalue string");
    v: value = (42, s, :std::string = "temporary string");
    std::cout << "(v.data1)$, (v.data2)$, (v.data3)$" << std::endl;

    s = "will be moved";
    w: value = (v.data1&, v.data2&, move s);
    std::cout << "(w.data1)$, (w.data2)$, (w.data3)$" << std::endl;
}
```
We will get the following error:
```
../tests/ctad.cpp2... ok (all Cpp2, passes safety checks)

../tests/ctad.cpp2:20:11: error: no viable constructor or deduction guide for deduction of template arguments of 'value'
    value v {42, s, std::string{"temporary string"}};
          ^
../tests/ctad.cpp2:6:13: note: candidate template ignored: couldn't infer template argument 'T'
    public: value(cpp2::in<T> t, Y& y, cpp2::in<Z> z);
            ^
../tests/ctad.cpp2:11:13: note: candidate template ignored: couldn't infer template argument 'T'
    public: value(cpp2::in<T const*> t, cpp2::in<Y*> y, Z&& z);
            ^
../tests/ctad.cpp2:1:52: note: candidate function template not viable: requires 1 argument, but 3 were provided
template<typename T, typename Y, typename Z> class value {
                                                   ^
1 error generated.
```

After this change cppfront generates Class template argument deduction (CTAD) for template class that has constructors with the same argument type list as template arguments (order matters). cppfront will add the following cpp1 code to `Cpp2 type definitions and function declarations` section:
```cpp
template<typename T, typename Y, typename Z> value(T const &, Y&, Z const &) -> value<T, Y, Z>;
template<typename T, typename Y, typename Z> value(T const* const &, Y* const &, Z&&) -> value<T, Y, Z>;
```
And will make the code compile and run successfuly:
```
42, lvalue string, temporary string
42, lvalue string, will be moved
```

All regression tests pass.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -5027,6 +5027,13 @@ public:
             {
                 printer.print_cpp2( " final", n.position() );
             }
+            
+            // CTAD
+            struct ctad_details {
+                std::string args;
+                std::string types;
+            };
+            auto ctads             = std::vector<ctad_details>{};
 
             //  Type definition
 
@@ -5052,6 +5059,52 @@ public:
 
                 if (decl->is_constructor()) {
                     found_constructor = true;
+
+                    if (n.template_parameters) {
+                        if (auto func = std::get_if<declaration_node::a_function>(&decl->type)) {
+                            const auto& params = (*func)->parameters->parameters;
+                            auto can_generate_ctad = std::equal(
+                                std::cbegin(params)+1, std::cend(params),
+                                std::cbegin(n.template_parameters->parameters), std::cend(n.template_parameters->parameters),
+                                [](const auto& arg, const auto& type) -> bool {
+                                    if (const auto* arg_type = std::get_if<declaration_node::an_object>(&arg->declaration->type)) {
+                                        assert(type->name());
+                                        return (*arg_type)->get_token() && (*arg_type)->get_token()->as_string_view() == type->name()->as_string_view();
+                                    } else {
+                                        return false;
+                                    }
+                                }
+                            );
+                            if (can_generate_ctad) {
+                                ctads.emplace_back();
+                                std::for_each(std::cbegin(params)+1, std::cend(params), [&](const auto& arg) {
+                                    if (const auto* type = std::get_if<declaration_node::an_object>(&arg->declaration->type)) {
+                                        auto& arg_t = *type;
+                                        assert(arg_t->get_token());
+                                        auto type_name = arg_t->get_token()->to_string(true);
+                                        auto arg_type = print_to_string(*arg_t);
+                                        switch (arg->pass) {
+                                            case passing_style::in     : arg_type += " const &"; break;
+                                            case passing_style::out    : arg_type = "cpp2::deferred_init<" + arg_type + ">*"; break;
+                                            case passing_style::inout  : arg_type += "&"; break;
+                                            case passing_style::move   :
+                                            case passing_style::forward: arg_type += "&&"; break;
+                                            case passing_style::copy   : break;
+                                            default: assert(!"ICE: invalid argument passing style");
+                                        }
+
+                                        if (ctads.back().args.empty()) {
+                                            ctads.back().args = arg_type;
+                                            ctads.back().types = type_name;
+                                        } else {
+                                            ctads.back().args += ", " + arg_type;
+                                            ctads.back().types += ", " + type_name;
+                                        }
+                                    }
+                                });
+                            }
+                        }
+                    }
                 }
                 if (decl->is_constructor_with_that()) {
                     found_that_constructor = true;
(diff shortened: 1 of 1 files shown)


source/cppfront.cpp
```diff
@@ -4919,6 +4925,52 @@ class cppfront
...
+                                std::for_each(std::cbegin(params)+1, std::cend(params), [&](const auto& arg) {
+                                    if (const auto* type = std::get_if<declaration_node::an_object>(&arg->declaration->type)) {
+                                        auto& arg_t = *type;
+                                        assert(arg_t->get_token());
+                                        auto type_name = arg_t->get_token()->to_string(true);
+                                        auto arg_type = print_to_string(*arg_t);
+                                        switch (arg->pass) {
+                                            case passing_style::in     : arg_type += " const &"; break;
+                                            case passing_style::out    :
+                                            case passing_style::inout  : arg_type += "&"; break;
```
JohelEGP, line 4955: See <https://godbolt.org/z/1dcWs4za1>.
```suggestion
                                            case passing_style::out    : arg_type.insert(0, "cpp2::deferred_init<") += ">*"; break;
                                            case passing_style::inout  : arg_type += "&"; break;
```

JohelEGP (changes requested):

source/cppfront.cpp
filipsajdak: Oh, thank you. It seems there is only a need to avoid `cpp2::in<T>` as it is the only not deducible type. So, I guess I can make it simpler to reuse the cpp2 way of passing except for passing `in` args.

filipsajdak: One comment about `data1 = (t = 42);` in the following code:
```cpp
 operator=: (out this, out t: T, inout y: Y, z: Z) = {
        data1 = (t = 42);
        data2 = y;
        data3 = z;
    }
```
I think we shall not support that kind of style of coding - there is very good alternative that works and is more clear about the intentions:
```cpp
    operator=: (out this, out t: T, inout y: Y, z: Z) = {
        t = 42;
        data1 = t;
        data2 = y;
        data3 = z;
    }
```

filipsajdak: Fixed.

hsutter:
Thanks! As I look at the examples, it strikes me that this should be (but isn't) falling into passing by `T const&` (not `in<T>`) when `T` is a template parameter type, which I did specifically for deduction.

It seems we should bypass `in<T>` in two more places.
1. For a constructor parameter when `T` is a template parameter on the class, pass `T const&` for CTAD deduction. We do that already for all functions when `T` is a template parameter on the function, for general deduction.
2. For a parameter `* <anything>`, pass `<anything> *` by value (not `in<T>`). We know it's a pointer, and should be passed by value.

If we did those two things:
- would that address these examples, as well as being generally useful?
- would that replace the need for this PR?

source/cppfront.cpp
JohelEGP: Did that work before? Now it errors with:
```output
build/_cppfront/main.cpp:30:18: error: cannot initialize a member subobject of type 'int' with an rvalue of type 'void'
        : data1{ t.construct(42) }
                 ^~~~~~~~~~~~~~~
```

JohelEGP:
That's indeed a general improvement.
I tested it by hijacking the `cpp2` identifier: <https://cpp2.godbolt.org/z/WaM6998T3>.

Unfortunately, `out` parameters also inhibit deduction: <https://cpp2.godbolt.org/z/j54j65j9q>.

JohelEGP:
A way to support the deduction of an `out` parameter
would be by changing the pass-by `out` syntax `out i`
from being lowered to an argument intended for a `cpp2::out` parameter
to being lowered to a value of `cpp2::out`: <https://cpp2.godbolt.org/z/34d5EfazK>.

hsutter:
> A way to support the deduction of an `out` parameter
> would be by changing the pass-by `out` syntax `out i`
> from being lowered to an argument intended for a `cpp2::out` parameter
> to being lowered to a value of `cpp2::out`: https://cpp2.godbolt.org/z/34d5EfazK.

I'm happy to try that... it creates an extra copy of the `out` but that's enough for an experiment and can be optimized later.

Pushed: e0125d7c0fdf1a2e166493fd2a7d6c88ab10ff7f

JohelEGP:
> it creates an extra copy

Please, see <https://github.com/hsutter/cppfront/pull/198#issuecomment-1483685807>.
You can initialize a parameter from a prvalue of a non-copyable non-movable type: <https://compiler-explorer.com/z/Y1x5nMT16>.
So there should be no extra copy.

hsutter:
OK, done... today's two commits (mentioned above) now apply all three changes discussed in this thread... quoting for convenience:

> 1. For a constructor parameter when `T` is a template parameter on the class, pass `T const&` for CTAD deduction. We do that already for all functions when `T` is a template parameter on the function, for general deduction.
> 2. For a parameter `* <anything>`, pass `<anything> *` by value (not `in<T>`). We know it's a pointer, and should be passed by value.
> 3. changing the pass-by `out` syntax `out i` from being lowered to an argument intended for a `cpp2::out` parameter to being lowered to a value of `cpp2::out`

With these changes, the motivating example in this PR description now works, and `out` is more deducible.

Thanks again for the examples and discussion!

**Question:** Even though the top example now works, I'm leaving this open for @filipsajdak to check and confirm... Filip, do these changes sufficiently address your motivation for this PR, so we should close it? Or are there additional examples that motivate the PR we should consider, and add those examples to this thread and rebase the PR?

hsutter:
> Please, see [#198 (comment)](https://github.com/hsutter/cppfront/pull/198#issuecomment-1529282080).

@JohelEGP Indeed, I've added a comment there asking you to please upgrade it to a whole new issue for argument-side qualifiers, so we don't lose this. I think we're getting to where we understand those well enough to enable more/all of them. Thanks for the reminder!

filipsajdak:
I will check the examples tomorrow and will let you know.

filipsajdak:
I have no other examples... I think we can close it.

PR #371, opened 2023-04-14
Fix showing qualifier in missing return type error

ntrel:
Triggered by:
```cpp
f: (inout x: int) -> forward = x;
```
Currently:
```
retfwd.cpp2(1,24): error: ''' must be followed by a type-id (at '=')
```
Note: Not sure why unary `+""` is a valid expression in Cpp1 (or Cpp2).

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -4890,7 +4890,7 @@ private:
                 else {
                     auto msg = std::string("'");
                     msg += to_string_view(pass);
-                    error( + "'' must be followed by a type-id");
+                    error(msg + "' must be followed by a type-id");
                 }
             }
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks!

> Note: Not sure why unary `+""` is a valid expression in Cpp1 (or Cpp2).

Oh, that's hilarious. The string literal's type in `char[N]`, which helpfully decays to `char*`, which helpfully allows arithmetic. Ick. Not worth preventing in Cpp2 right now, as we have more important things to do, but eventually we'd like to prevent all the lossy (which includes "narrowing" but also this) built-in type conversions.

Issue #325, opened 2023-04-07
[BUG] Returning a tuple with unnamed members


ntrel:
Possibly this should be an enhancement request. Some of these errors may be because the feature is not implemented? Perhaps I'm using the wrong syntax? If so, better diagnostics would be nice, especially for `f` (see 4).

**To Reproduce**
```c++
d: () -> (int, int) = (1,2);
e: () -> (int, int) = { return (1,2); }
f: () -> _ = (5, "hi");
g: () -> (int, std::string) = (5, "hi");

main: () -> int = {
    auto [a,b] = f();
    assert(a == 5);
    assert(b == "hi");
}
```
1. `d` can't be initialized by an expression:
```
rettup.cpp2(3,1): error: a function with named return value(s) must have a full { } body
```
But `d` doesn't used named return values. Also `f` can be initialized by a tuple expression without needing `{}`.

2. Commenting out `d`, spurious compile error for `e`:
```
rettup.cpp2(1,16): error: int - variable must be initialized on every branch path
```
There is no declared variable.

3. Commenting out `e`, spurious compile error for `g`:
```
rettup.cpp2(3,19): error: expected , in parameter list (at '::')
rettup.cpp2(3,19): error: missing function return after -> (at '::')
```
Probably it's parsing `std` as an identifier.

4. Commenting out `g`, `f` compiles OK but doesn't declare or use a tuple struct so the c++ compiler errors:
```c++
#line 2 "old/rettup.cpp2"
[[nodiscard]] auto f() -> auto { return 5, "hi";  }

#line 5 "old/rettup.cpp2"
[[nodiscard]] auto main() -> int{
    cpp2::assert_in_bounds(auto, a, b) = f();
```

JohelEGP:
https://github.com/hsutter/cppfront/blob/main/regression-tests/mixed-multiple-return-values.cpp2 doesn't demonstrate it, but `f: () -> (i: int, s: std::string)` returns a structure with data members `i` and `s`, just like in the parameter list.

ntrel:
@JohelEGP Yes, this is about unnamed members. BTW is there a pure cpp2 way to write `auto [a,b] = f();`?

JohelEGP:
> @JohelEGP Yes, this is about unnamed members.

Ah, I see. IIRC, you're right here.

> Possibly this should be an enhancement request.

> BTW is there a pure cpp2 way to write `auto [a,b] = f();`?

Not yet, it seems.

[...]

PR #376, opened 2023-04-15
Check return parameter has identifier and type

ntrel:
Note: type cannot be `_`.
Fixes most of https://github.com/hsutter/cppfront/issues/325#issuecomment-1509817899.


filipsajdak (changes requested):
.

filipsajdak (changes requested):
.

source/parse.h
```diff
@@ -4648,7 +4648,17 @@ class parser
             error("Cpp2 is currently exploring the path of not allowing default arguments - use overloading instead", false);
             return {};
         }
-
+        if (is_named && is_returns) {
+            auto tok = n->name();
```
hsutter, line 4652: Might still want to `assert(tok);` too.

source/parse.h
```diff
@@ -4648,7 +4648,17 @@ class parser
...
+            if (tok->type() != lexeme::Identifier) {
+                error("expected identifier, not '" + tok->to_string(true) + "'",
+                    false, tok->position());
+            }
+            else if (n->declaration->type.index() != declaration_node::active::a_type) {
```
hsutter, line 4657: This can be `n->declaration->is_type()` now

hsutter (changes requested):

regression-tests/pure2-return-tuple-no-identifier.cpp2
```diff
@@ -0,0 +1,3 @@
+d: () -> (int, int) = (1,2);
```
hsutter, line 1: This is intended to emit an error, and test filenames that should produce errors should have names ending with `-error.cpp2`.

ntrel:
Thanks, all 3 review points are done now.

source/parse.h
```diff
@@ -4648,7 +4648,18 @@ class parser
...
         }
-
+        if (is_named && is_returns) {
+            auto tok = n->name();
+            assert(tok);
+            if (tok->type() != lexeme::Identifier) {
+                error("expected identifier, not '" + tok->to_string(true) + "'",
+                    false, tok->position());
+            }
+            else if (!n->declaration->is_type()) {
```
hsutter, line 4658: Sorry, I gave an incomplete feedback on this line the first time: I think the right query is not "is this declaration a type" (a parameter shouldn't ever be a type, it should be an object) but rather "does this declaration have a concrete type"... so I think what you want is `else if (n->declaration->has_wildcard_type())`.

Did you run regressions? I think the current condition above would reject existing valid named return cases?

ntrel: Sorry, last night I checked the error tests but I forgot to check the correct named return test. Now fixed.

hsutter:
Looks good, thanks!

PR #379, opened 2023-04-16
Update regression tests for apple-clang-14 (23W15)

filipsajdak:
Long time no update.

It looks like apple-clang-14 has no support for polymorphic memory resources - (according to the compiler support table, it is supported by clang-16: https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features).

image

We shall replace `pure2-type-and-namespace-aliases.cpp2` to code that is portable and is supported by GCC, clang, MSVC.


filipsajdak:
I double check that but it looks like std::ranges are supported!

hsutter:
Thanks!

PR #380, opened 2023-04-17
[PROPOSE] Refactor args_t to use ranges instead of vector

filipsajdak:
In the current cppfront implementation (65fcd0f), we are using `std::vector<std::string_view>` to store `args` in `main()`. This change proposes to use std::ranges that can use lazy evaluation of ranges.

Thanks to this change we can avoid one allocation that is unnecessary.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -213,6 +213,7 @@
     #include <cstdio>
     #include <cstdint>
     #include <algorithm>
+    #include <ranges>
 
     #ifndef CPP2_NO_EXCEPTIONS
         #include <exception>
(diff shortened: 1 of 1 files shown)


JohelEGP (approved):
Definitely an improvement.

You could try either of these to not make `args_t` a template: <https://compiler-explorer.com/z/E9r7xcoMo>.

filipsajdak:
I have been thinking about your approach - I did not like the lambda living somewhere outside. If Herb will be open to making it dependent on ranges we can choose one solution that will fit the best. 

Thanks for feedback!

JohelEGP:
Only it's type lives outside. The view holds its instance.

filipsajdak:
@JohelEGP I have slept over it and decided to follow the path you propose but move the lambda directly to the type definition - like here: https://godbolt.org/z/Mvv6a4rs6 (I think the code looks simpler).

Thanks for the excellent feedback!

JohelEGP (approved):
Yeah, that's much better. Thank you.

MichaelCook (approved):

filipsajdak:
That is fine. The implementation is ready when there will be time to introduce it.

PR #384, opened 2023-04-18
Add diagnostics for returning a temporary, closes #381

### LINKED ISSUE #381: [BUG] Forward returning literals and rvalues not rejected.
This Cpp2 code
```
global: i32 = 42;

forward_return_literal: () -> forward _ = {
    return 1;
}

forward_return_rvalue: () -> forward _ = {
    return global + 1;
}
```
compiles to
```cpp
cpp2::i32 global {42}; 

[[nodiscard]] auto forward_return_literal() -> auto&&{
    return 1; 
}

[[nodiscard]] auto forward_return_rvalue() -> auto&&{
    return global + 1; 
}
```
which is problematic Cpp1. We should err when forward returning literals and rvalues.

filipsajdak:
The current implementation of cppfront (65fcd0f7cd7ccf07c4493d4f133505e3bab0b4fe) does not provide diagnostics for forward return temporary variables, e.g.:

```cpp
forward_return_literal: () -> forward _ = {
    return 1;
}

forward_return_rvalue: () -> forward _ = {
    return global + 1;
}
```
Generates:
```cpp
cpp2::i32 global {42}; 

[[nodiscard]] auto forward_return_literal() -> auto&&{
    return 1; // return reference to literal
}

[[nodiscard]] auto forward_return_rvalue() -> auto&&{
    return global + 1; // return reference to temporary variable
}
```

This change provides better diagnostics and handles most operators that might create a temporary variable:
* binary expression: `/`, `<<`, `<=>`, `<`, `<=`, `>>`, `>=`, `>`, `+`, `-`, `||`, `|`, `&&`, `*`, `%`, `&`, `^`, `==`, `!=`,
* prefix expression: `!`,
* postfix expression: `&`, `~`

All other operators that return references are accepted. 

All regression tests pass. Closes #381 

### Need to be checked

The below cases still might not be supported:

* using operators that return a reference on local variables,
* calling function that returns a non-reference,

### DIFF
--- a/regression-tests/pure2-forward-return-diagnostics-error.cpp2
+++ b/regression-tests/pure2-forward-return-diagnostics-error.cpp2
@@ -1,0 +1,53 @@
+fun1:  () -> forward _ = { return global   / 1; } // error: a 'forward' return type cannot return a temporary variable
+fun4:  () -> forward _ = { return global  << 1; } // error: a 'forward' return type cannot return a temporary variable
+fun5:  () -> forward _ = { return global <=> 1; } // error: a 'forward' return type cannot return a temporary variable
+fun6:  () -> forward _ = { return global   < 1; } // error: a 'forward' return type cannot return a temporary variable
+fun7:  () -> forward _ = { return global  <= 1; } // error: a 'forward' return type cannot return a temporary variable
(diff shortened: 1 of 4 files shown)


hsutter:
Thanks!

PR #418, opened 2023-05-05
Add hint for Cpp1 new expression

ntrel:


hsutter:
Thanks! I like this, and have applied it in a separate commit because I think the message is generally useful beyond *assignment-expression* only; e.g., it seems useful to diagnose it in cases like `f( new int )` too.

Issue #416, opened 2023-05-04
[BUG] Improve diagnostic for prefix increment/decrement


JohelEGP:
**Title**: Improve diagnostic for prefix increment/decrement.

**Minimal reproducer** (<https://cpp2.godbolt.org/z/6TEhs1d1o>):

```Cpp2
main: (args) = {
  a := 0;
  ++a;
}
```

**Commands**:

```bash
cppfront -clean-cpp1 main.cpp2
```

**Expected result**:

A diagnostic that explains
<https://github.com/hsutter/cppfront/wiki/Design-note%3A-Postfix-operators#-and--->.

**Actual result and error**:

```output
main.cpp2(1,16): error: ill-formed initializer (at '{')
main.cpp2(1,1): error: unexpected text at end of Cpp2 code section (at 'main')
main.cpp2(1,0): error: parse failed for section starting here
```

PR #419, opened 2023-05-05
Add hint for Cpp1 prefix operators

ntrel:
I'm assuming no Cpp2 expression starts with one of those tokens.


JohelEGP:
Resolves #416.

hsutter:
Thanks!

I'll merge this and apply the following updates separately, but for future reference:
- Regression tests that are intended to fail should be named "*-error.cpp2" so the quick sanity-check counts in my `run-tests` script account for them correctly
- Just a nit: These days I try to keep messages of the form "<thing> is not allowed, use <thing> instead" or "<thing> may not <thing>, did you mean <thing>?", though I'm sure I violate that all over the place myself esp. with earlier messages

Again, thanks.

PR #441, opened 2023-05-09
FIX: Remove next() on backtrack, fixes #404

### LINKED ISSUE #404: [BUG] Postfix operator to parenthesized expression isn't parsed in a statement
**Title**: Postfix operator to parenthesized expression isn't parsed in a statement.

**Minimal reproducer** (<https://cpp2.godbolt.org/z/eTGKsMdKn>):
```Cpp2
main: (args) = { (args)&; }
```

**Commands**:
```bash
cppfront -clean-cpp1 main.cpp2
```

**Expected result**:
Same as (<https://cpp2.godbolt.org/z/x8qov5xvE>):
```C++
main: (args) = (args)&;
```

**Actual result and error**:
```output
main.cpp2(1,16): error: ill-formed initializer (at '{')
main.cpp2(1,1): error: unexpected text at end of Cpp2 code section (at 'main')
main.cpp2(1,0): error: parse failed for section starting here
```

filipsajdak:
When trying to parse expressions like:
```cpp
(:() = 0) as t;
```

`parameter_declaration_list()` fail to succeed, which is expected but skips one symbol ('(') on returning, which makes the parser fail to process the expression successfully.

After this change above expression is parsed correctly. All regression tests pass.

Fixes #404

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -5523,7 +5523,6 @@ private:
             else {
                 error("invalid parameter list", true, {}, true);
             }
-            next();
             return {};
         }
 


hsutter:
Ah, thanks! Right, if we backtrack we shouldn't then do another `next()`, good catch.

PR #444, opened 2023-05-09
FIX: parsing of string literals that contains `//`, closes #442

### LINKED ISSUE #442: [BUG] `"//"` `"/*"` inside string literal treated as real comment and doesn't compile
**Describe the bug**
`"//" "/*" symbols inside string literals treated as comment start and break compilation, like shadowing closing }
For // it happens inside { block }

**To Reproduce**
https://cpp2.godbolt.org/z/5KEPj196s
Uncomment lines marked as error
```
main: () = {
	std::cout << "//123"; //123, ok
	//{ std::cout << "//456"; } //error

	:()->_ = "//Ok";
	//:()->_ = { return "//Error"; }

	a:="*/ ok";
	{ c:="*/ ok"; }
	//b:="/* error even without {} block";
}
//error: (x) = { return "//"; }
```
**Version**
 latest
**Command lines**
cppfront/cppfront $1.cpp2 -p
clang++-15 -Icppfront/include $1.cpp -std=c++20 -o $1

**Expected result**
Success compilation and usage of such string literals
**Actual result/error**
`call.cpp2(32,0): error: end of file reached with 1 missing } to match earlier { on line 1`

**Additional context**
I've encountered this inside lambda, when expanding it and changing short syntax to code block. Then discovered `"/*"` have same issue even without block.

`"*/"` doesn't seems to have problem.

filipsajdak:
In the following code:
```cpp
main: () = {
	{ std::cout << "//456"; } //error

	:()->_ = { return "//Error"; };

	b:="/* error even without {} block";
}
error: (x) -> _ = { return "//"; }
```
`//` and `/*` inside string literals are parsed as the beginning of comments.

Current change solves that. This FIX does not handle the handling of raw string literals.

Closes #442

### DIFF
--- a/source/io.h
+++ b/source/io.h
@@ -693,11 +693,18 @@ auto process_cpp2_line(
     auto found_end = false;
 
     auto prev = ' ';
+    auto in_string_literal = false;
+
     for (auto i = colno_t{0}; i < ssize(line); ++i) {
 
         if (in_comment) {
             switch (line[i]) {
             break;case '/': if (prev == '*') { in_comment = false; }
+            break;default: ;
+            }
+        } else if (in_string_literal) {
+            switch (line[i]) {
+            break;case '"': if (prev != '\\') { in_string_literal = false; }
             break;default: ;
             }
         }
@@ -731,6 +738,9 @@ auto process_cpp2_line(
 
             break;case '/':
                 if (prev == '/') { in_comment = false; return false; }
+
+            break;case '"':
+                if (prev != '\\') { in_string_literal = true; }
 
             break;default: ;
             }


JohelEGP:
> This FIX does not handle the handling of raw string literals.

I thought #69 did that.

filipsajdak:
Herb changes the parsing stages. There is a new simpler one that did not cover strong literals not raw strong literals.

hsutter:
> Herb changed the parsing stages. There is a new simpler one that did not cover string literals not raw string literals.

Could you remind me / elaborate?

filipsajdak:
> > Herb changed the parsing stages. There is a new simpler one that did not cover string literals not raw string literals.
> 
> Could you remind me / elaborate?

Hmmm... I made a mistake with this statement. I mistakenly treat a switch that is in the `process_cpp2_line()` function (in the `io.h` file) with the switch that is in the `lex_line()` function (in the `lex.h` file). I could not see parts that take into account raw string literals, and I thought that it was somehow lost during refactoring... my mistake.

Now I see that the `process_cpp2_line` stage was there all the time... I need to add support for raw string literals as well.

hsutter:
OK, thanks -- for now I've reopened #442 to track this part too, and the issue will then be re-closed via the coming new PR to apply the fix to raw strings.

Again, thanks!

PR #460, opened 2023-05-19
Update otdated regression test results

jarzec:
The 6 regression tests whose results are updated fail for g++-10 and clang-12 on Ubuntu.
It seems a commit that cleaned up the cpp1 output of cppfront failed to update those test results.


hsutter:
Thanks! I missed those, but I think they're up to date now when I ran another regression test run before seeing this. Please reopen if I missed any.

Again, thanks!

Issue #308, opened 2023-03-29
[BUG] Failing to build cppfront


AbhinavK00:
**Describe the bug**
cppfront build with clang fails with the following error
```
In file included from cppfront.cpp:18:
In file included from ./sema.h:21:
In file included from ./parse.h:21:
In file included from ./lex.h:21:
In file included from ./load.h:21:
In file included from ./common.h:18:
In file included from ./../include/cpp2util.h:203:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/memory:66:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/stl_uninitialized.h:67:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/ext/alloc_traits.h:34:
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/alloc_traits.h:514:4: error: no matching function for call to 'construct_at'
          std::construct_at(__p, std::forward<_Args>(__args)...);
          ^~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/vector.tcc:115:21: note: in instantiation of function template specialization 'std::allocator_traits<std::allocator<cpp2::cppfront::function_info>>::construct<cpp2::cppfront::function_info, const cpp2::token *, cpp2::declaration_node::declared_that_funcs>' requested here
            _Alloc_traits::construct(this->_M_impl, this->_M_impl._M_finish,
                           ^
cppfront.cpp:4317:35: note: in instantiation of function template specialization 'std::vector<cpp2::cppfront::function_info>::emplace_back<const cpp2::token *, cpp2::declaration_node::declared_that_funcs>' requested here
            current_function_info.emplace_back(
                                  ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/stl_construct.h:94:5: note: candidate template ignored: substitution failure [with _Tp = cpp2::cppfront::function_info, _Args = <const cpp2::token *, cpp2::declaration_node::declared_that_funcs>]: no matching constructor for initialization of 'cpp2::cppfront::function_info'
    construct_at(_Tp* __location, _Args&&... __args)
    ^
1 error generated.
```

**To Reproduce**
Try to build cppfront with the following command
[...]

PR #462, opened 2023-05-20
Add initial CI: Build steps for supported compilers

DyXel:
This PR adds a initial Github Actions workflow to compile cppfront for several platforms, compilers and compiler versions.

Compilers tested: gcc 10, gcc 11, gcc 12, clang 12, clang 14, Apple's clang 14 and Microsoft's vs2022.

Benefits of adopting CI right now:
 * Ensures that all supported compilers are actually tested (for build right now, for regressions/passthrough later)
 * Provides information to possible adopters that their compiler and compiler version is supported.
 * Eases the burden of test-building in different platforms for developers and contributors.

If interested we could move onto automating regression testing and passthrough testing next.

### DIFF
--- a/.github/workflows/build-cppfront.yaml
+++ b/.github/workflows/build-cppfront.yaml
@@ -1,0 +1,45 @@
+name: Multi-platform Build of cppfront
+on: push
+jobs:
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ilammy/msvc-dev-cmd@v1
+      - name: Compiler name & version
+        run: cl.exe
+      - name: Build
+        run: cl.exe source/cppfront.cpp -std:c++latest -MD -EHsc -experimental:module -W4
+  build-unix-like:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubuntu-latest]
+        compiler: [g++-10, g++-11, g++-12, clang++-12, clang++-14]
+        cxx-std: ['c++20', 'c++2b']
+        exclude:
+          # GCC 10 doesn't have support for c++23
+          - compiler: g++-10
+            cxx-std: 'c++2b'
+          # Clang 12 and 14 do not compile on 'c++2b' due to llvm/llvm-project#58206
+          - compiler: clang++-12
+            cxx-std: 'c++2b'
+          - compiler: clang++-14
+            cxx-std: 'c++2b'
+        include:
+          - runs-on: macos-latest
+            compiler: clang++
+            cxx-std: 'c++20'
+    runs-on: ${{ matrix.runs-on }}
+    env:
+      CXX: ${{ matrix.compiler }}
+      CXXFLAGS: -std=${{ matrix.cxx-std }} -Wall -Wextra -Wold-style-cast -pthread
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install compiler
+      if: startsWith(matrix.runs-on, 'ubuntu')
+      run: sudo apt-get install -y $CXX
+    - name: Compiler name & version
+      run: $CXX --version
+    - name: Build
+      run: $CXX source/cppfront.cpp $CXXFLAGS -o cppfront


hsutter:
Thanks! I confess I don't understand CI and GitHub actions well, but I'm open to trying this...

Can someone else who does understand this review this PR and give some commentary?

Also, in case it's useful, here are the flags I'm currently using for GCC, Clang, and MSVC in my local "make sure cppfront compiles clean" tests:

- GCC: `g++-10 cppfront.cpp -std=c++20 -Wall -Wextra -Wold-style-cast`

- Clang: `clang++-12 cppfront.cpp -std=c++20 -pthread -Wall -Wextra`

- MSVC: `cl.exe cppfront.cpp -std:c++latest -MD -EHsc -experimental:module -W4`

jarzec:
In fact a couple of months back I proposed simple GitHub CI that included a bash script for running tests (see https://github.com/hsutter/cppfront/issues/150 and https://github.com/hsutter/cppfront/pull/151). I did not go for CMake as I understood the goal was to limit dependencies in the project. Still, at that point @hsutter said it was to early a stage to already include that.
I've recently had some time to look at the project after a few months and I am quite impressed by the progress. An updated version of my bash script proved to be useful again as I found outdated tests that were, however, fixed through other means in the meantime.

Depending on how far we are ready to go e.g. to be able to run tests on both Linux, Windows and Mac we could look into an analogue of a bash script written e.g. in PowerShell or go all the way to a CMake that would allow for configuring builds and tests on all three platforms. This, however, adds a dependency to CMake and to an actual build system, like Ninja.

Another thing to consider about CI on GitHub are the limits. The free package includes 2000 CI/CD minutes/month. In addition only on Linux each minute counts as one towards the monthly limit. One minute of running time on Windows counts as 2. On mac it is even worse as every minute counts as 10. The CI is quite configurable, but care should be taken when planning all that.

To give some perspective: on my WSL Ubuntu a single compilation of cppfront with g++10 takes ~12s and with clang++-12 ~7s.
A single run of all regression tests using my bash script takes just below 3 minutes depending on the compiler.

In the current proposition the CI action that compiles cppfront itself on multiple platforms with multiple compilers is run on every push to any branch. This might be problematic for contributors who like to push often to their work-in-progress branches as the builds of each such push in their own forks will count towards their own CI limit.

As much as running CI on every push is cool, I would propose to start with CI runs only on pushes to the main branch to evaluate the usage. If the 2000 minutes run out too quickly e.g. the number of compilation options (compiler versions, C++ standards) could be limited. If not addition of test runs should be considered.

Alternatively, the CI actions could be executed on pull requests. The main branch could be protected from direct pushes - all changes would have to go through pull requests. The pull requests would be tested before merging. Failures would prevent merging of breaking changes into main. In addition e.g. nightly builds of the main branch itself could be scheduled.

Having said all that I agree a discussion is needed as to the nature/complication of CI for cppfront and a path to get there needs to be outlined.

DyXel:
I agree that we should keep the build process as simple as possible, I don't think at this time we need something like CMake or Ninja, perhaps as the complexity of the experiment grows we might consilder alternatives, but I think just having to run a single command to get a working cppfront has its beauty to it (its what drove me to test it to begin with!).

For usage limits, I'd say there's no need to worry about it as Actions usage is free for public repositories on standard Github-hosted runners or self-hosted runners. Even if the 2000 minutes limit was there I think Herb should still be good to go with current activity in the repo.

Indeed, the idea is to start the discussion again about automating the mundane parts (such as checking that every compiler works and regression/passthrough tests pass) so that developers can focus on the important implemention bits! I just felt like doing this after seeing https://github.com/hsutter/cppfront/pull/460, https://github.com/hsutter/cppfront/issues/308, etc.

jarzec:
My bad, I re-read the GitHub pricing page and in deep public repos have no limits. Given that this is the case I agree that the proposed workflow is a good idea. I saw you added the warning level flags suggested above. To simplify things you used the same flags for both clang and g++ which are a union of what was suggested for each compiler, but that looks reasonable.

I have just one more question to @hsutter: Given that this PR will likely go in, would you be interested in having another GitHub workflow that runs tests? I agree with @DyXel that CMake seems overkill at this stage, but what would you say for a simple bash script for Linux that runs regression tests and performs basic checks? The script would come in handy for contributors like your own `passthrough-tests/run-diffs.sh` or the `regression-tests/test-results/run-tests.bat` for Windows.

I could add another workflow that runs the tests that would be an evolution of https://github.com/hsutter/cppfront/issues/150.
Both the workflow proposed in this PR and possibly a testing one could be run on every push so that issues including failing tests could be quickly identified.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
+name: cppfront CI
```
jarzec, line 1: [Suggestion] I think the general practice with GitHub workflows is to make both the YAML file name and the name displayed in the GitHub UI more descriptive. How about something like `build-cppfront.yaml` and:
```suggestion
name: Multi-platform Build of cppfront
```
or something like that?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+      - uses: ilammy/msvc-dev-cmd@v1
+      - name: Compiler name & version
+        run: cl.exe
+      - name: Build
+        run: cl.exe source/cppfront.cpp -std:c++latest -MD -EHsc -experimental:module -W4
+  build-unix-like:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubuntu-20.04, ubuntu-22.04, macos-latest]
```
jarzec, line 17: [Question] What is the added value of running older compilers on ubuntu-20.04 and the newer ones on ubuntu-22.04? Running all the versioned g++ and clang variants on ubuntu-22.04 and skipping ubuntu-20.04 altogether would shorten  the list of exclusions by half, wouldn't it?

.github/workflows/ci.yml
DyXel: Done!

.github/workflows/build-cppfront.yaml
```diff
@@ -0,0 +1,56 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubuntu-latest, macos-latest]
+        compiler: [g++-10, g++-11, g++-12, clang++-12, clang++-14, clang++]
+        cxx-std: ['c++20', 'c++23']
+        exclude:
+          # Unversioned clang++ doesn't run on ubuntu
+          - runs-on: ubuntu-latest
+            compiler: clang++
```
JohelEGP, line 23: Besides this one,
I think you can remove all other exclusions
by changing `c++23` to `c++2b`,
removing `macos-latest` from `runs-on`,
and adding
```yaml
        include:
          - runs-on: macos-latest
            compiler: clang++
```

DyXel: I did some testing by changing `c++23` to `c++2b` and managed to simplify the matrix by using `include` as suggested. Summary:

- `gcc-10` doesn't recognize either `c++2b` or `c++23`
- `clang++-12` does recognize `c++2b` but fails to compile, doesn't recognize `c++23`
- `clang++-14` same as above

The only exclusion now is `gcc-10` on the `c++2b` configuration, also, I have attached the build logs of the failed builds since this doesn't seem intentional (why would it compile with `c++20` but not `c++2b`?).

[7_build-unix-like (ubuntu-latest, clang++-12, c++2b).txt](https://github.com/hsutter/cppfront/files/11534864/7_build-unix-like.ubuntu-latest.clang%2B%2B-12.c%2B%2B2b.txt)
[9_build-unix-like (ubuntu-latest, clang++-14, c++2b).txt](https://github.com/hsutter/cppfront/files/11534884/9_build-unix-like.ubuntu-latest.clang%2B%2B-14.c%2B%2B2b.txt)

jarzec: I am not sure about using `c++2b` with compiler versions that do not support `c++23` (as I would be with `c++2a` instead of `c++20`). There is a good reason for this lack of support - the standard is not fully implemented in the particular compiler version. A build failure might actually be the result of shortcomings of the compiler rather than issues in the code. It is true there might be bugs even with full support, but those are much less likely.

DyXel: I will leave up to Herb to decide if we support partial implementations, what I find weird here is that `c++2b` > `c++20`, yet `c++20` compiles but `c++2b` does not. I would expect for cppfront to still compile if the user opts-in to a partial implementations that is newer than `c++20`...

.github/workflows/build-cppfront.yaml
```diff
@@ -0,0 +1,40 @@
...
+            cxx-std: 'c++20'
+    runs-on: ${{ matrix.runs-on }}
+    env:
+      CXX: ${{ matrix.compiler }}
+      CXXFLAGS: -std=${{ matrix.cxx-std }} -Wall -Wextra -Wold-style-cast -pthread
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install compiler
+      if: startsWith(matrix.runs-on, 'ubuntu')
+      run: sudo apt-get install -y ${{ matrix.compiler }}
```
gregmarr, line 36: Should this be `$CXX` instead of `${{ matrix.compiler }}` for consistency, or can that not be used there?

.github/workflows/build-cppfront.yaml
JohelEGP: The issue seems to be <https://github.com/llvm/llvm-project/issues/58206>.

.github/workflows/build-cppfront.yaml
DyXel: I think it could be referenced directly yeah, but defining the `CXX ` envvar is also super common (specially when using build systems that expect those vars to be set).

.github/workflows/build-cppfront.yaml
jarzec: As I said previously, I was pretty sure this is a compiler bug not an issue in the code.

DyXel: Do we disable those 2 explicitly then for the time being? I can reference the issue in a comment in that case.

.github/workflows/build-cppfront.yaml
gregmarr: > I think it could be referenced directly yeah

I was actually saying the opposite:

    run: sudo apt-get install -y $CXX

.github/workflows/build-cppfront.yaml
JohelEGP: Please, do.

DyXel: Done!

.github/workflows/build-cppfront.yaml
DyXel: Done!

JohelEGP (approved):
The Unix side looks fine.

gregmarr (approved):

jarzec (approved):

.github/workflows/build-cppfront.yaml
JohelEGP: I was able to compile Cppfront with <https://github.com/llvm/llvm-project/commit/37b474a20e8a731bec9db7033e02013c07bd6fa4>.

JohelEGP:
This could have been the first step to eventually help prevent #744.

.github/workflows/build-cppfront.yaml
JohelEGP: Opened #818 to workaround <https://github.com/llvm/llvm-project/pull/65456>.

JakeShirley:
Looks like c1d3e0edc8b2e2a2c035f68b6c0012c5b7049a07 also fixed some things that created unclean tests.  CI would be great!

hsutter:
Thanks for your patience on this one.

Okay, this PR is just adding a single `.yaml` file. I don't deeply understand it, and I dislike merging things I don't understand and will have to maintain... but just adding a file seems low-risk, and presumably if there are issues I can always just delete it later?

So I think my main question should be: If I merge this PR, what will be the visible result in this repo?

Is there any reason not to merge this now?

---

> I have just one more question to @hsutter: Given that this PR will likely go in, would you be interested in having another GitHub workflow that runs tests? I agree with @DyXel that CMake seems overkill at this stage, but what would you say for a simple bash script for Linux that runs regression tests and performs basic checks? The script would come in handy for contributors like your own `passthrough-tests/run-diffs.sh` or the `regression-tests/test-results/run-tests.bat` for Windows.

OK, I'm ready to have a workflow that runs tests, ideally if it's as lightweight and noninvasive as this one which just adds a file.

DyXel:
Hey hello, indeed it's been a while. To your questions:

> if there are issues I can always just delete it later?

Yes, in fact, you can just disable the Actions at any point via the Actions tab. IIRC, for PRs, I think you need to give explicit permissions to enable running first for non-contributors (if anybody knows how to double-check please feel free to share).

> If I merge this PR, what will be the visible result in this repo?

When when you commit or if anybody opens a PR (and if you give them permissions), the CI will attempt to build cppfront for the matrix of compilers specified in the yaml and it will show if said builds were succesful for that commit or PR, as in, if they managed to produce the cppfront executable. The action will only succeed if _all_ currently selected compilers built, which makes trivial to see if the code change introduced any compilation problem for any compiler.

> Is there any reason not to merge this now?

Hmm, I would say that this could potentially be merged as-is if nobody has arguments for it, and then later we can add regression testing steps (I'd be happy to help in that case).

jarzec:
> > I have just one more question to @hsutter: Given that this PR will likely go in, would you be interested in having another GitHub workflow that runs tests? I agree with @DyXel that CMake seems overkill at this stage, but what would you say for a simple bash script for Linux that runs regression tests and performs basic checks? The script would come in handy for contributors like your own `passthrough-tests/run-diffs.sh` or the `regression-tests/test-results/run-tests.bat` for Windows.
> 
> OK, I'm ready to have a workflow that runs tests, ideally if it's as lightweight and noninvasive as this one which just adds a file.

Great! What I have so far in a slightly outdated branch is a single .yaml file with the workflow, similar to the one in this PR + a single bash script for running regression tests. I made this bash script portable, so that it can also run tests on Windows through Git Bash. As a result the very same script can be used to run regression tests on Mac, Linux and Windows, both locally and through GitHub actions. If that sounds interesting I can update the mentioned branch and open a PR.

hsutter:
I'll take that as "yes this is ready to merge" and I'll do that.

After merge, I evidently need to enable Actions... presumably the UI will be intuitive, otherwise I'll have questions :)

hsutter:
Here's what I see -- any ideas?

image

image

DyXel:
If you navigate the erroring build (the one with the red X) ( https://github.com/hsutter/cppfront/actions/runs/7234428378/job/19710793888 ), you can see the problem:
Screenshot_20231216_225640
My guess would be a missing header?

hsutter:
The usual `<algorithm>` header contains both `std::find_if` and `std::ranges::find_if` (if implemented).

I thought Apple Clang supported ranges, as of 14.0.3? Maybe that one isn't being built in C++20 mode, or this is using a pre-14.0.3 compiler?

DyXel:
```
Target: x86_64-apple-darwin21.6.0
```
From the same job. Seems like the macOS runner has 14.0.0 as latest. I don't know how much we care about supporting that specific version. You could find a workaround, or if that is too tedious/not worth, you can disable that job run in the yaml by removing the inclusion in lines 29-32.

hsutter:
😁 "Teach a man to fish..."

OK, looks like Apple Clang 14.0.3 is only available since Xcode 14.3, released in March 2023. So the VM/devbox must not be updated yet, I guess.

hsutter:
At first I disabled it, then I re-enabled it and just removed the use of `ranges::find_it`.

That's the only use of `ranges::` in the codebase, and I want cppfront to compile on reasonably recent compilers. Since Apple Clang 14 only started supporting ranges in March 2023, and that seems to be the only thing stopping use of Clang 12+ even on Xcode (AFAIK), it seems more reasonable to do this and not break Xcode Clang 12+ builds of cppfront.

And yes, this PR exposed that. 🎉 Thanks!

DyXel:
By the way, my recommendation if you don't want to start doing "fix compilation on X compiler" commits (or force pushes 💀 ) for the rest of times (I know, been there!), I would suggest first comitting to your own branches and/or opening PRs and then merging to main after CI completes. That way you also follow the flow that everybody else does, and then later you can enforce it (for example, by marking the main branch as protected).

hsutter:
Say, n00b question here: If I want to enable MacOS Apple Clang 12 testing, can I just change `compiler: clang++` to `compiler: [clang-12, clang++]`?

DyXel:
Assuming the runner has it installed, yes, you can follow the pattern used for Ubuntu. It does seem like the latest runner for macOS only has Clang 14.0.0 and Clang 15.0.0 though 😦  ( https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md )

Issue #387, opened 2023-04-21
[DOCS] Assorted comments on the grammar productions


JohelEGP:
### Table of contents

- Fixes
  * [**_alt-name_**](#alt-name)
  * [**_declaration_**](#declaration)
  * [**_return-list_**](#return-list)
  * [**_string-literal_**](#string-literal)
  * [**_template-argument-list_**](#template-argument-list)
- Need fixing
  * [**_name-specifier_**](#name-specifier)
  * [**_integer-suffix_**](#integer-suffix)
  * [**_interpolation_**](#interpolation)
- Simplifications
  * [**_type-id_**](#type-id)
  * [**_is-as-expression-target_**](#is-as-expression-target)
- Others
  * [**_parameterized-statement_**](#parameterized-statement)
- Outdated `//G` comments
  * [**_parameter-declaration_**](#parameter-declaration)
  * [**_punctuator_**](#punctuator)
  * [**_unqualified-id_**](#unqualified-id)
  * [**_operator_**](#operator)
  * [**_selection-statement_**](#selection-statement)

See also: #71, #392, #408.

JohelEGP:
### Others

**_parameter-declaration-list_**:
<a id="parameter-declaration-list"></a>
<a id="return-list"></a>

```C++
//G return-list:
//G     '->' type-id
//G     '->' parameter_declaration_list
```
The `'->' parameter_declaration_list` is target of #343.
It should probably be
`'->' type-qualifier-seq parameter-declaration-list`.
Actually, it should probably be a _type-id_.
```diff
 //G type-id:
-//G     type-qualifier-seq? qualified-id
-//G     type-qualifier-seq? unqualified-id
+//G     type-qualifier-seq? id-expression
+//G     type-qualifier-seq? parameter-declaration-list

 //G parameter-declaration:
 //G     this-specifier? parameter-direction? declaration
+//G     this-specifier? parameter-direction? identifier
+//G     this-specifier? parameter-direction? unnamed-declaration

 //G return-list:
 //G     '->' type-id
-//G     '->' parameter_declaration_list
```

**_inspect_**:
<a id="inspect"></a>

`inspect` should be able to result in a pointer to a function type.
Maybe it's just out of date.
```diff
 //G inspect-expression:
-//G     'inspect' 'constexpr'? expression '{' alternative-seq? '}'
-//G     'inspect' 'constexpr'? expression '->' type-id '{' alternative-seq? '}'
[...]

PR #476, opened 2023-05-30
Fix for operator recognition in cpp2 code sections.

MaxSagebaum:
For the cpp2 file:
```
operator*: (v1: double, v2: double) -> _ = v1 * v2;
```
The cppfront command `cppfront opMul.cpp2 -p` produces the error: 
```
opMul.cpp2(1,1): error: pure-cpp2 switch disables Cpp1 syntax
```

I took this as an opportunity to get familiar with the code of cppfront. The switch in `starts_with_operator` had the comments for the operators but the cases statements where missing.

With the missing cases added, the test file works now with cppfront.

I also checked all operators against the list in https://en.cppreference.com/w/cpp/language/operators
1. Additional missing operators: `&=`, `|=`
2. Operators that are checked but not defined: `&&=`, `||=`
3. Other missing operators: `,`, `->*`, `()`, `[]`

I added the operators in point 1. Should I remove the operators under point 2? Should I add the operators under point 3?

On a side note: Thanks for working on cpp2 and providing it to the public!

### DIFF
--- a/source/io.h
+++ b/source/io.h
@@ -221,6 +221,10 @@ auto starts_with_operator(std::string_view s)
         break;case '/':
               case '=':
               case '!':
+              case '*':
+              case '%':
+              case '^':
+              case '~':
             if (c2 == '=') { return j+2; }
             return j+1;
 
@@ -239,7 +243,7 @@ auto starts_with_operator(std::string_view s)
         break;case '|':
               case '&':
             if (c2 == c1 && c3 == '=') { return j+3; }
-            if (c2 == c1) { return j+2; }
+            if (c2 == c1 || c2 == '=') { return j+2; }
             return j+1;
 
             //  >>= >> >= >


JohelEGP:
> I also checked all operators against the list in https://en.cppreference.com/w/cpp/language/operators
> 
>     1. Additional missing operators: `&=`, `|=`
> 
>     2. Operators that are checked but not defined: `&&=`, `||=`
> 
>     3. Other missing operators: `,`, `->*`, `()`, `[]`

See also <https://github.com/hsutter/cppfront/issues/387#operator>.

JohelEGP:
>     operator*: (v1: double, v2: double) -> _ = v1 * v2;

That declaration works at type scope.

Perhaps the greater issue is that the parser that recognizes a Cpp2 declaration at the top-level
doesn't use the same logic for parsing identifiers as the parser _within_ a Cpp2 declaration.

MaxSagebaum:
Thanks for pointing out https://github.com/hsutter/cppfront/issues/387#operator most of it would be fixed when I also add a patch for point 2 and 3.


> > ```
> > operator*: (v1: double, v2: double) -> _ = v1 * v2;
> > ```
> 
> That declaration works at type scope.
> 
> Perhaps the greater issue is that the parser that recognizes a Cpp2 declaration at the top-level doesn't use the same logic for parsing identifiers as the parser _within_ a Cpp2 declaration.

You are right, that one would not write this operator. My original test had a custom class as arguments. Nevertheless, cppfront identifies it on a global level as a cpp1 line and will copy it to the generated source. The cpp1 compiler will then complain about this line.

JohelEGP:
By type scope, I meant in a type scope, not with UDT arguments.
Or more generally, within a Cpp2 declaration.
This should work, but doesn't: <https://cpp2.godbolt.org/z/z9nEYo6ar>.
```Cpp2
t: type = { }
ns: namespace = {
operator*: (_: t) = { } // Recognized as Cpp2.
}
operator*: (_: t) = { } // Not recognized as Cpp2.
main: () = { }
```
```output
build/_cppfront/main.cpp:19:1: error: 'operator*' does not name a type
   19 | operator*: (_: t) = { } // Not recognized as Cpp2.
      | ^~~~~~~~
```

MaxSagebaum:
Yes, my fix  takes care of 
```
operator*: (_: t) = { } // Not recognized as Cpp2.
```

PR #501, opened 2023-06-08
Fix for downcast to parent.

MaxSagebaum:
Fixes for down and up cast in class hierarchies.


MaxSagebaum:
Removed work in progress status.

hsutter:
Looks good, thanks!

PR #516, opened 2023-06-17
When Clang defines _MSC_VER, prefer Clang versions of attribute macros

jbatez:
WIthout this, compiling with Clang for Windows results in a lot of `warning: unknown attribute 'forceinline' ignored [-Wunknown-attributes]` warnings.


hsutter:
Thanks, Jo!

Issue #475, opened 2023-05-30
[BUG] move-assignment operators defaulting to memberwise semantics makes it impossible to cleanup old value


jbatez:
Consider a smart pointer type for [`IUnknown`](https://learn.microsoft.com/en-us/windows/win32/api/unknwn/nn-unknwn-iunknown) that automatically calls [`IUnknown::Release`](https://learn.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iunknown-release) where appropriate:

```cpp2
ComPtr: type = {
    m_raw: *IUnknown;

    // ...

    operator=: (inout this, move that) = {
        // Take the raw pointer from `that` first.
        // If `&that == &this`, setting it to nullptr
        // first ensures we don't release it.
        new_raw := std::exchange(that.m_raw, nullptr);
        if m_raw { m_raw*.Release(); }
        m_raw = new_raw;
    }

    // ...
}
```

cppfront generates the following code for the move-assignment operator:

```cpp
    auto ComPtr::operator=(ComPtr&& that) noexcept -> ComPtr& {
        m_raw = std::move(that).m_raw;
        // Take the raw pointer from `that` first.
        // If `&that == &this`, setting it to nullptr
        // first ensures we don't release it.
        auto new_raw {std::exchange(std::move(that).m_raw, nullptr)}; 
        if (m_raw) {CPP2_UFCS_0(Release, (*cpp2::assert_not_null(m_raw))); }
        m_raw = std::move(new_raw);
        return *this;
    }
```

cppfront inserted `m_raw = std::move(that).m_raw;` before my code, meaning there's no way for me to call `Release()` on the old value.

jbatez:
I wonder if the general structure of generated move-assignment operators should be something more like:

```cpp
auto Foo::operator=(Foo&& that) noexcept -> Foo& {
    auto tmp {std::move(that)};
    this->~Foo();
    member1 = std::move(tmp.member1);
    member2 = std::move(tmp.member2);
    // ...
    // insert given logic here
    // ...
    return *this;
```

jbatez:
I suppose it is possible. But definitely unintuitive, and could be awkward if you're dealing with multiple member variables. I'd argue that doing something destructor-ish with the old value is the norm for custom move-assignment operators and you shouldn't have to jump through IIFE hoops to do it.

This isn't the only case where you want to run code before memberwise assignment. Argument validation is another big one, and that'd be useful for constructors as well as assignment operators.

[...]

PR #518, opened 2023-06-18
Statements before and between member initializers in special member functions

jbatez:
This is a proposed solution for Issue https://github.com/hsutter/cppfront/issues/475

Instead of allowing only `out` parameter assignments before and between member initializers in special member functions, this change allows _all_ statements. When generating constructors, it uses IIFE's and the comma operator to insert statements in the member initializer list. And when generating assignment operators, it does exactly what you'd expect.

Since statements can now go before or after member initializers, it leads to the question: where do statements go in a function with only implicit member initializers? I came up with the following rules:

1. By default, statements are inserted as late as possible while still maintaining the given order relative to explicit member initializers. This means a function without any explicit member initializers will have its statements inserted after all the implicit initializers.
2. To insert statements before the implicit initializers, add a special `this = that` statement after the leading statements. Note: this only works for functions that include `that` as a parameter. If we want something similiar for functions without `that`, we could add a `this = {}` or `this = default` syntax, but I have a harder time imagining real-world use-cases for that.

This change also makes implicit `this.member = that.member` initialization opt-in. Without an explicit `this = that` statement in the body, `that` functions default to the same default initialization behavior as all the other special member functions. I personally find this more consistent with the rest of the language and less surprising, but I'm open to other ideas. I've updated the `copyable` metatype accordingly.

I've also gotten rid of implicit member assignment in explicit assignment (i.e. `inout this`) operators. The way I see it, if you're writing custom assignment operators, it's because there's something special about your type that can't be described by defaults. Implicit assignment operators (i.e. the ones that get generated alongside constructors when you write `out this`) still assign to every member to match the corresponding constructor, and you can still use `this = that` to opt-in to implicit memberwise copy or move.

This change is a work-in-progress. I'm still working on cleaning up the existing regression tests and adding new tests and will likely find a few more bugs or edge cases I hadn't considered. And if you accept this approach, I need to update the documentation.

One obvious design hole is: what should we do about declarations between initializers? They can't escape the IIFE's in the cpp1 initializer lists, but they're definitely useful in assignment operators (see the example in the linked issue).

Edit: `return` statements before and between member initialzers also raise some interesting questions.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -4749,28 +4749,23 @@ public:
 
                     //  If there are any pending statements, wedge them into
                     //  this initializer using an immediately invoked lambda
+                    auto& mem_inits = current_functions.back().prolog.mem_inits;
                     if (!pending_statements.empty())
                     {
                         auto object_type = print_to_string( *(*object)->get_object_type() );
-                        auto mem_init = separator + object_name + "{ [&]() -> " + object_type + " { ";
+                        mem_inits.push_back(separator + object_name + "{ [&]() -> " + object_type + " {");
                         for (auto& stmt : pending_statements) {
-                            mem_init += stmt + "; ";
+                            mem_inits.push_back("      " + stmt + ";");
                         }
                         pending_statements.clear();
-                        mem_init += "return " + initializer + "; }() }";
-                        current_functions.back().prolog.mem_inits.push_back(std::move(mem_init));
+                        mem_inits.push_back("      return " + initializer + ";");
+                        mem_inits.push_back("  }() }");
                     }
                     else {
                         if (initializer == "{}") {
                             initializer = "";
                         }
-                        current_functions.back().prolog.mem_inits.push_back(
-                            separator +
-                            object_name +
-                            "{ " +
-                            initializer +
-                            " }"
-                        );
+                        mem_inits.push_back(separator + object_name + "{ " + initializer + " }");
                     }
                     separator = ", ";
                 }
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -4700,7 +4700,7 @@ public:
 
                         //  Flush any pending statements
                         for (auto& stmt : pending_statements) {
-                            current_functions.back().prolog.statements.push_back(stmt);
+                            current_functions.back().prolog.statements.push_back(stmt + ";");
                         }
                         pending_statements.clear();
 
@@ -4754,7 +4754,7 @@ public:
                         auto object_type = print_to_string( *(*object)->get_object_type() );
                         auto mem_init = separator + object_name + "{ [&]() -> " + object_type + " { ";
                         for (auto& stmt : pending_statements) {
-                            mem_init += stmt;
+                            mem_init += stmt + "; ";
                         }
                         pending_statements.clear();
                         mem_init += "return " + initializer + "; }() }";
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -4757,8 +4757,7 @@ public:
                     auto& mem_inits = current_functions.back().prolog.mem_inits;
                     if (!pending_statements.empty())
                     {
-                        auto object_type = print_to_string( *(*object)->get_object_type() );
-                        mem_inits.push_back(separator + object_name + "{ [&]() -> " + object_type + " {");
+                        mem_inits.push_back(separator + object_name + "{ [&]() -> decltype(auto) {");
                         for (auto statement : pending_statements) {
                             mem_inits.push_back(
                                 "      " + print_to_string(*statement, false) + ";"
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -4497,8 +4497,8 @@ public:
             //      - deferred_statements is the list of statements to insert before the next explicit member initializer
             auto objects             = n.parent_declaration->get_type_scope_declarations(n.objects);
             auto body_statements     = n.get_initializer_statements();
-            auto pending_statements  = std::vector<std::string>{};
-            auto deferred_statements = std::vector<std::string>{};
+            auto pending_statements  = std::vector<statement_node*>{};
+            auto deferred_statements = std::vector<statement_node*>{};
 
             auto object     = objects.begin();
             auto statement  = body_statements.begin();
@@ -4623,8 +4623,7 @@ public:
                     if (!found_explicit_init)
                     {
                         // Queue it up to insert before the next explicit initializer
-                        deferred_statements.push_back( print_to_string(**statement, false) );
-                        (*statement)->emitted = true;
+                        deferred_statements.push_back( *statement );
                         ++statement;
                         continue;
                     }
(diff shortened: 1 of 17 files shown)


jbatez:
I just realized I don't need to use the comma operator, I can just return the initializer from the lambda. That way, variables declared before a member initializer can be used in that member initializer. E.g.:

````
t: type = {
    m: i32;
    operator=: (out this, a: i32) = {
        x: i32 = a;
        m = x;
    }
}
````

generates:

```
    t::t(cpp2::in<cpp2::i16> a)
        : m{ [&]() -> cpp2::i16 { cpp2::i16 x {a}; return std::move(x); }() }
    {
    }

    auto t::operator=(cpp2::in<cpp2::i16> a) -> t& {
        cpp2::i16 x {a}; 
        m = std::move(x);
        return *this;
    }
```

**Comment by JohelEGP** [comment]
> One obvious design hole is: what should we do about declarations between initializers? They can't escape the IIFE's in the cpp1 initializer lists, but they're definitely useful in assignment operators (see the example in the linked issue).

My understanding it that permitting that just comes naturally to Cpp2.
But what Cppfront can do is limited, given that it has to lower to Cpp1.

What if you delegated to a generated constructor to tap into the power of the stack?
Have this:
```Cpp2
t: type = {
  public x: i32;
  operator=: (out this, y: i32) = {
    z := y;
    x = z;
  }
}
```
Generate this:
```C++
class t {
  public: cpp2::i32 x;
  struct cpp2_stack_3_2 { };
  public: explicit t(cpp2::in<cpp2::i32> y);
  private: explicit t(cpp2::in<cpp2::i32> y, cpp2_stack_3_2, auto);
  public: auto operator=(cpp2::in<cpp2::i32> y) -> t& ;

  public: t(t const&) = delete; /* No 'that' constructor, suppress copy */
  public: auto operator=(t const&) -> void = delete;
};

  t::t(cpp2::in<cpp2::i32> y)
                                    : t{ y, cpp2_stack_3_2{}, [] {
    struct cpp2_stack_t {
      cpp2::deferred_init<cpp2::i32> z;
    };
    return cpp2_stack_t{};
                                    }() }
  {}
  t::t(cpp2::in<cpp2::i32> y, t::cpp2_stack_3_2, auto cpp2_stack)
                                    : x{ (cpp2_stack.z.construct(y), cpp2_stack.z.value()) }
  {}
  auto t::operator=(cpp2::in<cpp2::i32> y) -> t& {
                                    x = y;
                                    return *this;
  }
```
See <https://cpp2.godbolt.org/z/EGb3jY8zr>.

jbatez:
That's a clever trick. It gets complicated pretty quickly, though; you have to track down all the references to `z` and replace with `cpp2_stack.z.value()`. E.g.

```
t: type = {
  public x: i32;
  operator=: (out this, y: i32) = {
    z := y;
    std::cout << z << std::endl;
    w := z;
    x = w;
  }
}
```

Generate:

```
class t {
  public: cpp2::i32 x;
  private: struct cpp2_stack_3_2 { };
  public: explicit t(cpp2::in<cpp2::i32> y);
  private: explicit t(cpp2::in<cpp2::i32> y, cpp2_stack_3_2, auto);
  public: auto operator=(cpp2::in<cpp2::i32> y) -> t& ;

  public: t(t const&) = delete; /* No 'that' constructor, suppress copy */
  public: auto operator=(t const&) -> void = delete;
};

  t::t(cpp2::in<cpp2::i32> y)
                                    : t{ y, cpp2_stack_3_2{}, [] {
    struct cpp2_stack_t {
      cpp2::deferred_init<decltype(auto{y})> z;
      cpp2::deferred_init<decltype(auto{z.value()})> w;
    };
    return cpp2_stack_t{};
                                    }() }
  {}
  t::t(cpp2::in<cpp2::i32> y, t::cpp2_stack_3_2, auto cpp2_stack)
                                    : x{ [&]() -> cpp2::i32 {
                                          cpp2_stack.z.construct(y);
                                          std::cout << cpp2_stack.z.value() << std::endl;
                                          cpp2_stack.w.construct(cpp2_stack.z.value());
                                          return cpp2_stack.w.value();
                                      }() }
  {}
  auto t::operator=(cpp2::in<cpp2::i32> y) -> t& {
                                    auto z{y};
                                    x = z;
                                    return *this;
  }
```
See https://cpp2.godbolt.org/z/89EsnhMxM.

I'll give it a try, but it might be more work than I bargained for at this time.

source/cppfront.cpp
```diff
@@ -4537,67 +4548,111 @@ class cppfront
...
+                                    stmt_pos,
+                                    "'this = that' may only appear once"
+                                );
+                                return;
+                            }
+                            else
+                            {
+                                // All the statements before 'this = that' need to be inserted
+                                // before the first member initializer whether or not it's explicit
+                                pending_statements = deferred_statements;
```
gregmarr, line 4585: This is still a good optimization to prevent allocation:

    pending_statements = std::move(deferred_statements);

You can also do it like this:

    std::swap(pending_statements, deferred_statements);
    deferred_statements.clear();

source/cppfront.cpp
```diff
@@ -4633,11 +4633,9 @@ class cppfront
                 // If explicit, upgrade deferred statements to pending
                 if (found_explicit_init)
                 {
-                    pending_statements.insert(
-                        pending_statements.end(),
-                        deferred_statements.begin(),
-                        deferred_statements.end()
-                    );
```
gregmarr: You can make these moves and keep the `insert`.

                    pending_statements.insert(
                        pending_statements.end(),
                        std::make_move_iterator(deferred_statements.begin()),
                        std::make_move_iterator(deferred_statements.end());

jbatez:
I think I'm done with this for now. I didn't do the work to handle things declared before a member initializer and used after. The more I thought about it, the more complicated it got. I think it'd be awesome if the feature was there, but I unfortunately don't have the time to implement it myself.

source/cppfront.cpp
```diff
@@ -4684,40 +4752,28 @@ class cppfront
...
-                        auto ret = std::string{};
-                        for (auto& init : out_inits) {
-                            ret += init + ", ";
+                    //  If there are any pending statements, wedge them into
+                    //  this initializer using an immediately invoked lambda
+                    auto& mem_inits = current_functions.back().prolog.mem_inits;
+                    if (!pending_statements.empty())
+                    {
+                        auto object_type = print_to_string( *(*object)->get_object_type() );
+                        mem_inits.push_back(separator + object_name + "{ [&]() -> " + object_type + " {");
```
JohelEGP, line 4761: Fixing the return type to the type of the data member to initialize is NG.
This should achieve the semantics one would expected reading the Cpp2 code:
```suggestion
                        mem_inits.push_back(separator + object_name + "{ [&]() -> decltype(auto) {");
```

JohelEGP (changes requested):

jbatez:
The original issue has been addressed by https://github.com/hsutter/cppfront/commit/cdf71bdca64536a005f2491d8c19f1d05a1c62f6

hsutter:
Thanks, Jo!

PR #528, opened 2023-06-25
fix(parse): requires clause on one-liner and add error when used on non-template

### LINKED ISSUE #472: [BUG] Non-empty line after one-liner type with requires clause fails assertion
<a id="title"/>

**[T](#title)itle**: Non-empty line after one-liner type with requires clause fails assertion.

<a id="rep"/>
<a id="cpp2"/>

**[M](#rep)inimal reproducer** (<https://cpp2.godbolt.org/z/bfh5bfTj4>):

```Cpp2
t: type requires true = { }
main: () = { }
```

<a id="cmd"/>
<details><summary>
<b><a href="#cmd">C</a>ommands</b>:
</summary>

```bash
cppfront -clean-cpp1 main.cpp2
```
</details>

<a id="actual"/>

**[A](#actual)ctual result and error**:

<a href="#out">O</a>utput.

```output
cppfront: source/cppfront.cpp:523: void cpp2::positional_printer::align_to(cpp2::source_position): Assertion `psource && 0 <= curr_pos.lineno && curr_pos.lineno < std::ssize(psource->get_lines())+1' failed.
```

filipsajdak:
- Add handling of lack of empty line after oneliner.
- Add parsing error when requires is used on non-template

All regression tests pass.

Closes #472

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -6466,14 +6466,20 @@ private:
         }
 
         //  Next is optionally a requires clause
-        if (curr() == "requires") {
-            if (!n->template_parameters) {
-                error("'requires' only valid for a template");
+        if (curr() == "requires")
+        {
+            if (
+                n->is_type()
+                && !n->template_parameters
+                )
+            {
+                error("'requires' is only valid for a type with a template parameter list");
                 return {};
             }
+
             n->requires_pos = curr().position();
             next();
-            auto e = expression();
+            auto e = logical_or_expression();
             if (!e) {
                 error("'requires' must be followed by an expression");
                 return {};
--- a/source/parse.h
+++ b/source/parse.h
@@ -6467,6 +6467,10 @@ private:
 
         //  Next is optionally a requires clause
         if (curr() == "requires") {
+            if (!n->template_parameters) {
+                error("'requires' only valid for a template");
+                return {};
+            }
             n->requires_pos = curr().position();
             next();
             auto e = expression();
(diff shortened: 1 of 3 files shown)


source/parse.h
```diff
@@ -6467,6 +6467,10 @@ class parser
 
         //  Next is optionally a requires clause
         if (curr() == "requires") {
+            if (!n->template_parameters) {
+                error("'requires' only valid for a template");
+                return {};
+            }
```
JohelEGP, line 6473: It's also valid for templated entities (non-templates within templates).

filipsajdak: Could you provide an example?

JohelEGP:
```Cpp2
t: <T> type = {
  f: () requires std::regular<T> = { }
}
```

hsutter: Also, I think the above won't handle the case of generic parameters that use wildcards instead of an explicit template parameter list? Such as

```
f: (x) requires std::regular<decltype(x)> = { }
```

hsutter: I do think there's value in diagnosing a `requires` on a non-templated type for _type_ declarations.

Rationale:
- writing a stray `requires` on a function already gives a nice Cpp1 diagnostic (e.g., MSVC says `error C7599: 't::f': a trailing requires clause is only allowed on a templated function`)
- writing a stray `requires` on a type currently lowers it to a naked `requires` at global scope before the type which leads to a harder-to-understand Cpp1 diagnostic (e.g., MSVC says `error C2059: syntax error: 'requires'`, which is correct and indicates the rough issue, but it's as specific)

So diagnosing the type case in Cpp2 delivers the primary diagnostic usability improvement, plus I think it's easy to code, I think it just needs a `n->is_type()` to the condition Filip already wrote

I'll modify the PR to try that out...

hsutter:
Thanks!

PR #556, opened 2023-07-22
Fix tests and make them platform-independent

jarzec:
- Update tests that finish through `std::terminate`.
  - Add `std::set_terminate(std::abort);` to avoid platform-dependent output on termination.
- Move `regression-tests/test-results/apple-clang-14/` to `regression-tests/test-results/clang-14/`.
  - After the mentioned update the results are the same for clang++-14 on Linux and clang++ (which is v14) on macOS.
  - Update some outdated reference data files for clang-14.
- Remove `regression-tests/test-results/clang-12/` as the results of using clang-12 are identical to those for clang-14.

Results of running the tests on:
- Ubuntu 22.04 with:
  - [g++-12](https://github.com/jarzec/cppfront/actions/runs/5633145364/job/15261740560) (using reference data for gcc-13; v13 is not available by default in Ubuntu 22.04, but the results are exactly the same for v12)
  - [clang++-12](https://github.com/jarzec/cppfront/actions/runs/5633145364/job/15261740608) (using reference data for clang-14)
  - [clang++-14](https://github.com/jarzec/cppfront/actions/runs/5633145364/job/15261740643)
- macOS with
  - [clang++](https://github.com/jarzec/cppfront/actions/runs/5633145364/job/15261740669) (which is v14)

NOTE: One delicate thing about macOS is that some parts of the std library are not implemented (see the detailed compilation errors in the [mentioned run on macOS](https://github.com/jarzec/cppfront/actions/runs/5633145364/job/15261740669)). The test execution bash script I am using to run the test in the mentioned GitHub workflow is detecting compilation errors of the type: `"error: no member named '.*' in namespace 'std"` and simply skips further testing rather than causing a test failure. A Summary of skipped/failed tests is reported at the end.
My rationale here was that C++ implementation shortcomings should not raise cppfront errors.

I will be happy to create a PR with the bash script I used to run the tests or even the GitHub workflow to execute the tests on pushes/PRs if this is deemed useful.


regression-tests/test-results/clang-14/mixed-bounds-check.cpp.execution
```diff
@@ -1,2 +1 @@
 Bounds safety violation: out of bounds access attempt detected
-libc++abi: terminating
```
jarzec, line 2: This is the platform-dependent message (c.f. the one on Ubuntu below) removed thanks to `std::set_terminate(std::abort);`.

regression-tests/test-results/gcc-13/mixed-bounds-check.cpp.execution
```diff
@@ -1,2 +1 @@
 Bounds safety violation: out of bounds access attempt detected
-terminate called without an active exception
```
jarzec, line 2: This is the platform-dependent message (c.f. the one on macOS above) removed thanks to std::set_terminate(std::abort);.

regression-tests/test-results/clang-14/clang-version.output
jarzec, line 1: This is the output of macOS `clang++ --version`. I wasn't sure what to do with it.

regression-tests/test-results/apple-clang-14/mixed-float-literals.cpp.execution
jarzec, line 1: This file was not detected as move because it was outdated. The original is empty, while the current version of the test prints values to avoid `unused` warnings.

regression-tests/test-results/apple-clang-14/mixed-float-literals.cpp.output
jarzec, line 1: As mentioned above, this file was outdated: `unused` warnings were fixed by printing values.

regression-tests/test-results/apple-clang-14/pure2-type-and-namespace-aliases.cpp.output
jarzec, line 1: This file had compilation issues due to shortcomings of std implementation on macOS (`error: no member named 'pmr' in namespace 'std'`).
More tests are currently failing on macOS with `no member named 'for_each' in namespace 'std::ranges'` - see the description of this PR.

hsutter:
Thanks! I think the core of this PR is the three `std::set_terminate` calls, right?

I'd rather not delete all the Clang-12 regression test results and change them to Clang 14... I'd like to keep the checked-in ones on 12 to make sure things continue to run on 12.

Would you please trim the PR to not update the test results? Then after merge I can just re-run the regressions on my system (staying on the compilers I'm using) which will reduce the diffs and not change the checked-in Clang regression version.

Thanks1

jarzec:
Sure, can do. I just noticed that if they are up to date the content of clang-14 and clang-12 directories is identical.
I will bring the apple-clang-14 up to date. I will add the compilation issue results for the 6 tests that fail due to incomplete std implementation (one of those tests already had those failures included in expected test results).
Would that work for you?

hsutter:
I think so -- so just add the `set_terminate` calls, and update the Apple directory? SGTM, thx!

PR #574, opened 2023-08-06
Implement `sizeof<T>`

### LINKED ISSUE #566: [BUG] Can't use `sizeof`
<a id="title"/>

**[T](#title)itle**: Can't use `sizeof`.

<a id="rep"/>
<a id="cpp2"/>

**[M](#rep)inimal reproducer** (<https://cpp2.godbolt.org/z/8bs65Tzcz>): `main: () -> int = sizeof(i32);`

<a id="cmd"/>
<details><summary>
<b><a href="#cmd">C</a>ommands</b>:
</summary>

```bash
cppfront main.cpp2
clang++17 -std=c++23 -stdlib=libc++ -lc++abi -pedantic-errors -Wall -Wextra -Wconversion -fsanitize=undefined -Werror=unused-result -I . main.cpp
```
</details>

<a id="expected"/>

**[E](#expected)xpected result**: `Program returned: 4`.

<a id="actual"/>

**[A](#actual)ctual result and error**: `main.cpp2(1,25): error: use 'sizeof(()', not 'sizeof ('`.

ntrel:
Fix sizeof/new hints with parentheses.
Fixes #566.

### DIFF
--- a/regression-tests/mixed-fixed-type-aliases.cpp2
+++ b/regression-tests/mixed-fixed-type-aliases.cpp2
@@ -1,5 +1,6 @@
 #include <typeinfo>
 #include <iostream>
+#include <cassert>
 
 namespace my {
     using u16 = float;
@@ -13,9 +14,11 @@ test: (x:_) = {
 }
 
 main: (args) -> int = {
+    assert(sizeof<my::u16> == 4);
     y: my::u16 = 42;
     test(y);
 
+    assert(sizeof<u16> == 2);
     z: u16 = 42;
     test(z);
 
--- a/regression-tests/pure2-sizeof-new-error.cpp2
+++ b/regression-tests/pure2-sizeof-new-error.cpp2
@@ -1,0 +1,6 @@
+main: () = {
+    p := new i32;
+    p = new(i32);
+    i := sizeof(i32);
+    i = sizeof i32;
+}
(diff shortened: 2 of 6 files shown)


JohelEGP:
Thank you for doing this!
I wouldn't have thought that `sizeof` was supposed to work with `<>`.
However, it doesn't seem to work (<https://cpp2.godbolt.org/z/K8nPMeaa4>):
```Cpp2
main: () -> int = sizeof<i32>;
```
```C++
[[nodiscard]] auto main() -> int { return sizeof<cpp2::i32>;  }
```
```output
main.cpp2: In function 'int main()':
main.cpp2:1:49: error: expected primary-expression before '<' token
main.cpp2:1:59: error: expected primary-expression before '>' token
main.cpp2:1:60: error: expected primary-expression before ';' token
```

ntrel:
@JohelEGP You're right, I heard it would use `<>` but there are `passthrough-tests` that use it with parentheses. I'll have a look.

ntrel:
@JohelEGP I think it wasn't implemented. `sizeof<i32>` should work now.

hsutter:
Thanks but I think this is fixed with the commit that closed #566.

I see the consistency of using `<` `>` for `sizeof` as well as `new`, but in Cpp2 the type passed to `new` really is a template argument and I didn't see a reason to change from the familiar function-like syntax `sizeof(thing)` to `sizeof<thing>` ...? I was thinking of `sizeof` as a constexpr function for which `(` `)` would be fine.

I'll close this for now as the other commit addressed #566, but feel free to open a new issue if you want to provide rationale for changing `sizeof(thing)` to `sizeof<thing>` (e.g., what problems does it solve, the usual questions).

Thanks!

Issue #262, opened 2023-03-01
[SUGGESTION] Support for a "`main` v2" signature


aslilac:
The usual `argc`/`argv` signature requires familiarity some C idioms that don't really have broad influence on C++. An alternative signature could allow new learners to not be bogged down by the dubious `char**` and `\0` termination, and would instead provide them with a clearer, more familiar data type, that would be less error and abuse-prone. Ideally it would only take a single argument, `args`, which would be a standard list container, holding some number of arguments. Perhaps `main : (args: std::vector<std::string_view>) -> int`.

**Will your feature suggestion eliminate X% of security vulnerabilities of a given kind in current C++ code?** 

I've seen some wild forms of argument parsing out there that abuse `argc`. The percentage would definitely be small, but I'm sure there are people out there misusing `argc` enough to have allowed a memory safety bugs to slip in.

**Will your feature suggestion _automate or eliminate_ X% of current C++ guidance literature?** 

It's largely agreed that passing a length and a C array is bad practice when writing your own functions, but you don't get to dictate the signature of `main`. Requiring `main` to follow an idiom that is otherwise frowned upon (and in a place that every C++ programmer is bound to come across!) raises a lot of questions about *why* things are this way, and why it's a bad idea to use a signature like `main`'s elsewhere in your code. If `main` took something a little more idiomatic, those conversations can be kicked down the road until much later, instead of requiring they be addressed while trying to write a "Hello, world!"

Some references:
- Jason Turner recently did a whole video that basically ended with the advice "you should probably be constructing a `std::vector<std::string_view>` immediately, and then only using that.
- The SerenityOS project has a header file which can be included in any individual program that lets you write a `serenity_main` function with a nicer type signature, instead of writing the usual `main` function.

**Describe alternatives you've considered.**
Some potential types that could be used for `args`

- `std::array<_>`
- `std::vector<_>`
- `_<std::string>`
- `_<std::string_view>`

hsutter:
Thanks, and thanks again to everyone else who has made this suggestion in past issues. This has been on my mind as well.

[...]

PR #600, opened 2023-08-17
Update regression tests for apple-clang-14 (23W33)

filipsajdak:
I am catching up with regression tests for apple clang 14.


regression-tests/test-results/apple-clang-14/mixed-captures-in-expressions-and-postconditions.cpp.output
```diff
@@ -1,7 +0,0 @@
-mixed-captures-in-expressions-and-postconditions.cpp2:12:18: error: no member named 'for_each' in namespace 'std::ranges'
-    std::ranges::for_each(vec, callback);
-    ~~~~~~~~~~~~~^
-mixed-captures-in-expressions-and-postconditions.cpp2:14:18: error: no member named 'for_each' in namespace 'std::ranges'
-    std::ranges::for_each(std::move(vec), std::move(callback));
-    ~~~~~~~~~~~~~^
-2 errors generated.
```
jarzec, line 7: That's cool!
These errors are present on `macos-latest` GitHub runners (see [here](https://github.com/jarzec/cppfront/actions/runs/5913839684/job/16038580118) under "Run regression tests - Linux and macOS version").
I am guessing this is thanks to some sort of update? What kind of update is necessary? Is it just Xcode or something more complicated? The change to `clang-version.output` seems to suggest a full OS version bump but maybe I'm mistaken.

JohelEGP: I guess. See <https://github.com/hsutter/cppfront/issues/262#issuecomment-1512136235>.

hsutter:
Thanks!

PR #655, opened 2023-09-03
Add source/cpp2decl.cpp

nigeltao:
I don't expect you to merge this as is, but a Pull Request seemed the easiest way to share a concrete example of the sort of tools I'd like to have and perhaps discuss how language features (e.g. attributes, modifiers, reflection, etc.) should affect syntax.

The cpp2 language is obviously still changing but also, long term, parsing cpp2 code should probably be factored out as a library (separate from the cppfront transpiler) and this can be another example of a customer for that future library.

### DIFF
--- a/source/cpp2decl.cpp
+++ b/source/cpp2decl.cpp
@@ -1,0 +1,321 @@
+//  Copyright (c) Herb Sutter
+//  SPDX-License-Identifier: CC-BY-NC-ND-4.0
+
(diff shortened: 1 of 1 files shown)


JohelEGP:
That's what this category is for: <https://github.com/hsutter/cppfront/discussions/categories/show-and-tell>.

nigeltao:
Thanks. Moved to https://github.com/hsutter/cppfront/discussions/657

PR #669, opened 2023-09-09
Add `is` overload for free function predicate

filipsajdak:
The current implementation of `is` does not handle generic functions as predicates.

The following case works:
```cpp
less_than_10_a: (i : int) -> bool = i < 10;

main: () = {
  12 is (less_than_10_a);
}
```

It works also when the function has a generic return type:
```cpp
less_than_10_b: (i : int) -> _ = i < 10;

main: () = {
  12 is (less_than_10_b);
}
```

But, it does not work when the function has a generic argument. The following case does not work:
```cpp
less_than_10_c: (i) -> _ = i < 10;

main: () = {
  12 is (less_than_10_c); // error
}
```

This change introduces additional overload for `is`, making the free function predicates acceptable by `is` (the function needs to return `bool` or generic type). That makes the following cases work:
```cpp
less_than_10_c: (i) -> _ = i < 10;
less_than_10_d: (i) -> bool = i < 10;

main: () = {
  5 is (less_than_10_c); // works
  2 is (less_than_10_d); // works
}
```

Functions that return other types than `bool` or `auto` do not work.

All regression tests pass. Add a test for this case and update the regression test results.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -894,6 +894,12 @@ inline constexpr auto is( auto const& x, auto const& value ) -> bool
     return false;
 }
 
+// Free function predicate case
+template< typename X >
+constexpr auto is( X const& x, bool (&value)(X const&) ) -> bool
+{
+    return value(x);
+}
 
 //-------------------------------------------------------------------------------------------------------------
 //  Built-in as
--- a/regression-tests/mixed-inspect-values-2.cpp2
+++ b/regression-tests/mixed-inspect-values-2.cpp2
@@ -16,6 +16,10 @@ constexpr auto empty = [](auto&& x){
     return std::empty(x);
 };
 
+auto negative(auto v) {
+    return v < 0;
+}
+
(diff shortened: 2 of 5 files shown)


include/cpp2util.h
```diff
@@ -882,6 +882,12 @@ inline constexpr auto is( auto const& x, auto const& value ) -> bool
     return false;
 }
 
+// Free function predicate case
+template< typename X >
+inline constexpr auto is( X const& x, bool (&value)(X const&) ) -> bool
```
JohelEGP, line 887:
```suggestion
constexpr auto is( X const& x, bool (&value)(X const&) ) -> bool
```
A template function is implicitly `inline`.

filipsajdak: Oh, yes. I forgot to remove it. Thank you!

filipsajdak:
I am closing this one as it is covered by #701

PR #691, opened 2023-09-21
Fix cppfront incomplete type build error (fix #690)

### LINKED ISSUE #690: [BUG] Error when building cppfront with Apple clang 15
**Describe the bug**
Error when building `cppfront` due to a static declaration within a class that uses the full type of the class itself.

```
source/parse.h:1029:50: note: in instantiation of member function 'std::vector<cpp2::template_argument>::~vector' requested here
    static inline std::vector<template_argument> no_template_args;
                                                 ^
source/parse.h:1015:8: note: definition of 'cpp2::template_argument' is not complete until the closing '}'
struct template_argument
```

The error shows compiling `cppfront` on a MacOS, older compilers are less strict and this might be the reason why the error has crept in. It still appears to be a true problem.

**To Reproduce**
On a MacOS jus build `cppfront`
```
g++ source/cppfront.cpp -std=c++20 -o cppfront
```
**Additional context**
The version of the compiler is the following.
```
> g++ --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: x86_64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

fedapo:
I'm submitting an effort to fix the problem.
While it should work, the function `template_arguments()` returns a non-const reference to a static object that should stay empty, yet it could potentially be populated by the caller of the function. I'm not entirely ok with this, but I haven't dug any deeper, could the reference be changed to point to a const object?

UPDATE:
Upon @hsutter suggestion I changed the empty vector to be a const global. The `template_arguments()` functions now return a const ref.
Again, this was my attempt to fix this issue. Should a better implementation be provided, I'm ok with that.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -1024,9 +1024,6 @@ struct template_argument
 
     auto to_string() const
         -> std::string;
-
-    //  So that template_arguments() accessors can return a reference to an empty arg list
-    static inline std::vector<template_argument> no_template_args;
 };
 
 
@@ -3643,11 +3640,14 @@ auto function_type_node::is_destructor() const
 auto primary_expression_node::template_arguments() const
     -> std::vector<template_argument>&
 {
+    // So that the function can return a reference to an empty arg list
+    static std::vector<template_argument> no_template_args;
+
     if (expr.index() == id_expression) {
         return std::get<id_expression>(expr)->template_arguments();
     }
     // else
-    return template_argument::no_template_args;
+    return no_template_args;
 }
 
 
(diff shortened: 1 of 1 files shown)


JohelEGP:
Maybe `no_template_args` should be a member of `primary_expression_node` instead.

hsutter:
I was about to make it a global static, and fix the accessors to return `const&` (the missing `const` was just an oversight).

fedapo:
@hsutter Yes, I took a second look and I was about to try an change all accessors. I'm ok with closing this ticket and let you do it or continue with your suggestion. Your call.
Thanks for the feedback.

PR #693, opened 2023-09-21
Extend std C++ include header list (fix #692)

### LINKED ISSUE #692: Incomplete list of standard lib C++ headers in non-module case
File `cpp2util.h` has the list of headers of the C++ standard library to implement the Cpp2 feature of an always-on standard library. This list is repeated in a few branches of a pre-processor if-else statement to allow pre and post C++23  build scenarios. 

The non-module, pre-C++23 list is incomplete.

I suggest to make this list as complete as the other case and, possibly, sort the headers alphabetically of quicker checking whether an item is missing.
I'm aware that this list is temporary and the steady-state solution is a simple `import std;`, still, it'd be nice to be able to run some tests with `cppfront` with today's compilers for a larger class of cases.

fedapo:
The new list is taken from the CPP2_USE_MODULES case and sorted alphabetically. This is a bit repetitive and could have been avoided, but I chose not to mess with the preprocessor if-else logic. I can change it, should several voices prefer that.


hsutter:
Thanks! To make it easier for me to track what's being added, I went ahead and sorted the list myself (I think? it's been a long Sunday, please also fix any ordering oversights). So if you could please update the PR then I can see what headers are being added in the diff.

fedapo:
@hsutter Sure. Done.

include/cpp2util.h
```diff
@@ -197,39 +197,128 @@
 //  Otherwise, we're not in -pure-cpp2 and so just #include
 //  what we need in this header to make this self-contained
```
JohelEGP, line 198: Now this comment isn't true anymore.

hsutter:
Ah, now I see what you're doing... I think you're suggesting that even non-`pure`/modules mode should just include all standard headers always? I was slightly worried about possible compile time impact, but then I suppose we're already doing that in `-pure` mode (when modules aren't available, which is most of the time right now in practice).

hsutter:
Thanks @fedapo !

The original reason to provide the entire standard library in `-pure-cpp2` mode was that module is import is _blazing fast_ -- it's faster to use C++23 `import std;` than to compile iostream Hello World. So it made sense to try the experiment of just making it the default.

The reason I provided a way to include all std headers was as a fallback path for compilers that didn't have real module import yet (most of them) or C++23 `import std;` (essentially all of them).

I've tried out the suggestion here, and I'm seeing that `#include` of the entire standard library makes compiling small files (including the regression tests) noticeably sluggish -- not a whole lot, but noticeable.

However, I see value in providing a way to include the entire standard library even in non-pure-cpp2 mode, so I'm going to try out providing it as command line options. The idea is to make it "zero-overhead," i.e., let users can pay for the additional compilation work only if they ask for it:

```
  -im[port-std]           import the entire C++ standard library via 'import std;'
  -in[clude-std]          #include the entire C++ standard library - ignored if -import-std is set
  -p[ure-cpp2]            Allow Cpp2 syntax only - also sets -import-std
```

So now a mixed-Cpp1-and-Cpp2 file can easily opt into the convenience of _either_ importing (fast!) or including (wait a couple of extra seconds) the entire standard library. Note `-pure-cpp2` retains its current behavior because it implies `-import-std`.

I'll push that commit instead of this PR, but it was directly inspired by this. Again, thanks @fedapo !

fedapo:
@hsutter Thanks for the notice.
Fine with me to get the full set of std imports through an _ad hoc_ command-line switch.

Issue #575, opened 2023-08-09
[BUG] Non-type template parameter used as _type-id_ is not diagnosed


JohelEGP:
<a id="title"/>

**[T](#title)itle**: Non-type template parameter used as _type-id_ is not diagnosed.

<a id="desc"/>

**[D](#desc)escription**:

I encountered this when poking Cppfront to get it to emit a _type-constraint_.
The issue is alleviated by resolving #425,
as it would only happen with an NTTP by `copy`.

<a id="rep"/>
<a id="cpp2"/>

**[M](#rep)inimal reproducer** (<https://cpp2.godbolt.org/z/WehjhjPjY>):

```Cpp2
t: @struct <T: std::regular> type = {
  v: T;
}
f: <T: std::regular> (v: T) = _ = v;
g: <T: std::regular> (forward v: T) = _ = v;
h: <T: std::regular> (move v: T) = _ = v;

main: () = {
  _ = :t = (0); // Not OK (works, should error)
  f(0); // error (by chance?)
  g(0); // error (by chance?)
  h(0); // Not OK (works, should error)
}
```

<a id="cmd"/>
<details><summary>
<b><a href="#cmd">C</a>ommands</b>:
</summary>

```bash
cppfront main.cpp2
clang++17 -std=c++23 -stdlib=libc++ -lc++abi -pedantic-errors -Wall -Wextra -Wconversion -fsanitize=undefined -Werror=unused-result -I . main.cpp
```
</details>

<a id="expected"/>

**[E](#expected)xpected result**: A diagnostic when using an NTTP as a type.

<a id="actual"/>

**[A](#actual)ctual result and error**:

<a id="cpp1"/>
<details><summary>
<a href="#cpp1">C</a>pp2 lowered to Cpp1:
</summary>

```C++


//=== Cpp2 type declarations ====================================================


#include "cpp2util.h"

template<std::regular T> class t;
  

//=== Cpp2 type definitions and function declarations ===========================

template<std::regular T> class t {
  public: T v; 
};
template<std::regular T> auto f(cpp2::in<T> v) -> void;
template<std::regular T> auto g(auto&& v) -> void
CPP2_REQUIRES (std::is_same_v<CPP2_TYPEOF(v), T>)
;
template<std::regular T> auto h(T&& v) -> void;

auto main() -> int;
  

//=== Cpp2 function definitions =================================================


template<std::regular T> auto f(cpp2::in<T> v) -> void { (void) v;  }
[...]

Issue #492, opened 2023-06-04
[BUG] `operator is` should be constrained or answered statically


JohelEGP:
<a id="title"/>

**[T](#title)itle**: `operator is` should be constrained or answered statically.

<a id="desc"/>

**[D](#desc)escription**:

- Constrained: `operator is` should be non-viable given a type-unsafe input.
- Answered statically: See <https://quuxplusone.github.io/blog/2022/01/04/test-constexpr-friendliness/>.

This is to implement [P2392][]'s §4.2.
It's also to correctly resolve #433,
as split off from #491:

<details><summary>
Comments moved from #491.
</summary>

---

The case of this PR is:
> The alternative's [_is-as-expression-target_][] matches
> but the statement is ill-formed for the inspected type.

[P2392][] doesn't say what should happen in this case.
If one extracted the wording from the design, it would be UB by omission

IIUC what [P1371][] says, then it would similarly be UB by omission:
> When inspect is executed, its condition is evaluated and matched in order (first match semantics) against each
> pattern. If a pattern successfully matches the value of the condition and the boolean expression in the guard
> evaluates to true (or if there is no guard at all), then the value of the resulting expression is yielded or control is
> passed to the compound statement, depending on whether the inspect yields a value. If the guard expression
> evaluates to false, control flows to the subsequent pattern.
>
> If no pattern matches, none of the expressions or compound statements specified are executed. In that case if
> the inspect expression yields void, control is passed to the next statement. If the inspect expression does not
> yield void, std::terminate will be called.

It also links to
<https://github.com/solodon4/Mach7>,
<https://github.com/mpark/patterns>, and
<https://github.com/jbandela/simple_match>.
But this corner case doesn't seem to be mentioned in their README.

[_is-as-expression-target_]: https://github.com/hsutter/cppfront/issues/387#is-as-expression-target
[P2392]: https://wg21.link/p2392
[P1371]: https://wg21.link/p1371

_Originally posted by @JohelEGP in https://github.com/hsutter/cppfront/issues/491#issuecomment-1575125192_

---

How about this instead?

For a dependent `inspect`,
when the [_is-as-expression-target_][] of an alternative is well-formed,
the corresponding statement should also be well-formed.

So we move
from a runtime-checked contract
to a compile-time assertion.

[_is-as-expression-target_]: https://github.com/hsutter/cppfront/issues/387#is-as-expression-target

[...]

PR #701, opened 2023-09-24
Rework of `is` that adds new functionalities or simplify implementation

### LINKED ISSUE #689: [BUG] `e is func` is false
**Describe the bug**
With cppfront from git ecd37263f9b5a71f5beb18affc275d76ee537f9f and g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0:
```c++
pred: (x: int) -> int = x < 20;

main: () = 
{
    p2 := :(x) -> _ = x < 20;
    b := 5 is (p2); // true
    std::cout << b << "\n";
    b = 5 is (pred); // false
    std::cout << b << "\n";
}
```
I expected both `is` expressions to produce `true`, but the second one is `false`.

A secondary problem is that rewriting `pred` to be a template fails to compile:
```c++
pred: (x) -> _ = x < 20;
```
```
lessthan.cpp2:48:17: error: no matching function for call to ‘is(int, <unresolved overloaded function type>)’
... snip candidates ...
```

filipsajdak:
This change reworks a big part of the code for `is`.

* added concepts for simplify and utilize concept subsumption rules,
  * inspired by @JohelEGP - I have learned more here: https://andreasfertig.blog/2020/09/cpp20-concepts-subsumption-rules/
* Added `type_find_if` to iterate over `variant` types to simplify `is` for `variant`,
* rewrite `is` overloads to use `concepts` instead of `type_traits` (to utilize concepts subsumption rules),
* remove `operator_is`, implementation of `is`  for variants now use `type_find_if`,
* `is` that is not using argument `x` for inspection returns `std::true_type` or `std::false_type`
* add many atomic concepts that were used to build complex concepts - that are required for concept subsumption rules,

Closes #689
Closes #669 

Cases handled by the new `is()`:
* type is type,
* type is template,
* type is type_trait,
* type is concept, (no cpp2 syntax)
* variable is template,
* variable is type, (also works for variant, any, optional)
* variable is value, (also works for variant, any, optional)
* variable is empty, (also works for variant, any, optional)
* variable is type_trait,
* variable is predicate, (works also with free function and generic functions; forbids implicit cast of arguments)

# Tests

* All tests pass,
* Added tests that present all use cases handled by this change

# Issue

* does not compile on `macos-11, clang++, c++20` (one of the CI targets)

The original PR contained a rework of `as()` and `to_string()` - they will be provided in separate PRs.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1901,39 +1901,32 @@ constexpr auto as( X const& x ) -> T
 //  std::optional is and as
 //
 
-//  is Type
+//  std::optional variable is Type
 //
-template<typename T, typename X>
-    requires std::is_same_v<X,std::optional<T>>
-constexpr auto is( X const& x ) -> bool
-    { return x.has_value(); }
-
-template<typename T, typename U>
-    requires std::is_same_v<T,empty>
-constexpr auto is( std::optional<U> const& x ) -> bool
-    { return !x.has_value(); }
-
-
-//  is Value
-//
-template<typename T>
-constexpr auto is( std::optional<T> const& x, auto&& value ) -> bool
-{
-    //  Predicate case
-    if constexpr (requires{ bool{ value(x) }; }) {
-        return value(x);
-    }
-    else if constexpr (std::is_function_v<decltype(value)> || requires{ &value.operator(); }) {
-        return false;
-    }
-
-    //  Value case
-    else if constexpr (requires{ bool{ x.value() == value }; }) {
-        return x.has_value() && x.value() == value;
-    }
-    return false;
+template<not_same_as<empty> T, specialization_of_template<std::optional> U>
+    requires not_same_as<T, U> && not_same_as<T, pointee_t<U>>
+constexpr auto is( U&& ) -> std::false_type {
+    return {};
 }
 
+template<not_same_as<empty> T, specialization_of_template<std::optional> U>
+    requires not_same_as<T, U>
+constexpr auto is( U&& x ) {
+    return x.has_value();
+}
+
+//-------------------------------------------------------------------------------------------------------------
+//  pointer_like variable is value
+//
+
+template <pointer_like X, typename V>
+constexpr auto is( X const& x, V && value) -> decltype(auto) {
+    if constexpr (pointer<X> && pointer<V>) {
+        return x == value;
+    } else {
+        return !is<empty>(x) && is(*x, CPP2_FORWARD(value));
+    }
+}
 
 //  as
 //
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1856,40 +1856,38 @@ auto as( std::variant<Ts...> const& x ) -> decltype(auto) {
 //  std::any is and as
 //
 
-//  is Type
+//  std::any variable is Type
 //
-template<typename T, typename X>
-    requires (std::is_same_v<X,std::any> && !std::is_same_v<T,std::any> && !std::is_same_v<T,empty>)
-constexpr auto is( X const& x ) -> bool
+template<typename T>
+constexpr auto is( same_type_as<std::any> auto && x ) -> bool
     { return x.type() == Typeid<T>(); }
 
-template<typename T, typename X>
-    requires (std::is_same_v<X,std::any> && std::is_same_v<T,empty>)
-constexpr auto is( X const& x ) -> bool
+template<std::same_as<empty> T>
+constexpr auto is( same_type_as<std::any> auto && x ) -> bool
     { return !x.has_value(); }
 
 
-//  is Value
+//  std::any variable is Value
 //
-inline constexpr auto is( std::any const& x, auto&& value ) -> bool
-{
-    //  Predicate case
-    if constexpr (requires{ bool{ value(x) }; }) {
-        return value(x);
-    }
-    else if constexpr (std::is_function_v<decltype(value)> || requires{ &value.operator(); }) {
-        return false;
-    }
 
-    //  Value case
-    else if constexpr (requires{ bool{ *std::any_cast<CPP2_TYPEOF(value)>(&x) == value }; }) {
-        auto pvalue = std::any_cast<CPP2_TYPEOF(value)>(&x);
-        return pvalue && *pvalue == value;
-    }
-    //  else
-    return false;
+template <std::same_as<std::any> X, has_defined_argument V>
+    requires not_same_as<std::any, argument_of_t<V>>
+constexpr bool is( X const& x, V && value ) {
+    auto* ptr = std::any_cast<argument_of_t<V>>(&x);
+    return ptr && value(*ptr);
 }
 
+template <std::same_as<std::any> X, std::equality_comparable V>
+    requires (!has_defined_argument<V>)
+constexpr bool is( X const& x, V && value ) {
+    if constexpr (pointer_like<V>) {
+        auto* ptr = std::any_cast<pointee_t<V>>(&x);
+        return ptr && !is<empty>(value) && is(*ptr, *value);
+    } else {
+        auto* ptr = std::any_cast<std::remove_cvref_t<V>>(&x);
+        return ptr && is(*ptr, std::forward<V>(value));
+    }
+}
 
 //  as
 //
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -379,6 +379,11 @@ concept specialization_of_template_type_and_nttp = requires (X x) {
 template <template <typename...> class C>
 concept type_trait = std::derived_from<C<int>, std::true_type> || std::derived_from<C<int>, std::false_type>;
 
+template<typename X>
+concept boolean_testable = std::convertible_to<X, bool> && requires(X&& x) {
+  { !std::forward<X>(x) } -> std::convertible_to<bool>;
+};
+
 template <typename X>
 concept polymorphic = std::is_polymorphic_v<std::remove_cvref_t<X>>;
 
@@ -484,6 +489,66 @@ struct aligned_storage {
     alignas(Align) unsigned char data[Len];
 };
 
+//-----------------------------------------------------------------------
+//
+//  A type_find_if for iterating over types in parameter packs
+//
+//  Note: the current implementation is a workaround for clang-12 internal error.
+//  Original implementation does not need type_it and is implemented
+//  using lambda with explicit parameter type list in the following way:
+//
+//    template <typename... Ts, typename F>
+//    constexpr auto type_find_if(F&& fun)
+//    {
+//        std::size_t found = std::variant_npos;
+//        [&]<std::size_t... Is>(std::index_sequence<Is...>){
+//            if constexpr ((requires { {CPP2_FORWARD(fun).template operator()<Is, Ts>()} -> std::convertible_to<bool>;} && ...)) {
+//                (((CPP2_FORWARD(fun).template operator()<Is, Ts>()) && (found = Is, true)) || ...);
+//            }
+//        }(std::index_sequence_for<Ts...>());
+//        return found;
+//    }
+//
+//  The workaround is not needed in gcc-12.1+, clang-13+, msvc 19.29+
+//
+//  Note2: the internal if constexpr could have else with static_assert.
+//  Unfortunatelly I cannot make it work on MSVC.
+//
+//-----------------------------------------------------------------------
+//
+template <std::size_t Index, typename T>
+struct type_it {
+    using type = T;
+    inline static const std::size_t index = Index;
+};
+
+template <typename... Ts, typename F>
+constexpr auto type_find_if(F&& fun)
+{
+    std::size_t found = std::variant_npos;
+    [&]<std::size_t... Is>(std::index_sequence<Is...>){
+        if constexpr ((requires { {CPP2_FORWARD(fun)(type_it<Is, Ts>{})} -> boolean_testable;} && ...)) {
+            ((CPP2_FORWARD(fun)(type_it<Is, Ts>{}) && (found = Is, true)) || ...);
+        } 
+    }(std::index_sequence_for<Ts...>());
+    return found;
+}
+
+template <typename F, template<typename...> class C, typename... Ts>
+constexpr auto type_find_if(C<Ts...>, F&& fun)
+{
+    return type_find_if<Ts...>(CPP2_FORWARD(fun));
+}
+
+template <typename T, typename... Ts>
+constexpr auto variant_contains_type(std::variant<Ts...>)
+{
+    if constexpr (is_any<T, Ts...>) {
+        return std::true_type{};
+    } else {
+        return std::false_type{};
+    }
+}
 
 //-----------------------------------------------------------------------
 //
@@ -848,6 +913,26 @@ using in =
         T const&
     >;
 
+
+template<class T, class U>
+[[nodiscard]] constexpr auto&& forward_like(U&& x) noexcept
+{
+    constexpr bool is_adding_const = std::is_const_v<std::remove_reference_t<T>>;
+    if constexpr (std::is_lvalue_reference_v<T&&>)
+    {
+        if constexpr (is_adding_const)
+            return std::as_const(x);
+        else
+            return static_cast<U&>(x);
+    }
+    else
+    {
+        if constexpr (is_adding_const)
+            return std::move(std::as_const(x));
+        else
+            return std::move(x);
+    }
+}
 
 //-----------------------------------------------------------------------
 //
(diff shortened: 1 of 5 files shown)


JohelEGP:
> * add possibility to check if object fulfills concept by using `x is (:<T:std::integral>() = {})` syntax,

See <https://github.com/hsutter/cppfront/issues/575#issuecomment-1676199109>:
> That isn't yet supported, but my intent is to spell it as `<T: type is std::regular>`.

filipsajdak:
> > * add possibility to check if object fulfills concept by using `x is (:<T:std::integral>() = {})` syntax,
> 
> See [#575 (comment)](https://github.com/hsutter/cppfront/issues/575#issuecomment-1676199109):
> 
> > That isn't yet supported, but my intent is to spell it as `<T: type is std::regular>`.

Thank you. @hsutter propose the syntax on cpp2 side. I have prepared the implementation on C++20 side, and the intention is to be able to use it e.g. in `if` or `inspect` (assuming `<concept>` will be syntax for using in such cases - similarly as we are using `(value)` with comparing to values or to predicates):
```cpp
inspect x -> std::string {
  is <std::integral> = "x is integral type";
  is <std::floating_point> = "x is floating point type";
  is _ = "other type";
}
```

JohelEGP:
If `<>` will be used to disambiguate concepts, shouldn't
`x is (:<T:std::integral>() = {})` be
`x is (:<T:<std::integral>>() = {})`?

filipsajdak:
I don't know. I know that on cppfront, we don't know if something is a concept or a type (at least, I did not find a way to distinguish that on the C++ side). e.g. in `x is T`, currently, `T` can be a type or a template - it will be distinguished on the C++ side using overloads `is<T>(x)`... I did not find a way to make it work when T is a concept.

JohelEGP:
Yeah, there's no way.
You can only specialize it to get a prvalue `bool` back.
You can't use it any other way, not even in a _requires-expression_.

filipsajdak:
In the cases you mentioned, I guess that we don't need additional `<>` to distinguish that as type and concept are correct in this context.

JohelEGP:
I think I understand now.
You're relying on an accidental feature.
In that case, `T` is the name of a NTTP.
The lowered syntax happens to work for Cpp1 terse concept syntax.
See my reply to Herb's reply: <https://github.com/hsutter/cppfront/issues/575#issuecomment-1676200359>.

filipsajdak:
I was interested in making C++ code work. Based on that I was able to call it from cpp2.

regression-tests/mixed-as-for-variant-20-types.cpp2
```diff
@@ -1,75 +1,37 @@
...
+}
+
 main: () -> int = {
 
-    v: std::variant<X< 0>, X< 1>, X< 2>, X< 3>, X< 4>, X< 5>, X< 6>, X< 7>, X< 8>, X< 9>, 
-                    X<10>, X<11>, X<12>, X<13>, X<14>, X<15>, X<16>, X<17>, X<18>, X<19>> = X<0>();
+    v := make_variant(std::make_index_sequence<30>());
+
+    for_each_type(v, :<T>(x : T) = {
+        v&$* = X<T::index>(); std::cout << "v as X<" << std::setw(2) << T::index << "> = " << std::setw(1) << v&$* as X<T::index> as int << std::endl;
```
JohelEGP, line 24:
```suggestion
        v&$* = X<T::index>(); std::cout << "v as X<(T::index:2)$> = (v&$* as X<T::index> as int:1)$\n";
```
IIUC, after commit e38838e678a2f52f1c811cf7cb7f819ab3925d62.

regression-tests/mixed-overview-of-as-casts.cpp2
```diff
@@ -0,0 +1,196 @@
...
+
+        vd: VD = ();
+        vp2 : *VA<2> = vd&;
+        print("( vp2* as VC )", expect_throws(:() = ( vp2$* as VC ) is (cpp2::nonesuch);), "throws");
+
+    }
+    {// Variant
+        v : std::variant<int, long, float, double, std::string> = 42;
+
+        print("( v{42} as int ) is (42)", ( v as int ) is (42), true);
```
JohelEGP, line 78: Seems like `fmt` uses parentheses: <https://github.com/fmtlib/fmt/blob/44f3d8a77cd7e05e9da92d68635abdb4da626e9e/doc/api.rst#formatting-variants>.

JohelEGP (commented):
This is preexisting, but I have previously commented at <https://github.com/hsutter/cppfront/pull/529/files/06d8e5f1229454b6b1a1a392be00d887b03d5e0c#diff-9b9f6d5ded63da9a82c7f23fde117cd494a355abad3624569e71a7bf73f974deR81-R82>:
```Cpp2
      // Like in P2392, the cases of the built-in `is` should be a chain of conditions.
      // Using overloads to implement that is tedious and error-prone.
```
I think that has more potential for a simpler implementation.
I wouldn't be surprised if the use of `non`-concepts or negative constraints dropped to zero.

include/cpp2util.h
```diff
@@ -1019,6 +1167,34 @@ inline constexpr auto is( auto const& x, auto const& value ) -> bool
...
+inline constexpr auto is( X const& x, bool (*value)(X &) ) -> bool {
+    if constexpr ( requires { { value(x) } -> std::convertible_to<bool>; } ){
+        return value(x);
+    }
+    return false;
+}
+
+template <typename X>
+inline constexpr auto is( X const& x, bool (*value)(X) ) -> bool {
+    if constexpr ( requires { { value(x) } -> std::convertible_to<bool>; } ){
```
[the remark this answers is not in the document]
filipsajdak: Oh, you are right... this is my mistake. I will take another look on that and check it.

include/cpp2util.h
```diff
@@ -1053,15 +1229,16 @@ inline constexpr auto program_violates_type_safety_guarantee = sizeof...(Ts) < 0
 
 //  For literals we can check for safe 'narrowing' at a compile time (e.g., 1 as std::size_t)
 template< typename C, auto x >
-inline constexpr bool is_castable_v =
+concept castable = requires{ C{x}; } || (
     std::is_integral_v<C> &&
     std::is_integral_v<CPP2_TYPEOF(x)> &&
     !(static_cast<CPP2_TYPEOF(x)>(static_cast<C>(x)) != x ||
         (
             (std::is_signed_v<C> != std::is_signed_v<CPP2_TYPEOF(x)>) &&
```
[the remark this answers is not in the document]
filipsajdak: Oh, yes... I was reminding myself to correct this line.

regression-tests/mixed-as-for-variant-20-types.cpp2
filipsajdak: I will try this syntax. I was not confident in using it.

regression-tests/mixed-overview-of-as-casts.cpp2
filipsajdak: I used them before Herb introduced fmt... should I replace them with something else?

JohelEGP: No, I think it's OK.
What stood out to me was the braces.
It looked like initialization, but upon closer inspection, it's the held value.
I got curious as to what `fmt` does, and found out it's `variant(42)`.

JohelEGP:
> I was interested in making C++ code work. Based on that I was able to call it from cpp2.

Yeah, the functionality works fine.
It works if the lambda was declared with Cpp1,
and given `:<T> () requires std::integral<T> {}`, too.

filipsajdak:
I found errors in my use of concepts (in terms of benefit from subsumption rules). I will update it soon.

JohelEGP:
Although it's in a different context,
note that already I'm using `<T>` to mean a template template parameter:
<https://github.com/hsutter/cppfront/pull/603/files#diff-1c3784de2023bbe4a493b11b40cad588b19a1ca2219eaeb42589910b99ea3effR1>.

filipsajdak:
> > ```
> > inspect x -> std::string {
> >   is <std::integral> = "x is integral type";
> >   is <std::floating_point> = "x is floating point type";
> >   is _ = "other type";
> > }
> > ```
> 
> 
> Although it's in a different context,
> note that I'm already using `<T>` to mean the natural thing,
> a template template parameter:
> https://github.com/hsutter/cppfront/pull/603/files#diff-1c3784de2023bbe4a493b11b40cad588b19a1ca2219eaeb42589910b99ea3effR1.

Cool! I will take a look at that.

Meanwhile, I am working on correcting this PR, as I found some things that could be improved...

include/cpp2util.h
```diff
@@ -1091,13 +1099,18 @@ constexpr auto is( X const& ) -> bool {
...
 template< polymorphic C, polymorphic X >
 constexpr auto is( X const& x ) -> bool {
     return Dynamic_cast<C const*>(&x) != nullptr;
 }
 
-template< polymorphic C, polymorphic X >
-constexpr auto is( X const* x ) -> bool {
+template< polymorphic C, pointer X >
+    requires polymorphic<std::remove_pointer_t<X>>
+constexpr auto is( X const& x ) -> bool {
```
JohelEGP, line 1113: You should still take the pointer by value.

filipsajdak: I will check it. I had an issue with an ambiguous calls with pointers.

filipsajdak: OK, works! Thanks will correct it!

include/cpp2util.h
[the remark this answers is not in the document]
filipsajdak: All of the overloads were not needed - only one is. Thank you @JohelEGP. Fixed!

include/cpp2util.h
filipsajdak: Fixed!

regression-tests/mixed-as-for-variant-20-types.cpp2
filipsajdak: It is not supported by apple-clang-14

JohelEGP (commented):
This goes a long way to addressing #492.
It falls short due to:
> Once operator is is constrained or answered statically,
> it becomes necessary to fix inspect
> to omit alternatives whose conditions are non-viable.

Which doesn't seem to be a problem here yet.
That may be due to `is` overloads like for `std::variant`,
which still returns `bool` even though sometimes it's always `false`.
I argue that, ideally, that'd be constrained so as
to get a compile-time error outside `inspect`, and
an elided alternative in an `inspect`.

include/cpp2util.h
```diff
@@ -933,6 +1030,32 @@ inline auto to_string(auto&& value, std::string_view) -> std::string
 }
 #endif
 
+//-----------------------------------------------------------------------
+//
+//  and "as std::string" for the same cases
+//
+template <typename T>
+concept has_to_string_overload = requires{ static_cast<std::string (*)(T const&)>(&to_string); };
```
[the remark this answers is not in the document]
filipsajdak: I will check these... all my tests pass. Nevertheless, I will double-check these.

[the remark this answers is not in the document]
JohelEGP: It seems like only the last one doesn't pass: <https://cpp2.godbolt.org/z/871q7P5W7>.
I wonder why the others do.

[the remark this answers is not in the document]
JohelEGP: `cpp2::has_to_string_overload<std::string>` passes thanks to
```C++
template<typename T>
inline auto to_string(T const& sv) -> std::string
    requires (std::is_convertible_v<T, std::string_view> 
              && !std::is_convertible_v<T, const char*>)
{
    return std::string{sv};
}
```
See <https://cpp2.godbolt.org/z/5ae4Pxr3r>.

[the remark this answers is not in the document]
JohelEGP: It makes no sense.
Actually calling it chooses the overload returning `std::string const&`:
<https://cpp2.godbolt.org/z/YzPf9ovjv>.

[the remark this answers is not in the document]
filipsajdak: You were right. I made a mistake here. I put the corrected version here: https://godbolt.org/z/WbGqGnEY1

the `const char*` was matching the `as(bool x)` and there were more issues... I believe I have solved them... I will run the tests after Bjarne's talk.

include/cpp2util.h
```diff
@@ -945,7 +946,8 @@ inline auto to_string(char const& t) -> std::string
     return std::string{t};
 }
 
-inline auto to_string(char const* s) -> std::string
+template <std::same_as<char const*> X>
+inline auto to_string(X& s) -> std::string
```
JohelEGP, line 950:
```suggestion
inline auto to_string(const X& s) -> std::string
```
Otherwise, can't pass a `const` `char const*`: <https://cpp2.godbolt.org/z/q44111ezM>.

regression-tests/mixed-type-safety-1.cpp2
```diff
@@ -13,7 +13,7 @@ class Square : public Shape { };
 print: ( msg: std::string, x: _ ) =
     std::cout << msg << x << "\n";
 
-print: ( msg: std::string, b: bool ) =
+print: <B:std::convertible_to<bool>> ( msg: std::string, copy b: B ) =
```
JohelEGP, line 16: FYI, what I did at <https://github.com/hsutter/cppfront/pull/529/files#diff-508db04ed7e2433d79b037f5473ae5ca5330b4b04b842136ad0c95a40142fee8>
was to remove this function and preface `main` with `std::cout << std::boolalpha;`.

include/cpp2util.h
filipsajdak: fixed

regression-tests/mixed-type-safety-1.cpp2
filipsajdak: I just thought that two versions of the `print` functions were to show that the result will match the one with `bool`. If it is not needed then your version is the right one.

JohelEGP: We'd have to ask @hsutter for the purpose.

filipsajdak:
I have reworked a lot during my preparation for the talk on cppcon. After the talk, I have done even more... I am changing this PR to the Draft for a moment as I will clean it up. The current status is that it can handle many more cases and switches to perfect forwarding.

I will be traveling today, so I may manage to clean it up.

include/cpp2util.h
```diff
@@ -1269,132 +1345,197 @@ inline constexpr auto is( X const& x, bool (*value)(X const&) ) -> bool {
...
+          && (!narrowing_to<typename std::remove_cvref_t<X>::value_type,C>) 
+          && (!std::same_as<std::remove_cvref_t<X>,C>)
+          && (!can_bound_to<typename std::remove_cvref_t<X>::value_type,C>)
+auto as( X&& x ) -> decltype(auto) {
+        //std::cout << __PRETTY_FUNCTION__ << " " << __LINE__ << std::endl;
+    return forward_like<forward_like_t<C,X&&>>(*x);
+}
+
+template <typename T>
+using pointee_t = std::remove_cvref_t<decltype(*std::declval<std::remove_cvref_t<T>>())>;
```
JohelEGP, line 1425: Seems like you want `std::iter_value_t<T>`.

include/cpp2util.h
```diff
@@ -1433,50 +1578,59 @@ auto is( std::variant<Ts...> const& x, C const& value ) {
...
-    T* ptr = nullptr;
-    type_find_if<Ts...>([&]<typename It>(It const&) -> bool {
-        if constexpr (std::is_same_v< typename It::type, T >) { if (x.index() ==  It::index) { ptr = &std::get<It::index>(x); return true; } }; return false;
-    });
-    if (!ptr) { Throw( std::bad_variant_access(), "'as' cast failed for 'variant'"); } 
-    return *ptr;
-}
+template <typename X, template<typename...> class C>
+concept specialization_of_template = requires (std::remove_cvref_t<X> x) {
+    { is<C>(std::forward<X>(x)) } -> std::same_as<std::true_type>;
```
JohelEGP, line 1575:
```suggestion
concept specialization_of_template = requires {
    { is<C>(std::declval<X>()) } -> std::same_as<std::true_type>;
```

include/cpp2util.h
```diff
@@ -1846,23 +2003,26 @@ CPP2_FORCE_INLINE constexpr auto cmp_greater_eq(auto&& t, auto&& u) -> decltype(
...
+    } else if constexpr ( requires { {as<C>(std::forward<X>(x))} -> std::same_as<nonesuch_<casting_errors::no_to_string_cast>>; } ) {
         static_assert(
-            program_violates_type_safety_guarantee<C, decltype(x)>,
-            "No safe 'as' cast available - please check your cast"
-        );
-    }
-    //  else
-    return as<C>(CPP2_FORWARD(x));
+            program_violates_type_safety_guarantee<C, X>,
+            "No cpp2::to_string overload exists for this type!");
```
JohelEGP, line 1980: How is this violated?
There's a catch-all overload `inline auto to_string(...) -> std::string`.

JohelEGP (changes requested):
This many overloads, constraints, and negative constrains, really is nightmare fuel.
I really think we should tame these overloads with more `if constexpr` instead of overloads.

For example, rather than one overload with first template parameter `std::same_as<std::string>`
for each kind of function argument,
e.g.,
`std::same_as<std::any> const&`
```C++
typename X>
    requires has_to_string_overload<std::remove_cvref_t<X>>
            && (!std::same_as<std::remove_cvref_t<X>, std::any>)
            && (!std::same_as<std::remove_cvref_t<X>, std::string>)
auto as( X&& x )
```
```C++
typename X>
    requires pointer_like<std::remove_cvref_t<X>>
auto as( X&& x )
```
```C++
typename X>
    requires pointer_like<std::remove_cvref_t<X>>
             && std::same_as<typename X::element_type, std::string>
auto as( X&& x )
```
```C++
typename X>
    requires (!has_to_string_overload<std::remove_cvref_t<X>>)
            && (!std::same_as<std::remove_cvref_t<X>, std::string>)
            && (!pointer_like<std::remove_cvref_t<X>>)
auto as( X&& )
```
have one single overload with first template parameter `std::same_as<std::string>`,
and factor in the overloads' function parameters, like
```C++
template <std::same_as<std::string> C, typename X>
auto as( X&& x ) -> decltype(auto) {
    // No `std::same_as<std::any> const&`,
    // instead rewrite is as `cpp2::op_as` to further tame the big `as` overload.
    if constexpr (requires { op_as<C>(std::forward<X>(x)); }) {
        return op_as<C>(std::forward<X>(x)); // `op_as` for `::std` types need to before these `as` overloads.
    }
    else if constexpr (has_to_string_overload<std::remove_cvref_t<X>>
            && (!std::same_as<std::remove_cvref_t<X>, std::any>)
            && (!std::same_as<std::remove_cvref_t<X>, std::string>)) {
        return to_string(x);
    }
    else if constexpr (pointer_like<std::remove_cvref_t<X>>) {
        if (x) {
            return as<C>(forward_like<X>(*x));
        }
        return std::string("(empty)");
    }
    else if constexpr (pointer_like<std::remove_cvref_t<X>>
             && std::same_as<typename X::element_type, std::string>) {
        if (x) {
            return as<C>(forward_like<X>(*x));
        }
        return std::string("(empty)");
    }
    else if constexpr ((!has_to_string_overload<std::remove_cvref_t<X>>)
            && (!std::same_as<std::remove_cvref_t<X>, std::string>)
            && (!pointer_like<std::remove_cvref_t<X>>)) {
        return nonesuch_<casting_errors::no_to_string_cast>{};
    }
}
```
Then it becomes much easier
to see what's going on,
to simplify by removing negative requirements handled by earlier branches,
and possibly more good stuff.

include/cpp2util.h
filipsajdak: `pointee_t` is also used for things like `std::optional`, smart pointers, etc... in general, anything you can dereference.

include/cpp2util.h
```diff
@@ -1269,132 +1345,197 @@ inline constexpr auto is( X const& x, bool (*value)(X const&) ) -> bool {
...
-    (std::is_enum_v<From> && std::is_floating_point_v<To>) ||
-    // [dcl.init.list] 7.4
-    (std::is_integral_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    (std::is_enum_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    // [dcl.init.list] 7.5
-    (std::is_pointer_v<From> && std::is_same_v<To, bool>);
+constexpr inline nonesuch_<> nonesuch;
+
+template <typename X>
+concept nonesuch_specialization = requires (std::remove_cvref_t<X> x) {
```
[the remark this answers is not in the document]
filipsajdak: I will check that... I have started to put `std::remove_cvref_t<X>` in a lot of the places as I was switching to perfect forwarding... and all type traits stop to work.

include/cpp2util.h
filipsajdak: Yes, that is right. There is a catch-all overload. That works for `cpp2::to_string` but is incompatible with the behavior of `as` when the cast is impossible.

The contract for `as` is that when you call it, you either have the correct output or an error (`static_assert` when possible and an exception on runtime).

Having `as` silently return `"(customize me - no cpp2::to_string overload exists for this type)" breaks that contract. We can detect that on compile time, so I think `static_assert` is the right choice. 

A similar thing will happen when you try to cast `int` to `size_t` (`size_t` cannot represent all values of `int`, so you will get static_assert with a suggestion to use `unsafe_narrow<T>()`

include/cpp2util.h
filipsajdak: most probably `std::remove_cvref_t<X>` and `std::forward<X>(x)` are not needed. I have been wondering what is better: use `std::declval<X>()` or `requires(X x) { ... }`. The second option is more compelling to me (especially after writing many concepts.

I am continue to clean this code up.

filipsajdak:
Regarding many overloads. It is my work-in-progress thing, and I want to clean it up. Having multiple overloads allows us to create methods that will return `std::true_type`/`std::false_type` in parallel with methods that return other values.

As we want to allow for the custom operator as, and operator is, I started to worry that having a `catch them all` method will make that harder. I also want to avoid situations when I need to change old functions when I am adding a new cast.

filipsajdak:
I am preparing a lot of test cases as it is really hard to spot all those errors. I have also found some errors that were silently there but changing the approach made them errors.

include/cpp2util.h
JohelEGP: It works with those: <https://compiler-explorer.com/z/af1c11aa5>.

include/cpp2util.h
JohelEGP: I agree that choosing that overload is wrong.
It's good to know that using `as std::string` doesn't choose it.
I guess it works this way because the concept that checks for the overload checks for conversion to function pointer, and the `(...)` parameter list doesn't work with those.

JohelEGP:
> Having multiple overloads allows us to create methods that will return `std::true_type`/`std::false_type` in parallel with methods that return other values.

If a single branch is chosen, there's no chance for ambiguity in the return type.

JohelEGP:
> I also want to avoid situations when I need to change old functions when I am adding a new cast.

I think it's easier to add new a branch than a new overload.

include/cpp2util.h
filipsajdak: Yes. It does not cast.

filipsajdak: That is why I introduce this concept.

filipsajdak:
I will check both styles.

filipsajdak:
Ha! I think I found the way to support `is` with `type_trait`:

(@JohelEGP, yes, I followed your approach with `if constexpr`s)
```cpp
template <template <typename> class C, typename X>
    requires std::derived_from<C<std::remove_cvref_t<X>>, std::true_type>
            || std::derived_from<C<std::remove_cvref_t<X>>, std::false_type>
auto is( X&& ) -> decltype(auto) {
    std::cout << __PRETTY_FUNCTION__ << std::endl;
    if constexpr ( 
        C<X&&>::value 
        || C<std::remove_reference_t<X>>::value
        || C<std::remove_cv_t<X>>::value 
        || C<std::remove_cvref_t<X>>::value 
    ) {
        return std::true_type{};        
    } else {
        return std::false_type{};
    }
}
```

I am not 100% sure if it will work for all cases, but it seems to work for the following cases:
```cpp
    "type_trait"_test = : () = {
        i  : int       = 42;
        ci : const int = 24;

        expect((i is std::is_const) == false) << "i is std::is_const";
        expect((ci is std::is_const) == true) << "ci is std::is_const";
        expect((ci is std::is_integral) == true);
        expect((ci is std::is_floating_point) == false);

        expect((i is std::is_reference) == true);
        expect((i is std::is_lvalue_reference) == true);
        expect((i is std::is_rvalue_reference) == false);
        expect(((move i) is std::is_rvalue_reference) == true);
        expect((42 is std::is_rvalue_reference) == true);
        expect((i is std::is_signed) == true);
        
        expect((i is std::is_unsigned) == false);
        expect((std::as_const(i) is std::is_unsigned) == false);
        expect(((move i) is std::is_unsigned) == false);

        vc: VC = ();
        expect((vc is std::has_virtual_destructor) == true);
        a: A = ();
        expect((a is std::has_virtual_destructor) == false);
    };
```

JohelEGP:
That's interesting.
It works when the base characteristic is `std::bool_constant`.
It doesn't work generally, like for <https://en.cppreference.com/w/cpp/meta#Property_queries>
(e.g. `8 is std::alignment_of<std::string>`).

filipsajdak:
It is how `is` works. It inspects the variable (or type) on the left-hand side.

In theory we could have `as` to get alignment and is to check it. 

```cpp
std::string as std::alignment_of is 8
```

But it probably messes with the meaning of `as`

filipsajdak:
But then maybe better will be to just ask:
```cpp
std::alignment_of<std::string> is 8
```

JohelEGP:
Right.
It makes sense when you view the second argument as being a special case of a predicate, in this case a type trait predicate.
`8 is std::alignment_of_v<std::string>` should work just fine.

JohelEGP:
Here's a case where it falls short (<https://compiler-explorer.com/z/9WqYhf8e6>):
```Cpp2
#include <concepts>
#include <type_traits>

template <template <typename> class C, typename X>
    requires std::derived_from<C<std::remove_cvref_t<X>>, std::true_type>
            || std::derived_from<C<std::remove_cvref_t<X>>, std::false_type>
auto is( X&& ) -> decltype(auto) {
    if constexpr ( 
        C<X&&>::value 
        || C<std::remove_reference_t<X>>::value
        || C<std::remove_cv_t<X>>::value 
        || C<std::remove_cvref_t<X>>::value 
    ) {
        return std::true_type{};        
    } else {
        return std::false_type{};
    }
}

static_assert(!std::is_constructible_v<int&>);
int i = 0;
static_assert(decltype(is<std::is_default_constructible>(i))::value);
```
Both pass.
It's unfortunately fragile to dealing with references, and probably `const`, when the traits are sensible to them.
Getting such information from the deduced parameter's type is error prone.

filipsajdak:
But `i` which is `int` is default constructible. In this case `is` inspects the variable. 

The behavior could be different when inspecting types.

filipsajdak:
@JohelEGP @hsutter I managed to clean it up enough to start the review. There might be still some issues to be cleaned. What I would like to check is the functionality. The tests pass.

I have some doubts regarding recursive checks of types like `variant` or `optional`. To illustrate it:

```cpp
o : std::optional<std::optional<int>> = 42;

expect((o is 42) == true);
expect((o is 24) == false);
expect((o is :(v) v == 42;) == true); // works thanks to template< class T, class U > constexpr bool operator==( const optional<T>& opt, const U& value);
expect((o is :(v:std::optional<int>) v == 42;) == true);
expect((o is :(v:int) v == 0;) == false);

// ---
v : std::variant<std::optional<int>, std::variant<int>> = std::optional(42);
expect((v is 42) == true);
expect((v is (std::optional(42))) == true);
expect((v is std::optional<int>) == true);

// ---
v.emplace<1>(24);
expect((v is 24) == true);
expect((v is (std::optional(42))) == false);
expect((v is std::optional<int>) == false);

// ---
o2 : std::optional<std::variant<std::optional<int>, double>> = 112;
expect((o2 is 112) == true);
```
(I am using `boost::ut` library for my tests - happy to see it working with cppfront).

filipsajdak:
I will rewrite the commits to be more descriptive but I would like to confirm the functionality first.

JohelEGP:
> I have some doubts regarding recursive checks of types like `variant` or `optional`. To illustrate it:

What are your doubts?

filipsajdak:
Well...

Asking `x is std::variant`, what do you expect? If `x` is `std::variant` to get `true`, that is fine. What should happen when `x` is `std::variant<std::variant<std::variant<int, long, std::variant<int>>, int, long>, int, long>` that have one of the internal `variant` initialized?

Currently, it is inspected internally, and if any internal `variant` is initialized, then the answer is `true`.

filipsajdak:
And if none of the internal `variant` is initialized then it returns `true` as well as `x` is `std::variant` itself.

JohelEGP:
> Asking `x is std::variant`, what do you expect?

A consistent behavior that is not surprising in generic contexts.
I don't expect most types to be specialize `operator is` for _`value`_` is `_`template`_.
So here, I'd expect the built-in behavior.

What do you think `x is std::basic_string`, where `x` has type `std::variant<std::string>`, should return?
I think it should have the built-in behavior, i.e., `false`.

Without universal template parameters, that can't generally give the correct answer.
Today, `std::integer_sequence<int, 17, 29>() is std::integer_sequence` is ill-formed: <https://cpp2.godbolt.org/z/c81dxxGj8>.

filipsajdak:
@JohelEGP support for `std::integer_sequence` added. Now the following code works:
```cpp
std::integer_sequence<int, 17, 29>() is std::integer_sequence; // returns true
```
I have also checked the following cases:
```cpp
std::integer_sequence<int, 17>() is std::integer_sequence;
std::integer_sequence<bool, true>() is std::integer_sequence;
std::integer_sequence<bool, false>() is std::integer_sequence;
std::integer_sequence<int, 17, 29, 13, 42, 42, 54, 64>() is std::integer_sequence;
```

filipsajdak:
> What do you think `x is std::basic_string`, where `x` has type `std::variant<std::string>`, should return? I think it should have the built-in behavior, i.e., `false`.

In the current implementation, `x is std::basic_string` will return `true` (that is why I doubt if it should do that kind of inspection). Although, that kind of behaviour might be expected in context of `inspect`:
```cpp
inspect x -> std::string {
  is std::basic_string = "basic_string";
  is _ = "unknown";
};
```
In such context it is probably expected to do "deep inspection" of the `x`

JohelEGP:
> Closes #701

This should be #717.

filipsajdak:
Thanks

include/cpp2util.h
```diff
@@ -959,577 +1250,459 @@ inline auto to_string(auto&& value, std::string_view) -> std::string
...
 
+//  Type is Type
+//
+
+template <typename X, typename C>
+auto is() -> std::false_type { return {}; }
+
+template <typename X, typename C>
+    requires std::same_as<X, C> || std::derived_from<X,C>
+auto is() -> std::true_type { return {}; }
```
ntrel, line 1261: I have found an issue:
```c++
auto x = is<T, U>();
x = !is<T, U>(); // error: no match for ‘operator=’ (operand types are ‘std::false_type’ {aka ‘std::integral_constant<bool, false>’} and ‘std::true_type’ {aka ‘std::integral_constant<bool, true>’})
```
So I think the return types need to be `bool`.

gregmarr: @ntrel These functions are not intended to be used like that.
  
>`is` that is not using argument `x` for inspection returns `std::true_type` or `std::false_type`

This allows the compiler to eliminate at compile time possibilities that depend solely on the types involved, and not the runtime value of any objects.

You should never need to assign these values to variables, and if you do, you should never try to reassign one of those variables.

JohelEGP: An `is` expression should result in `bool`.
That means defining its result from the the call to a user-defined `operator is` as `bool(`_`call to user-defined `_`operator is)`.

ntrel: OK, thanks.

filipsajdak:
@hsutter I am fixing `is()` overloads - prototype available here: https://godbolt.org/z/4b8oq8err (I needed to change a couple of other things to improve it). Sorry for the delay... I am fixing the implementation and will align the tests with tests done in Compiler Explorer. I will probably need to do the same for `as()`. Sorry for the inconvenience - I made some refactorings previously, and it seems I did not test it properly (I am fixing it now)/

hsutter:
No worries, I appreciate all the thought and care you're putting into this.

filipsajdak:
I have fix is() cases and I am working on as() cases.

filipsajdak:
@hsutter I moved the implementation forward. Most of the cases are working... I see that CI checks failed on macOS-11 and macOS-latest. I just realized the CPU architecture is the main difference between my platform and the one used on the CI side.

CI uses `x86_64`, and I am using `arm64`. Maybe it will be better to have separate regression tests for apple-clang-14, one for x88_64 and one for arm64. I am still adding test cases and I am trying to fix all found issues and clean up the implementation.

hsutter:
> CI uses `x86_64`, and I am using `arm64`. Maybe it will be better to have separate regression tests for apple-clang-14, one for x88_64 and one for arm64.

Sounds good -- I'm not the expert there so I'd welcome anyone to provide a PR to add that. Thanks!

filipsajdak:
@hsutter, I started splitting this big PR into smaller ones. I believe this one covers the `is()` refactor and can be reviewed and hopefully merged.

I am sorry for the late delivery... life.

filipsajdak:
Wait a sec... sth... is wrong with this PR... fixing.

filipsajdak:
@hsutter ok, now fixed. I messed up with branches. But now it is ready.

hsutter:
Thanks! I'm just looking at merging #956 which @JohelEGP mentioned should go first -- not sure it that will require a rebase, of this, but should be just a rebase?

filipsajdak:
@hsutter @JohelEGP I am looking at change #956, which only brings replacing multiple `is()` and `as()` overloads and replaces them with one `is()` and one `as()` functions and multiple `if constexpr`s inside them.

My implementation of `is()` introduces new functions using function overload (as that was the pattern previously used by Herb). I searched for some rules that might guide me here. I have found one that supports @JohelEGP way: https://github.com/isocpp/CppCoreGuidelines/issues/2099 

I am not insisting on choosing my current solution. Let's agree on one way of implementing this kind of function. I will adjust my implementation to follow what will be decided. I have been worried about having functions with multiple constexpr ifs, but maybe it is just me.

Opinions?

I really want to deliver it as it has not been merged for a long time, and it is troublesome to keep it in a mergeable state.

JohelEGP:
A chain of constexpr if is worrisome indeed, but less so than overloading.

The next step would be to rewrite non-built functions as `op_is`/`op_as`
and let the built-ins forward to those user-defined overloads
(including the ones for `std::` types in `cpp2::`).

For example, you add this `is`-type overload:
```C++
template< typename C, can_bound_to<C> X >
    requires not_same_as<X, std::any>
constexpr auto is( X && ) -> std::true_type {
    return {};
}
```
That `requires` is necessary because everything is named `is`.

The next one, with `requires (!std::derived_from<std::remove_cvref_t<X>, C>)`,
can be avoided thanks to the definite order if a chain of if.
Same for `requires cannot_bound_to<X, C>`.
The other one will also be easy to review
since it's just an addition to the chain.

Only the first one can't be integrated into the chain.
You can write it as an overload
until I get around the next rewrite for `op_is`/`op_as`.

hsutter:
> I really want to deliver it as it has not been merged for a long time, and it is troublesome to keep it in a mergeable state.

Again, thanks for this and for waiting. I've been watching to see when I should review this -- I got the impression you were making one more change with the last discussion above, but maybe I misunderstood? Sorry if I missed that it's ready, please just let me know when it's ready to review and mergeable and I'll look at it next.

filipsajdak:
@hsutter I am closing this... next merge request ongoing.

PR #726, opened 2023-10-04
Update *return-list* grammar for forward/move

ntrel:


JohelEGP (approved):

hsutter:
Thanks!

source/parse.h
```diff
@@ -7455,8 +7455,11 @@ class parser
     //G
     //G return-list:
     //G     expression-statement
-    //G     '->' type-id
-    //G     '->' parameter_declaration_list
+    //G     '->' return-direction? type-id
+    //G     '->' parameter-declaration-list
+    //G
+    //G return-direction: one of
+    //G     'forward' 'move'
```
JohelEGP, line 7462: Now that I think about it, Herb prefers not encoding semantic rules in the grammar.
That would mean removing this _return-direction_ and instead using _parameter-direction_.

PR #728, opened 2023-10-04
Extract positional_printer class to own file(positional_printer.hpp)

lexasub:
I was extract positional_printer class to own file(positional_printer.hpp) and add more "moving"
Why I did it - partly so that the pool request in the future for this class would be simply and in order to create modules in the future.


hsutter:
Thanks, I appreciate the suggestion! I'm going to decline it with thanks right now because I really want to keep the structure simple with a single sequence of headers, one for common things and then one for each compilation step (I/O, lexing, parsing, reflection), and the main driver that does the lowering. In the future when it's more complex, e.g., we want to create different .exes for utilities besides the cppfront compiler that may use subsets of the code, it may make sense to be more granular.

Sorry, but thanks again!

PR #753, opened 2023-10-15
Implement `if` alternatives for `inspect`

ntrel:
Implement another kind of *alternative*:
`if` *logical-or-expression* `=` *statement*

See section 3.1 of https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2392r0.pdf.

Note: Part of the `cppfront.cpp` diff is just an unindent, so it helps to ignore whitespace changes.


hsutter:

PR #759, opened 2023-10-16
reserve `identifier is` for *type-id* `is`

ntrel:
Relevant lines from the grammar:
```c
    //G is-as-expression:
    //G     prefix-expression
    //G     is-as-expression is-type-constraint
    //G     is-as-expression is-value-constraint
    //GTODO     type-id is-type-constraint
```
Both *prefix-expression* and *type-id* can be an *unqualified-id*. `T is` is not implemented yet, but when it is, it would ~~conflict with~~ be parsed as `identifier is` to test an expression.

I understand the usual way of disambiguating between a type or an expression is to require parens to force an expression, so I have done that here to fix 5 `regression_tests`. I have added a temporary error for `identifier is`, because that should be the unimplemented `T is` form.

_Update_: So I've moved the *type-id* form above *prefix-expression* in the grammar.

### DIFF
--- a/regression-tests/mixed-inspect-templates.cpp2
+++ b/regression-tests/mixed-inspect-templates.cpp2
@@ -16,10 +16,10 @@ fun: (v : _) -> std::string = {
 }
 
 fun2: (v : _) -> std::string = {
-    if v is std::vector  { return "std::vector";  }
-    if v is std::array   { return "std::array";   }
-    if v is std::variant { return "std::variant"; }
-    if v is my_type      { return "my_type";      }
+    if (v) is std::vector  { return "std::vector";  }
+    if (v) is std::array   { return "std::array";   }
+    if (v) is std::variant { return "std::variant"; }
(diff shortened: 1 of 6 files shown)


filipsajdak:
I have implemented `T is X` on the cpp side here: https://github.com/hsutter/cppfront/pull/701. If you can express it on the cpp2 side you can generate a `cpp2::is<T, X>()` code. I have tested that on cpp side.

ntrel:
@filipsajdak Thanks! I've started working on *type* `is` *type* only for now, your Cpp1 for that looks quite simple - https://github.com/ntrel/cppfront/tree/type-is. *type* is *template* looks much more involved.

filipsajdak:
I have implemented solution for types on cpp2 side. I am struggling with type as a template. 

If you will need any explanation just let me know.

ntrel:
Closing in favour of #782.

Issue #358, opened 2023-04-11
[BUG] Recognize _postfix-expression_ in _is-value-constraint_


JohelEGP:
The grammar permits the _expression_ of _is-value-constraint_ to be a _postfix-expression_. But if the _postfix-expression_ is not a _primary-expression_, it fails as below.

<details><summary>Relevant Cpp2 grammar extract.</summary>

```C++
//G postfix-expression:
//G     primary-expression
//G     postfix-expression postfix-operator     [Note: without whitespace before the operator]
//G     postfix-expression '[' expression-list ']'
//G     postfix-expression '(' expression-list? ')'
//G     postfix-expression '.' id-expression
//G
//G is-as-expression:
//G     prefix-expression
//G     is-as-expression is-type-constraint
//G     is-as-expression is-value-constraint
//G     is-as-expression as-type-cast
//GTODO     type-id is-type-constraint
//G
//G is-type-constraint
//G     'is' type-id
//G
//G is-value-constraint
//G     'is' expression
```

</details>

<details><summary>Current full Cpp2 grammar extracted from the sources.</summary>

```C++
//G binary-digit:
//G     one of '0' '1'
//G
//G digit: one of
//G     binary-digit
//G     one of '2' '3' '4' '5' '6' '7' '8' '9'
//G
//G hexadecimal-digit:
//G     digit
//G     one of 'A' 'B' 'C' 'D' 'E' 'F'
//G
//G nondigit:
//G     one of 'a'..'z'
//G     one of 'A'..'Z'
//G     _
//G
//G identifier-start:
//G     nondigit
//G
//G identifier-continue:
//G     digit
//G     nondigit
//G
//G identifier:
//G     identifier-start
//G     identifier identifier-continue
//G     'operator' operator
//G
//G simple-escape-sequence:
//G     '\' { any member of the basic character set except u, U, or x }
//G
//G hexadecimal-escape-sequence:
//G     '\x' hexadecimal-digit
//G     hexadecimal-escape-sequence hexadecimal-digit
//G
//G universal-character-name:
//G     '\u' hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit
//G     '\U' hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit hexadecimal-digit
//G
//G escape-sequence:
//G     hexadecimal-escape-sequence
//G     simple-escape-sequence
//G
[...]

PR #782, opened 2023-10-25
Implement `Type is Type`

ntrel:
This supplants #759. The same rationale applies for requiring `(identifier) is` to test an expression, see that PR for details. I added a diagnostic for `identifier is expr` where `expr` is not a type-id, because `Type is expr` is not valid - instead it will suggest using `(identifier) is`.

`cpp2util.h` changes are taken from #701 by @filipsajdak. That makes `std::runtime_error is std::exception` true, which seems correct, although I think that is not mentioned in P2392.

I had to add a `speculative` parameter to `parser::type_id()` which is `true` when there might not be a type (but an expression instead). This avoids an error for `ptr*` which is parsed as a postfix-expression from prefix-expression from is-as-expression.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -959,6 +959,15 @@ inline auto to_string(auto&& value, std::string_view) -> std::string
 //  TODO: Does this really warrant a new synonym? Perhaps "is void" is enough
 using empty = void;
 
+//  Type is Type
+//
+
+template <typename X, typename C>
+auto is() -> std::false_type { return {}; }
+
+template <typename X, typename C>
+    requires std::same_as<X, C> || std::derived_from<X,C>
+auto is() -> std::true_type { return {}; }
 
 //  Templates
 //
--- a/source/parse.h
+++ b/source/parse.h
@@ -1438,10 +1438,9 @@ struct is_as_expression_node
     {
         if (type) {
             return type->position();
-        } else if (expr) {
-            return expr->position();
         } else {
-            assert(!"ICE: missing case");
+            assert (expr);
+            return expr->position();
         }
     }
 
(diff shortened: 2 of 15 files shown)


include/cpp2util.h
```diff
@@ -959,6 +959,15 @@ inline auto to_string(auto&& value, std::string_view) -> std::string
...
 using empty = void;
 
+//  Type is Type
+//
+
+template <typename X, typename C>
+auto is() -> std::false_type { return {}; }
+
+template <typename X, typename C>
+    requires std::same_as<X, C> || std::derived_from<X,C>
```
JohelEGP, line 969:
```suggestion
    requires std::derived_from<X,C>
```
`std::same_as<X, C> <= std::derived_from<X, C>` (implication).

source/parse.h
```diff
@@ -6015,10 +6030,10 @@ class parser
 
     //G is-as-expression:
     //G     prefix-expression
+    //GTODO     type-id is-type-constraint
```
JohelEGP, line 6033:
```suggestion
    //G     type-id is-type-constraint
```

JohelEGP (changes requested):

include/cpp2util.h
JohelEGP: Sorry, I was wrong.
`std::same_as` can be `true` for non-class types,
whereas `std::derived_from` can only be `true` for classes.

JohelEGP:
What about the [_alternative_](https://github.com/hsutter/cppfront/issues/387#is-as-expression-target) grammar?

JohelEGP:
> I added a diagnostic for `identifier is expr` where `expr` is not a type-id, because `Type is expr` is not valid - instead it will suggest using `(identifier) is`.

If this implements <https://github.com/hsutter/cppfront/issues/358#issuecomment-1516815580>,
please add "Closes #358" to the opening comment.

hsutter:

PR #785, opened 2023-10-27
Enable debugging with LLDB (Xcode, Qt Creator, etc) v2

bluetarpmedia:
Debugging with LLDB (used by Xcode and Qt Creator) requires the following:

* Print line directives before each function definition
* Use absolute paths for filenames in line directives

### Changes in this PR

1. Print line directives before each function definition
2. Add a new command line option to use absolute paths in line directives (defaults to off). LLDB users will have to turn this on.
3. Update the `test-results` with the new line directives preceding function definitions

I manually verified the line numbers in the `test-results`. I also tested these changes with Visual Studio and the debugging experience continues to work as before, as well as with Xcode/Qt Creator after the changes.

### Changes to test-results

New line directives have been inserted before function definitions.

### ~~Bugfix?~~

There's one change that appears to be a bug fix.

See `pure2-bugfix-for-memberwise-base-assignment.cpp`:

**Before this PR:**
```
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
auto Base::operator=([[maybe_unused]] Base const& that) -> Base&  { std::cout << "(out this, that)\n";
                                      return *this;
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"    <---- bug?
                                     }
```

That second `#line 3` appears to be a bug.

**After this PR:**
```
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
auto Base::operator=([[maybe_unused]] Base const& that) -> Base&  { std::cout << "(out this, that)\n";
                                      return *this; }
```

### DIFF
--- a/regression-tests/test-results/mixed-bounds-safety-with-assert-2.cpp
+++ b/regression-tests/test-results/mixed-bounds-safety-with-assert-2.cpp
@@ -35,6 +35,7 @@ auto add_42_to_subrange(auto& rng, cpp2::in<int> start, cpp2::in<int> end) -> vo
         std::cout << i << "\n";
 }
 
+#line 10 "mixed-bounds-safety-with-assert-2.cpp2"
 auto add_42_to_subrange(auto& rng, cpp2::in<int> start, cpp2::in<int> end) -> void
 {
     cpp2::Bounds.expects(cpp2::cmp_less_eq(0,start), "");
--- a/regression-tests/test-results/mixed-bounds-safety-with-assert.cpp
+++ b/regression-tests/test-results/mixed-bounds-safety-with-assert.cpp
@@ -35,6 +35,7 @@ auto print_subrange(auto const& rng, cpp2::in<int> start, cpp2::in<int> end) -> 
     print_subrange(std::move(v), 1, 13);
 }
 
+#line 9 "mixed-bounds-safety-with-assert.cpp2"
 auto print_subrange(auto const& rng, cpp2::in<int> start, cpp2::in<int> end) -> void{
     cpp2::Bounds.expects(cpp2::cmp_less_eq(0,start), "");
     cpp2::Bounds.expects(cpp2::cmp_less_eq(end,CPP2_UFCS_0(ssize, rng)), "");
--- a/regression-tests/test-results/mixed-captures-in-expressions-and-postconditions.cpp
+++ b/regression-tests/test-results/mixed-captures-in-expressions-and-postconditions.cpp
@@ -41,6 +41,7 @@ auto insert_at(cpp2::in<int> where, cpp2::in<int> val) -> void;
 
 std::vector<int> vec {}; 
 
+#line 19 "mixed-captures-in-expressions-and-postconditions.cpp2"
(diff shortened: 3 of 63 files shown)


gregmarr:
Why do you think it's a bug?  The `}` on that line is generated from the code on line 3 of the `cpp2` source file.  I think the actual bug here is that there isn't a `#line 3` before the `return *this;`.

It should either be this:
```
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
  auto Base::operator=([[maybe_unused]] Base const& that) -> Base&  { std::cout << "(out this, that)\n";
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
                                      return *this;
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
                                     }
```
or this:
```
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
  auto Base::operator=([[maybe_unused]] Base const& that) -> Base&  { std::cout << "(out this, that)\n"; return *this; }
```

bluetarpmedia:
> Why do you think it's a bug? The `}` on that line is generated from the code on line 3 of the `cpp2` source file. I think the actual bug here is that there isn't a `#line 3` before the `return *this;`.

If I put a breakpoint on line 3 in the `Cpp2` file:

```
1 Base: type = {
2   operator=: (out this) = { }
3   operator=: (out this, that)       = std::cout << "(out this, that)\n";
4   operator=: (implicit out this, _) = std::cout << "(implicit out this, _)\n";
5 }
...
```

then I want the debugger to break at the beginning of the function. So if there are multiple `#line 3` directives in the PDB (for example) pointing at different `Cpp1` lines then I was concerned it may choose one of the other directives instead of the one preceding the function.

But your example here:
```
#line 3 "pure2-bugfix-for-memberwise-base-assignment.cpp2"
  auto Base::operator=([[maybe_unused]] Base const& that) -> Base&  { std::cout << "(out this, that)\n"; return *this; }
```
seems ideal to me and would remove all ambiguity. I'm not sure which part of cppfront puts a newline before the generated return but I think it's separate to the line directive changes I made.

gregmarr:
If it can't handle multiple line directives pointing at the same line, then it has no hope of properly debugging this, as those end up all over the place.  

> I'm not sure which part of cppfront puts a newline before the generated return but I think it's separate to the line directive changes I made.

That may be true, but it would be good to know exactly why this PR removes an existing correct line number, so it doesn't cause another issue later.

bluetarpmedia:
>That may be true, but it would be good to know exactly why this PR removes an existing correct line number, so it doesn't cause another issue later.

I've updated the PR to restore the original behaviour of inserting the line directive prior to the closing brace in these emitted assignment operators.

As a result, the changes to the regression tests in this PR only involve inserting new line directives prior to function definitions.

bluetarpmedia:
This PR has been replaced by PR #932.
It was easier to start again based on the new `to_cpp1.h` code rather than updating this version.

Issue #432, opened 2023-05-07
[BUG] `cppfront` confused by function expression in `for`'s (`next`) expression


JohelEGP:
**Title**: Confused by function expression in `for`'s (`next`) expression.

**Description**:

It seems something enters a bad state.
Unparenthesized, it complains about a missing `do`.
Parenthesized, lifetime analysis says the loop variable is uninitialized.

**Minimal reproducer** (<https://cpp2.godbolt.org/z/Er81zc91E>, <https://cpp2.godbolt.org/z/a75sW77bs>):

```Cpp2
main: (args) = {
  for args next :() = 0 do (arg) std::cout << arg;
}
```
```Cpp2
main: (args) = {
  for args next (:() = 0) do (arg) std::cout << arg;
}
```

**Commands**:

```bash
cppfront -clean-cpp1 main.cpp2
```

**Actual result and error**:

```output
main.cpp2(2,25): error: 'for range' must be followed by 'do ( parameter )' (at 'do')
```
```output
main.cpp2(2,49): error: local variable arg is used before it was initialized
  ==> program violates initialization safety guarantee - see previous errors
```

filipsajdak:
`:() = 0` unnamed function should end with `;` (semicolon). There is a bug that I have introduced by adding a check for double semicolons after short function syntax. Probably can be fixed by the following patch:

```diff
diff --git a/source/parse.h b/source/parse.h
index f716925..3165ec0 100644
--- a/source/parse.h
+++ b/source/parse.h
@@ -3606,6 +3606,7 @@ private:
                     && curr().type() != lexeme::LeftParen               // not imediatelly called
                     && curr().type() != lexeme::RightParen              // not as a last argument to function
                     && curr().type() != lexeme::Comma                   // not as first or in-the-middle, function argument
+                    && curr().type() != lexeme::Keyword                 // not as next expression
                 ) {
                     // this is a fix for a short function syntax that should have double semicolon used
                     // (check comment in expression_statement(bool semicolon_required))
```

Additionally, `arg` is not treated as initialized... don't know why.

Also, shall we allow statements as for-body? I was sure that it require a compound statement as ifs.

JohelEGP:
> `:() = 0` unnamed function should end with `;` (semicolon).

When? Always? When not an _expression-statement_ (the only case that would require double `;`s, AFAIK)?

filipsajdak:
[...]

PR #795, opened 2023-11-01
Bugfix #778: Flag for entry into for loop after range has been processed

### LINKED ISSUE #778: [BUG] Passing lambda to function call of for loop causes incorrect uninitialized variable error
**Describe the bug**
Passing a lambda to a for loop parameter's function call produces an incorrect "use before initialization" error.

**To Reproduce**
Sample code:
```
first: (forward f, forward _...) -> forward _ = f;

main: () = {
   ints: const std::array = (1, 2, 3, 4, 5);
   // OK
   for ints.first() do (i) {
      std::cout << i;
   }

   // OK
   for ints.first(1) do (i) {
      std::cout << i;
   }

   // Error
   for ints.first(:(x) x) do (i) {
      std::cout << i;
   }

   // OK
   temp := ints.first(:(x) x);
   for temp do (i) {
      std::cout << i;
   }
}
```

Commands:
```
./cppfront test.cpp2
```

Expected result:
Program compiles without error.

Actual result/error:
```
test.cpp2...
test.cpp2(17,20): error: local variable i is used before it was initialized
```

**Additional context**
I found this while trying to use `std::views::filter` and this seems to be the root cause of the error.

fknauf:
Fixes #778, and it's possibly relevant for #432.

I ran the regression tests and spotted no differences in output. I also have a bash script now to run the tests on a Linux machine, should I put that in here as well?


JohelEGP (approved):
Thank you.
The approach looks correct to me.
Can you add these tests for confirmation?
```Cpp2
issue_778: (args) = {
   for :() args$;() do (i) _ = i;
   for :(x) x;(args) do (j) _ = j;
   for :(x) x;(args) next _ = :() args$;() do (k) _ = k;
   for :(x) x;(args) next _ = :(x) x;(args) do (l) _ = l;
}
```

fknauf:
I added the tests. The output was compileable and looked reasonable to me. Then I merged the new state of the main branch and ran the tests again. All tests are green, so it seems to work and not break anything.

fknauf:
I'm not 100% sure I agree that flags like this one are the correct way to handle parsing context like this, but it seemed like the least invasive way to handle it. In a new design I'd maybe consider encoding information like this in the parse tree so that the safety of the expression becomes an S-attribute.

hsutter:
(Catching up with PRs, sorry for the delay)

hsutter:
Thanks! I just made a minor style tweak to follow the style I was already using for such so-far-very-occasional extra information.

PR #796, opened 2023-11-01
Ensure valid type name for operators with multiple returns.

iprtel:
This allows operators to have multiple returns.

To enable multiple returns from a function  a new struct is created. That struct is named after the function. If the function is a operator this name contains characters that are invalid for type names.

Consider:
```
operator*: (this) -> (x:int, y:int)=  {
        x = x_current;
        y = y_current;
        return ;
}
```
results in 
```
error: expected a type
[build]    53 |         public: [[nodiscard]] auto operator*() const& -> operator*_ret;
```


hsutter:
Thanks! I did it just for `()`, `[]`, and `*` for now because those seem to be the main use cases (and after I picked those, I noticed they were also the ones you chose to exercise which is at least some confirmation), and I'm happier to rein in the overuse/abuse of operator overloading where types really should normally "do as the ints do" (e.g., `<` should return a boolean value, not something more complex). But it's easy to expand this in the future for more operators by just adding them to the renaming table.

PR #867, opened 2023-12-03
Update regression tests for apple-clang-14 (23W48)

filipsajdak:
Catching up with tests for apple-clang-14

Needed before merging #701


hsutter:
Thanks!

PR #893, opened 2023-12-16
CI: Add macos-11 Clang 13 runner, Add badge to Readme

DyXel:
Apple Clang 13 seems to be the lowest version we can go with the pre-installed compilers in the existing runners.

Tried adding Apple M1 (arm64) runner, but it seems like it requires Github Enterprise licensing at this time. Also tried adding Apple Clang 15 (via macos-13 runner), but I got a bunch of linking errors, I think it has to do with different compilers using different versions of libc++, but I am not completely sure. You can try yourself by adding:
```yaml
          - runs-on: macos-13
            compiler: /usr/local/opt/llvm@15/bin/clang
            cxx-std: 'c++20'
```

Finally, this commit also adds a bit of flavor to the Readme by showing that the main branch is currently passing CI.

### DIFF
--- a/.github/workflows/build-cppfront.yaml
+++ b/.github/workflows/build-cppfront.yaml
@@ -27,6 +27,9 @@ jobs:
           - compiler: clang++-14
             cxx-std: 'c++2b'
         include:
+          - runs-on: macos-11
+            compiler: clang++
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks!

PR #894, opened 2023-12-18
Update apple-clang-14 tests

jarzec:
Update outdated tests files in `regression-tests/test-results/apple-clang-14/`.
The update was possible thanks to the regression test workflow added [here](https://github.com/jarzec/cppfront/tree/github-test-workflow-5).
[Here](https://github.com/jarzec/cppfront/actions/runs/7253964295) is the result of running the regression tests.

Note that `std::ranges::for_each` is not supported by clang-14, hence the committed compilation errors.


hsutter:
Thanks!

PR #896, opened 2023-12-20
Update bounds check apple-clang-14 test

jarzec:
Update another MacOS test missed in a recent PR. The issue was spotted thanks to the tools from https://github.com/hsutter/cppfront/pull/897.

### DIFF
--- a/regression-tests/test-results/apple-clang-14/mixed-bounds-check.cpp.execution
+++ b/regression-tests/test-results/apple-clang-14/mixed-bounds-check.cpp.execution
@@ -1,1 +1,1 @@
-Bounds safety violation: out of bounds access attempt detected - attempted index 5, [min,max] range is [0,4]
+Bounds safety violation: out of bounds access attempt detected - attempted access at index 5, [min,max] range is [0,4]


hsutter:
Thanks! For this and for the description with this example of how the test flow works.

The reason the Apple Clang test was not updated in my PR that committed the diagnostic message change is because that's the only one of the checked-in tests I can't run locally on my machine and then commit as part of the PR.

Re CRLF/LF: In just the past couple of weeks I've noticed a few times when I'm checking regression test on a commit and many of the files have only CRLF/LF differences. This seems to be a new thing and I haven't investigated the root cause. I vaguely recall that back when I started cppfront I looked at `autocrlf` but no setting was perfect -- could the setting be improved?

jarzec:
> Thanks! For this and for the description with this example of how the test flow works.
> 
> The reason the Apple Clang test was not updated in my PR that committed the diagnostic message change is because that's the only one of the checked-in tests I can't run locally on my machine and then commit as part of the PR.
> 
> Re CRLF/LF: In just the past couple of weeks I've noticed a few times when I'm checking regression test on a commit and many of the files have only CRLF/LF differences. This seems to be a new thing and I haven't investigated the root cause. I vaguely recall that back when I started cppfront I looked at `autocrlf` but no setting was perfect -- could the setting be improved?

Do you have a preference as to which one you would prefer CRLF or LF? I could create a PR that would enforce the `autocrlf` option for everyone an normalize the repo to the chosen option.

hsutter:
I always do my development and regressions test on Windows, so CRLF would work best for me if it doesn't impact others too much to optimize for that.

PR #906, opened 2023-12-23
Fix possible division by zero when using -verbose

RealY700:
I encountered an issue where using `-verbose` on a pure cpp2 file causes a floating point exception, as there are zero cpp1 lines, this PR fixes it.

To reproduce you may compile this simple file (or any other pure cpp2 file) with `-verbose`, other flags don't matter:
```cc
main: () -> int = {
  return 0;
}
```


hsutter:

RealY700:
I've now signed the CLA, thanks.

hsutter:
Thanks!

PR #911, opened 2023-12-28
CI: Enable pedantic warnings and treat all warnings as errors

DyXel:
This PRs makes the CI workflows fail if the compiler emits warnings for cppfront code (by using `-Werror` on clang/gcc and `-WX` on MSVC), it also adds `-pedantic` to clang/gcc to catch additional oversights.

Separately, I provide narrowing conversion "fixes", but I am not too fond of this commit, since it feels like the correct thing to do would be to promote `int` to _whatever `std::ssize` returns_ in most places, I didn't do that yet as I believe this requires more thought on the design of the compiler's code and want feedback first. The commit could be skipped and done by someone else, or the warning could be disabled (seen that in lots of projects 😛).

### DIFF
--- a/.github/workflows/build-cppfront.yaml
+++ b/.github/workflows/build-cppfront.yaml
@@ -9,7 +9,7 @@ jobs:
       - name: Compiler name & version
         run: cl.exe
       - name: Build
-        run: cl.exe source/cppfront.cpp -std:c++latest -MD -EHsc -experimental:module -W4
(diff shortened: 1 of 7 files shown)


hsutter:
Thanks!

FWIW two flags I use on GCC and Clang that I don't think are included under the other options: `-Wold-style-cast -Wunused-parameter`

DyXel:
> FWIW two flags I use on GCC and Clang that I don't think are included under the other options: `-Wold-style-cast -Wunused-parameter`

`-Wold-style-cast` is already there, `-Wunused-parameter` is not, feel free to add though!

PR #914, opened 2023-12-29
New returns optional of owning reference

iprtel:
This patch represents a suggestion:
Default new should return std::optional<owning_reference<T>> with owning_reference managing the life time of the object in a similar way as a unique_ptr would with the exception that it is not allowed to be null. 

This will allow to move the reference around without null checks.


hsutter:
Hi! Sorry it took so long to touch base on this. I think the current design is sufficient at least for now so I'll close this, but I appreciate the idea. Thanks!

PR #930, opened 2024-01-11
Only #include <stdfloat> on MSVC when in C++23 mode to remove a warning

### LINKED ISSUE #929: [BUG] MSVC (C++20) emits a warning when compiling lowered Cpp1 after using Cppfront with `-include-std`
**Describe the bug**
In C++20 mode, MSVC reports an `STL4038` warning regarding `<stdfloat>` when compiling Cpp1 code that was generated by Cppfront with the `-include-std` flag.

**To Reproduce**
Run Cppfront with the `-in` or `-include-std` flag on this code:
```
main: () -> int = {
    return 0;
}
```
Then compile the result with MSVC with `/std:c++20`.

See here on [Godbolt](https://godbolt.org/z/qn9x35Ynz).

This warning is emitted:
```
stdfloat(12): warning STL4038: The contents of <stdfloat> are available only with C++23 or later.
```

**Additional context**
In `cpp2util.h` the test for including `<stdfloat>` is:

```
#if __has_include(<stdfloat>)
```

This returns true even when building for C++20 with MSVC, and the `<stdfloat>` header contains this:

```
#if !_HAS_CXX23
_EMIT_STL_WARNING(STL4038, "The contents of <stdfloat> are available only with C++23 or later.");
#else // ^^^ !_HAS_CXX23 / _HAS_CXX23 vvv
```

bluetarpmedia:
The `__has_include` test for `<stdfloat>` returns true even in C++20 because the file exists, and the `<stdfloat>` header emits an error when not compiled with C++23.

Fixes #929

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -197,7 +197,9 @@
         #endif
         #include <stdexcept>
         #if __has_include(<stdfloat>)
-            #include <stdfloat>
+            #if (defined(_MSC_VER) && _HAS_CXX23) || !defined(_MSC_VER)
+                #include <stdfloat>
+            #endif
         #endif
         #ifdef __cpp_lib_jthread
             #include <stop_token>


include/cpp2util.h
```diff
@@ -197,7 +197,9 @@
         #endif
         #include <stdexcept>
         #if __has_include(<stdfloat>)
-            #include <stdfloat>
+            #if (defined(_MSC_VER) && _HAS_CXX23) || !defined(_MSC_VER)
```
gregmarr, line 200: This check can be simpler:

    #if !defined(_MSC_VER) || _HAS_CXX23

bluetarpmedia: Thanks, updated!

PR #932, opened 2024-01-11
Enable debugging with LLDB (Xcode, Qt Creator, etc)

bluetarpmedia:
Debugging with LLDB (used by Xcode and Qt Creator) requires the following:

- Print line directives before each function definition
- Use absolute paths for filenames in line directives

### Changes in this PR

1. Print line directives before each function definition
2. Add a new command line option to use absolute paths in line directives (defaults to off). LLDB users will have to turn this on.
3. Update the `test-results` with the new line directives preceding function definitions

I manually verified the line numbers in the updated `test-results` and also tested the debugging experience.

The implementation in this PR is significantly simpler than the previous one (#785, now closed) thanks to the recent refactoring that Herb made in `to_cpp1.h`.

### DIFF
--- a/regression-tests/test-results/mixed-bounds-safety-with-assert-2.cpp
+++ b/regression-tests/test-results/mixed-bounds-safety-with-assert-2.cpp
@@ -37,6 +37,7 @@ auto add_42_to_subrange(auto& rng, cpp2::in<int> start, cpp2::in<int> end) -> vo
         std::cout << i << "\n";
 }
 
+#line 10 "mixed-bounds-safety-with-assert-2.cpp2"
 auto add_42_to_subrange(auto& rng, cpp2::in<int> start, cpp2::in<int> end) -> void
 {
     if (cpp2::Bounds.has_handler() && !(cpp2::cmp_less_eq(0,start)) ) { cpp2::Bounds.report_violation(""); }
--- a/regression-tests/test-results/mixed-bounds-safety-with-assert.cpp
+++ b/regression-tests/test-results/mixed-bounds-safety-with-assert.cpp
@@ -36,6 +36,7 @@ auto print_subrange(auto const& rng, cpp2::in<int> start, cpp2::in<int> end) -> 
     print_subrange(std::move(v), 1, 13);
 }
 
+#line 9 "mixed-bounds-safety-with-assert.cpp2"
 auto print_subrange(auto const& rng, cpp2::in<int> start, cpp2::in<int> end) -> void{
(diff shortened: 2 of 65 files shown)


JohelEGP:
> 2\. LLDB users will have to turn this on.

Is it really an issue with LLDB?
I have no problems debugging with GCC + LLDB.
I do have problems debugging with Clang + LLDB.
The issue seems to be <https://stackoverflow.com/questions/20985673/how-to-configure-qt-creator-to-show-c-code-rather-than-disassembler?rq=3>.
I've solved it by making the input source to `cppfront` be an absolute path.

bluetarpmedia:
> I've solved it by making the input source to `cppfront` be an absolute path.

Yeah, supplying an absolute path as the argument to `cppfront` solves the problem, but I added the flag so that users (or tools) can still supply a relative path for convenience while ensuring that the line directives are absolute.

gregmarr:
It still boggles my mind that lldb won't work properly without redundant #line specifiers.  They truly are redundant.  Every single one of them has absolutely zero effect on the effective line number based on the previous #line specifiers in the file.  Has this been reported to lldb as a bug?

bluetarpmedia:
> It still boggles my mind that lldb won't work properly without redundant #line specifiers. They truly are redundant. Every single one of them has absolutely zero effect on the effective line number based on the previous #line specifiers in the file. Has this been reported to lldb as a bug?

I haven't investigated further to find out whether the problem is with clang+lldb or just lldb. What I found on macOS with Xcode and Qt Creator is that the line directives preceding function declarations were not enough for breakpoints and stepping to work, and the workaround I found was to put the directives before each function definition.

MaxSagebaum:
> > 2. LLDB users will have to turn this on.
> 
> Is it really an issue with LLDB? I have no problems debugging with GCC + LLDB. I do have problems debugging with Clang + LLDB. The issue seems to be https://stackoverflow.com/questions/20985673/how-to-configure-qt-creator-to-show-c-code-rather-than-disassembler?rq=3. I've solved it by making the input source to `cppfront` be an absolute path.

The same issue exists with GDB. I had to copy the `regex.h2` from the cppfront source directory to the test directory in order to debug the tests.

JohelEGP:
> GDB

Yeah, that's what I (and QtCreator) use, not LLDB.

hsutter:
Looks good -- do I need to worry about the 3 failing tests before merging, or are those failures spurious-ish/otherwise-non-blocking?

bluetarpmedia:
> Looks good -- do I need to worry about the 3 failing tests before merging, or are those failures spurious-ish/otherwise-non-blocking?

Those are the same failures happening in ‘main’ and I created #933 to fix that.

hsutter:
Thanks!

PR #933, opened 2024-01-12
Fix macOS and Ubuntu regression tests

bluetarpmedia:
Add missing test-results for the `pure2-bugfix-for-indexed-call` test for macOS `apple-clang-14` and ubuntu `gcc-10` and `gcc-13`.


jarzec (approved):

hsutter:
Thanks!

PR #936, opened 2024-01-12
CI: Add clang++-12 regression tests run on Ubuntu

jarzec:
The current regression tests workflow is missing a run using clang++-12.
This PR adds this compiler and updates the respective tests.

Note that this branch is based on https://github.com/hsutter/cppfront/pull/935, which should be merged first.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  regression-tests-linux-mac:
+  regression-tests:
     name: Run on ${{ matrix.os }} using ${{ matrix.compiler }}
     runs-on: ${{ matrix.os }}
     env:
@@ -32,7 +32,6 @@ jobs:
       run: |
         cd regression-tests
         bash run-tests.sh -c ${{ matrix.compiler }}
-      continue-on-error: true
 
     - name: Run regression tests - Windows version
       if: matrix.os == 'windows-latest'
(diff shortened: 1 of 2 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         os: [ubuntu-latest]
         compiler: [g++-10, g++-13]
         include:
+          - os: ubuntu-20.04
+            compiler: clang++-12
```
jarzec, line 23: clang++-12 is not installed by default on the `ubuntu-latest` image, but it is on `ubuntu-20.04`, so this is the easiest way to include it.

JohelEGP (approved):

hsutter:
Thanks! @JohelEGP @jarzec @bluetarpmedia I see a few apparently-related PRs by the three of you, is there an order they should be merged?

JohelEGP:
By their PR number should work OK.

JarekGlobus:
> Thanks! @JohelEGP @jarzec @bluetarpmedia I see a few apparently-related PRs by the three of you, is there an order they should be merged?

@hsutter To be more precise:
- #933 fixes test failures that managed to get through some time around merging of the regression test workflow.
- #935 improves the patch generation. It has build failures (present on the main branch) that will go away after merging #933.
- This PR adds run for clang 12. It is based on  #935 and should be merged after. It also has the same tests failing. Those failures will also be solved by #933.

hsutter:
Thanks!

PR #937, opened 2024-01-14
Fix: MSVC warning 5050 was not really suppressed

adah1972:
The pragma should be put before the import, not after.


adah1972:
The warning still appeared when I issued the command line as follows:

```
cppfront -p demo.cpp2
cl /nologo /std:c++20 /EHsc /Iinclude /MD /experimental:module demo.cpp
```

The warning is now gone after the change.

hsutter:

hsutter:
Thanks!

PR #938, opened 2024-01-16
Ignore Visual Studio and VSCode cache/workspace directories in Git

bluetarpmedia:
Add `.vs` and `.vscode` to `.gitignore`

Visual Studio creates the `.vs` cache directory when opening the `cppfront` folder.
VSCode creates the `.vscode` workspace directory if any workspace settings are specified.


hsutter:
Thanks!

PR #940, opened 2024-01-16
Update line numbers in test compilation errors

jarzec:
After #932 line numbers in compilation error messages in regression test for certain compilers require an update.


hsutter:
Thanks!

Issue #943, opened 2024-01-17
[BUG] MSVC CI regression-test job doesn't have modules support, meaning `pure2` tests succeed without running


bluetarpmedia:
CC @jarzec and @JohelEGP who are aware of this problem. The same problem happens here: https://github.com/modern-cmake/cppfront/issues/68

## Describe the bug
The CI regression-tests for MSVC are not able to run the `pure2` tests. The tests succeed which means real bugs or regressions can be hidden, unless they're exposed on other compilers.

There seems to be a problem with the GitHub-hosted runner for `windows-latest` (currently an alias for `windows-2022`) with Visual Studio Enterprise. The machine image is [documented](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) to have the experimental modules component installed for Visual Studio. (See `Microsoft.VisualStudio.Component.VC.Modules.x86.x64`)

But when `cl.exe` runs with `-std:c++latest -experimental:module` it reports fatal error C1011:

`cannot locate standard module interface. Did you install the library part of the C++ modules feature in VS setup?`

Strangely, this would mean that `__cpp_lib_modules` is defined to be `true` so that the `import` lines are enabled in `cpp2util.h`.

I updated the GitHub action in my own branch to try and manually install the C++ modules support before proceeding to run the tests, but that had no effect.

`"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" modify --channelId VisualStudio.17.Release --productId Microsoft.VisualStudio.Product.Enterprise -q --add "Microsoft.VisualStudio.Component.VC.Modules.x86.x64"`

So it would appear that the C++ modules component is correctly installed in the GitHub-hosted runner machine image, but for some reason MSVC still fails.

The CI `run-tests.sh` script detects the C1011 error and reports this to stdout: `Skipping further checks due to missing std modules support`

Until we can resolve the problem, I think it's better to change the MSVC tests to use cppfront with `-include-std` so that the tests can actually run rather than silently succeeding.

## To Reproduce
Here is a recent MSVC regression-tests job: https://github.com/hsutter/cppfront/actions/runs/7547757353/job/20548383595

Expand the `Run regression tests - Windows version` section and scroll down to the `pure2` tests. You'll see something like:

```
Testing pure Cpp2 code: pure2-bounds-safety-span.cpp2
        Generating Cpp1 code
        Compiling the generated Cpp1 code
            Skipping further checks due to missing std modules support
```

[...]

PR #944, opened 2024-01-17
MSVC regression-test job in CI now uses `-include-std` to workaround a problem with support for C++ modules

bluetarpmedia:
Fixes (or works around) #943

Note that this fix has exposed some MSVC `pure2` tests which were silently failing:
- pure2-main-args.cpp.execution
- pure2-repeated-call.cpp.execution
- pure2-union.cpp.execution
- pure2-variadics.cpp.execution

I updated all of them except `pure2-main-args.cpp.execution` since I wasn't sure what the correct result should be. The diff from the GitHub agent run is:

```
-args.argv[0]         is test.exe
+args.argv[0]         is D:\cppfront\cppfront
```


hsutter:
Thanks!

regression-tests/run-tests.sh
```diff
@@ -39,7 +39,7 @@ check_file () {
         failure=1
     else
         # Compare the content with the reference value checked in git
-        diff_output=$(git diff --ignore-cr-at-eol -- "$file")
+        diff_output=$(git diff --ignore-cr-at-eol -I"\#define CPP2\_.*\_STD" -- "$file")
```
gregmarr, line 42: Should this also be added to line 35?

bluetarpmedia: No, IIUC, this branch is to handle the case where the test result expected output does not exist in Git, so the `git diff` here is being used to add the new content to a larger `.diff` file. CC @jarzec

PR #945, opened 2024-01-17
Add null violation checks for `std::unique_ptr`, `std::shared_ptr`, `std::optional` and `std::expected` to prevent UB

bluetarpmedia:
`assert_not_null` now verifies `std::unique_ptr`, `std::shared_ptr`, `std::optional` and `std::expected` in addition to the original null pointer check.

These states are considered to be "null" and will result in a violation error if dereferenced or accessed:

- `std::unique_ptr` that owns nothing
- `std::shared_ptr` with no managed object
- `std::optional` with no value
- `std::expected` containing an unexpected value

Null violation errors will be reported in the following cases:
```
test: () = {
    ex: std::expected<int, bool> = std::unexpected(false);
    op: std::optional<int> = std::nullopt;

    up:= unique.new<int>(1);
    sp:= shared.new<int>(2);
    up.reset();
    sp.reset();

    v1 := ex*;   // Null violation error
    v2 := op*;   // Null violation error
    v3 := up*;   // Null violation error
    v4 := sp*;   // Null violation error
}
```

Add new regression tests:
- pure2-assert-expected-not-null
- pure2-assert-optional-not-null
- pure2-assert-shared-ptr-not-null
- pure2-assert-unique-ptr-not-null

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -240,6 +240,9 @@
     #ifndef CPP2_NO_EXCEPTIONS
         #include <exception>
     #endif
+    #ifdef __cpp_lib_expected
+        #include <expected>
+    #endif
     #if defined(__cpp_lib_format) || (defined(_MSC_VER) && _MSC_VER >= 1929)
         #include <format>
     #endif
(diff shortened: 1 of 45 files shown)


JohelEGP (approved):
This seems good enough.

It seems like a more general solution would use `x is T` when `CPP2_TYPEOF(x)` wraps a `T`.
But we just can't guarantee that the precondition of `x*` is the same as `x is T`.

hsutter:
Thanks!

JohelEGP:
The test output isn't platform-independent.
See #556:
> - Update tests that finish through `std::terminate`.
>   * Add `std::set_terminate(std::abort);` to avoid platform-dependent output on termination.

PR #947, opened 2024-01-17
Update `pure2-main-args` test to work on CI agent as well as locally

bluetarpmedia:
Strip any leading directory path from the `argv[0]` program name in `pure2-main-args` test so that there is a consistent result in the GitHub CI agent as well as when run locally.


hsutter:
Thanks!

PR #948, opened 2024-01-18
Small tweak to `run-tests-msvc-2022.bat`: change absolute path to relative

bluetarpmedia:
Replace `c:\github\cppfront\include` absolute path with a relative path; allows users to run the bat file with a different checkout directory of cppfront.

Note: The failing Windows test is the `pure2-main-args` test which is fixed by #947

EDIT: The above note no longer applies; I rebased to bring in the changes from #947 that were merged into `main`


hsutter:
Thanks!

PR #949, opened 2024-01-18
More elegant solution for pure2-main-args.cpp2

jarzec:
In #947 @bluetarpmedia proposed an improvement/fix for pure2-main-args.cpp2 that manually extracts the filename from `args.argv[0]`.
Here I propose a solution using `std::filesystem::path` that I find a bit more elegant.
Unfortunately, due to differences between the implementation details of `std::filesystem::path` between POSIX and Windows the type can be used directly in Cpp2 string interpolation only on POSIX systems. For that reason the solution is not as elegant as it could have been if it wasn't the case.


gregmarr (approved):

hsutter:
Thanks!

Issue #950, opened 2024-01-19
[SUGGESTION] Start using `import std;` in `cpp2util.h` with MSVC and re-enable `-import-std` in CI builds


bluetarpmedia:
Now that MSVC (with `/std:c++latest`) supports `import std;` it's possible to remove the Microsoft-only experimental `import std.core;` etc lines from `cpp2util.h` and replace them with `import std;` or `import std.compat;`.

As described in #943, the `import std.core;` code was not working in the GitHub-hosted CI runner, so the MSVC regression tests have been updated to run cppfront with `-include-std` to `#include` headers instead of importing any modules.

I experimented with `import std;` and GitHub-hosted runners and documented two methods to get them building correctly:

* running `msbuild` for a Visual Studio solution
* using `cl.exe` in a shell, which is applicable for cppfront's GitHub CI workflow

https://github.com/bluetarpmedia/msvc-import-std-github

So this suggestion involves 2 changes:

1. Update `cpp2util.h` to use `import std;` (where available)
2. Update the GitHub CI workflow to restore the default `-import-std` behaviour, and modify the MSVC build job to compile `import std;` modules support first before then compiling the tests (more info in my `msvc-import-std-github` repo)

hsutter:
I was thinking the same. Note that step #2 will require:

1. Build { `std`, `std.compat` } `.` { `ifc`, `obj`} with the identical switches we will use to build the test. The files probably should just stay in the MSVC build current directory since they don't need to be shared, and then we don't need to add a `/reference` switch on the command line. Here's what I used FWIW: `cl /std:c++latest /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.ixx"` and `cl /std:c++latest /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.compat.ixx"`.
    - Note: If the MSVC environment's regression-run cleanup does `del *.obj` (as I do in mine), it should be modified to not erase the `std*.obj` files... I changed my cleanup step to `del pure2-*.obj mixed-*.obj`.
2. Change `/experimental-module` to `std.obj std.compat.obj` to the MSVC command line (including those obj's seems to work fine even for `mixed-*` files that use `#include` and don't use them)

I can go ahead and make all these changes in one commit, including checking in my `std*.ifc` and `std*.obj` files in the MSVC regression test directory, and then everything should be consistent and working except for the CI flow... could you then make a PR to apply the same to the CI environment?

[...]

PR #953, opened 2024-01-22
MSVC regression-test job in CI now builds and links with C++ standard library modules to enable `import std.compat;`

bluetarpmedia:
* Re-enable the `-import-std` / `pure` behaviour
* The CI build script now compiles the `std` and `std.compat` modules and links the `std.obj` and `std.compat.obj` files when building the regression test binaries.

See #950 for more details.

### DIFF
--- a/regression-tests/run-tests.sh
+++ b/regression-tests/run-tests.sh
@@ -149,10 +149,14 @@ fi
 # in order to support `import std.compat;`.
 regression_test_link_obj=""
 if [[ "$cxx_compiler" == *"cl.exe"* ]]; then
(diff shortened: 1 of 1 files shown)


bluetarpmedia:
> Thanks! I guess rebuilding the four `std*` files once per regression test run is not too bad, it seems to take about as long as running a handful of test cases. That's reasonable.

Yes, that's what I was thinking -- I should've mentioned that sorry.

It means we won't run into problems if the compiler flags need to change, or if a Visual Studio update on the GitHub-hosted runner affects the `ifc` or `obj` files.

PR #957, opened 2024-01-25
CI: Update and fix

jarzec:
- Update the `regression-tests/run-tests.sh`
   - Unify the checks using `git diff`.
   - Add a command line argument `-l` for setting a label of the test run to be use in the name of the patch file (if it is created).
   - Update the extension of the generated patch files to be `.patch` which seems to be the accepted convention.
-  Update the `.github/workflows/regression-tests.yml` workflow file.
   - Update the checkout action to avoid GitHub action deprecation warnings.
   - Update the name of the eventual patch artefact file to include the OS identification.
- Update MacOS tests after 29a1f43687192212bd02c2b1da0b2731d7aa2fe4.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -27,13 +27,13 @@ jobs:
             compiler: cl.exe
     steps:
     - name: Checkout repo
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: Run regression tests - Linux and macOS version
       if: startsWith(matrix.os, 'ubuntu') || matrix.os == 'macos-13'
       run: |
         cd regression-tests
-        bash run-tests.sh -c ${{ matrix.compiler }}
+        bash run-tests.sh -c ${{ matrix.compiler }} -l ${{ matrix.os }}
 
     - name: Run regression tests - Windows version
       if: matrix.os == 'windows-latest'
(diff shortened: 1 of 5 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -27,27 +27,27 @@ jobs:
             compiler: cl.exe
     steps:
     - name: Checkout repo
-      uses: actions/checkout@v3
```
jarzec, line 30: [Explanation] v3 was causing deprecation warnings from GitHub

.github/workflows/regression-tests.yml
```diff
@@ -27,27 +27,27 @@ jobs:
...
+        bash run-tests.sh -c ${{ matrix.compiler }} -l ${{ matrix.os }}
       shell: cmd
 
     - name: Upload patch
       if: ${{ !cancelled() }}
       uses: actions/upload-artifact@v4
       with:
-        name: ${{ matrix.compiler }}-patch.diff
-        path: regression-tests/${{ matrix.compiler }}-patch.diff
+        name: ${{ matrix.os }}-${{ matrix.compiler }}.patch
```
jarzec, line 51: [Explanation] Adding the OS label to the patch file name will avoid issue e.g. if the same compiler is used at some stage on Ubuntu and MacOS.

hsutter:
Thanks!

PR #963, opened 2024-02-05
Fix Apple Clang build: Replace `std::integral` concept with `requires` clause

bluetarpmedia:
Apple Clang doesn't support the `std::integral` concept yet.


hsutter:
Thanks!

PR #966, opened 2024-02-06
Add a `static_cast` in `args_t::size` to silence a conversion warning

bluetarpmedia:
Clang reports this warning in `cpp2util.h` when `-Wconversion` is enabled:

```
cpp2util.h:1774:49: warning: implicit conversion changes signedness: 'int' to 'std::size_t' (aka 'unsigned long') [-Wsign-conversion]
 1774 |     auto size()   const -> std::size_t { return argc; }
      |                                          ~~~~~~ ^~~~
```

So this PR adds a `static_cast` to silence the warning and show that the conversion is intentional.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1729,6 +1729,19 @@ private:
 
 //-----------------------------------------------------------------------
 //
+//  An implementation of GSL's narrow_cast with a clearly 'unsafe' name
+//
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks!

Minor nit: Sadly, because of the dratted silliness of mixed signedness, this needs to be `unsafe_narrow`...

PR #972, opened 2024-02-07
CI `run-tests` script now attempts to detect Apple Clang 15 and other unsupported regression-tests compiler versions

bluetarpmedia:
After investigating some spurious macOS regression-test job failures, I found that the `macos-13` [GitHub runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md) has both Apple Clang 14 and 15 installed.

It appears that on some runs, running `clang++` incorrectly selects Apple Clang 15, which is currently unsupported as a compiler version for the regression-tests. For example, see this [log file](https://pipelinesghubeus8.actions.githubusercontent.com/P0ELWR4zeACuLLNGFIvV2HtYX5YtHJnpF65NFez9JahqM2Rk03/_apis/pipelines/1/runs/308/signedlogcontent/4?urlExpires=2024-02-07T06%3A35%3A52.5187795Z&urlSigningMethod=HMACV1&urlSignature=8xjIzaw6%2FzRNPALNC8yjzwhEWlA%2FAE80B9eXLi6%2Bg0Y%3D) for [this failed job](https://github.com/hsutter/cppfront/actions/runs/7810216607/job/21303297024?pr=971).

This PR makes 2 changes:
- If Apple Clang 15 is detected it tries to switch to 14 instead
- Prints an error message and aborts the job if an unsupported compiler version is found

CC @jarzec

### DIFF
--- a/regression-tests/run-tests.sh
+++ b/regression-tests/run-tests.sh
@@ -130,8 +130,15 @@ if [[ "$cxx_compiler" == *"cl.exe"* ]]; then
     compiler_version=$(cl.exe)
 else
     compiler_cmd="$cxx_compiler -I../../../include -std=c++20 -pthread -o "
-    
     compiler_version=$("$cxx_compiler" --version)
+
+    # We don't currently support Apple Clang 15 so try and switch to 14
+    if [[ "$compiler_version" == *"Apple clang version 15.0"* ]]; then
+        printf "Found Apple Clang 15, attempting to switch to Apple Clang 14"
+        cxx_compiler=$(xcodebuild -find clang++)
+        compiler_version=$("$cxx_compiler" --version)
+    fi
+
     if [[ "$compiler_version" == *"Apple clang version 14.0"* ]]; then
         exec_out_dir="$expected_results_dir/apple-clang-14"
     elif [[ "$compiler_version" == *"clang version 12.0"* ]]; then 
--- a/regression-tests/run-tests.sh
+++ b/regression-tests/run-tests.sh
@@ -131,22 +131,22 @@ if [[ "$cxx_compiler" == *"cl.exe"* ]]; then
 else
     compiler_cmd="$cxx_compiler -I../../../include -std=c++20 -pthread -o "
     
-    compiler_ver=$("$cxx_compiler" --version)
-    if [[ "$compiler_ver" == *"Apple clang version 14.0"* ]]; then
+    compiler_version=$("$cxx_compiler" --version)
+    if [[ "$compiler_version" == *"Apple clang version 14.0"* ]]; then
         exec_out_dir="$expected_results_dir/apple-clang-14"
-    elif [[ "$compiler_ver" == *"clang version 12.0"* ]]; then 
+    elif [[ "$compiler_version" == *"clang version 12.0"* ]]; then 
         exec_out_dir="$expected_results_dir/clang-12"
-    elif [[ "$compiler_ver" == *"clang version 15.0"* ]]; then 
+    elif [[ "$compiler_version" == *"clang version 15.0"* ]]; then 
         exec_out_dir="$expected_results_dir/clang-15"
-    elif [[ "$compiler_ver" == *"g++-10"* ]]; then
+    elif [[ "$compiler_version" == *"g++-10"* ]]; then
         exec_out_dir="$expected_results_dir/gcc-10"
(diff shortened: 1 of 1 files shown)


gregmarr:
Is there a reason to not add support for Clang 15?

hsutter:
> incorrectly selects Apple Clang 15, which is currently unsupported as a compiler version for the regression-tests

Is there an issue with Apple Clang 15, or just that there's no subdirectory for Apple Clang 15 in the test results?

jarzec:
@hsutter In fact I wanted to propose adding support for Clang 15. I think the `run-test.sh` script has some leftovers of my trials for that. This version is available on Ubuntu 2204 and is installed by default on the GitHub runner. This is also the case on the macos-13 runner. The added value of adding this version is that all regression tests can be compiled and run using it, so it would be the only POSIX compiler allowing for that (at least that was the case last time I tried). Can quickly prepare a PR with that.

jarzec (approved):

regression-tests/run-tests.sh
```diff
@@ -130,31 +130,39 @@ if [[ "$cxx_compiler" == *"cl.exe"* ]]; then
     compiler_version=$(cl.exe)
 else
     compiler_cmd="$cxx_compiler -I../../../include -std=c++20 -pthread -o "
-    
-    compiler_ver=$("$cxx_compiler" --version)
-    if [[ "$compiler_ver" == *"Apple clang version 14.0"* ]]; then
+    compiler_version=$("$cxx_compiler" --version)
+
+    # We don't currently support Apple Clang 15 so try and switch to 14
+    if [[ "$compiler_version" == *"Apple clang version 15.0"* ]]; then
```
jarzec, line 136: [Question] Are you sure Apple Clang 15 was selected? The [GitHub runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md) states that Clang 15 s installed from Homebrew. I checked and the output of `$(brew --prefix llvm@15)/bin/clang` is:
```
Homebrew clang version 15.0.7
Target: x86_64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /usr/local/opt/llvm@15/bin
```

bluetarpmedia: No, I'm not 100% sure, this is very much some investigative work!

All I can say for certain right now is that on some macOS job runs (e.g. the one linked above, and I've seen it other times too), `clang++` does not resolve to Apple Clang 14. That's why the script produces the `not found for compiler: 'clang++'` error message.

One of the updates in this PR is to provide more info about why the job fails. If this PR is merged then in the future we'll get more info due to this:
```
printf "Unhandled compiler version:\n$compiler_version\n\n"
```
That may show `Apple clang 15.0 etc` or something else.

Yeah, the `macos-13` runner confusingly has both LLVM Clang installed as well as multiple versions of Xcode (with Apple Clang).

image

Xcode 14.3 is meant to be the default so `clang++` would resolve to Apple Clang 14, but I suspect there's a problem which is causing Xcode 15 to be active. But, it could also somehow be invoking LLVM Clang 15. To me that seemed less likely because, like you noted, it's installed via Homebrew and needs the `brew --prefix llvm@15` to get the correct path.

bluetarpmedia:
> Is there a reason to not add support for Clang 15?

Nope, that's definitely an option. (Note there's Clang 15 as well as Apple Clang 15.)

bluetarpmedia:
> Is there an issue with Apple Clang 15, or just that there's no subdirectory for Apple Clang 15 in the test results?

Yeah, the latter, because there are no expected test-results yet for Apple Clang 15.

Even if we add test-results for Clang 15 and Apple Clang 15, I think it's still worth adding this part of the PR:

```
printf "Unhandled compiler version:\n$compiler_version\n\n"
```

That way in the future we'll get more details when the job fails due to a new compiler on the GitHub runner that we don't yet have test-results for.

hsutter:
@jarzec I see an "approved these changes" from you midway up above, just checking: Does this look ready to merge to you?

jarzec:
> @jarzec I see an "approved these changes" from you midway up above, just checking: Does this look ready to merge to you?

Yes, please do merge. I've started to look into including Clang 15 on top of the updates here.
(Sorry, I previously responded from my work account).

hsutter:
Thanks!

PR #974, opened 2024-02-09
CI: Add Clang 15 support in regression tests

jarzec:
Add support for Clang 15 in regression tests.
  - Include the Homebrew version of Clang 15 as supported compiler.
    - The test results are identical as for Apple Clang 14 - the existing `test-results/apple-clang-14` directory can be reused.
  - Switch the default C++ standard used for regression test compilation to `c++2b`. Stick to `c++20` for:
    - GCC 10 that simply does not support `c++2b`. Due to this lack of support this version might possibly be dropped leaving only GCC 13.
    - Clang 15 on Ubuntu gives strange issues with `c++2b` - to be investigated further.

NOTE:
  - The switch to `c++2b` makes the lowered `pure2-assert-expected-not-null.cpp2` compile for GCC 13.
  - The result dir `test-results/clang-15` added by this PR differs from `test-results/apple-clang-14` only for:
    - `mixed-captures-in-expressions-and-postconditions.cpp2`
    - `pure2-type-and-namespace-aliases.cpp2`
  - While preparing the PR I was able to confirm the spurious selection of Apple clang version 15.0.0 in some runs. Unfortunately the solution using `xcodebuild -find clang++` doesn't seem to help as can be seen in [this build](https://github.com/jarzec/cppfront/actions/runs/7850439073/job/21425802250) that tested another approach to use that command. It seem that is such cases XCode 15 is used by default. ~~The problem will require further investigation or a fix of the GitHub `macos-13` runner. Re-running jobs failing due to this issue will remain a workaround.~~ I've added a command to select Xcode 14. This should hopefully solve the issue.

REMARK: Ubuntu 2404 will likely come out with Clang up to v18 and GCC up to v14 available directly from apt, without the need to install them manually (see [here](https://packages.ubuntu.com/noble/devel/)). This might be a good point to start looking into including newer compilers with better C++23 support and possibly dropping those without any (namely GCC 10).


.github/workflows/regression-tests.yml
```diff
@@ -17,18 +17,26 @@ jobs:
...
           - os: windows-latest
             compiler: cl.exe
     steps:
     - name: Checkout repo
       uses: actions/checkout@v4
 
+    - name: Prepare compilers
+      if: matrix.os == 'macos-13'
+      run: |
+        sudo xcode-select --switch /Applications/Xcode_14.3.1.app
```
jarzec, line 37: [Explanation] The path is taken from [here](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md).

hsutter:
Thanks! I was about to hit Merge but I see there are more commits -- LMK when this is ready.

regression-tests/run-tests.sh
```diff
@@ -129,31 +129,39 @@ if [[ "$cxx_compiler" == *"cl.exe"* ]]; then
...
-        compiler_version=$("$cxx_compiler" --version)
+    # Verify the compiler command
+    which "$cxx_compiler" > /dev/null
+    if [[ $? != 0 ]]; then
+        printf "The compiler '$cxx_compiler' is not installed\n\n"
+        exit 2
     fi
 
-    if [[ "$compiler_version" == *"Apple clang version 14.0"* ]]; then
+    cpp_std=c++2b
```
JohelEGP, line 139: The tests should be run with C++20.
This makes it not do so, right?
We can't even have tests for C++23 (e.g., see commit 76f6d52c0ec9b9690bbac60ec829d3ccc133f68c).

jarzec: Oh, sorry, I didn't know. The confusion came from the fact that `pure2-assert-expected-not-null.cpp2` uses `std::expected` which is a C++-23 feature. I also saw PRs adding more support for `std::expected` to cppfront.
All tests currently only work for MSVC which uses `c++latest`. This seems to mean C++23 (at least for `std` implementation) given that the test with `std::expected` does compile. As I mentioned it the PR description using `c++2b` makes this test compile and work for GCC 13 as well.
As you can see from the fact that all the GitHub actions succeed, `c++2b` works for all compilers on Ubuntu 2204 for the current tests. Only GCC 10 doesn't support that option. Clang 15 also has problems with this option but this seems to be a GitHub runner issue.
Given that it is 2024 already and C++26 is WIP I think it makes a lot of sense to start using C++23 features for compilers that support them and look into compilers that support even more of them.
This PR allows to select the C++ standard used for each compiler.
It is already the case that compilation issues are committed as expected test results for certain compilers. The test scripts lists all such cases at the end. In fact, currently no compiler on POSIX platforms can compile the lowered code for all the tests. I would argue it will be very hard for a future-oriented project like cppfront to either only support older features or support newer features but not to have tests of that support.
All that being said, I believe this is @hsutter's call in the end.

jarzec: @hsutter I you are OK with compiling all regression tests with `c++2b` for compilers that support that this PR is ready for merging, I you prefer to revert to `c++20` I can do that.

hsutter:
Great, thanks again.

PR #975, opened 2024-02-12
Make regression tests only run on `main`

hsutter:
I think this is the right way to express they should run on `main` and PRs on `main`... it doesn't need to run on the `docs` branch.

@jarzec does this diff look right to accomplish that?

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -2,8 +2,12 @@ name: Regression tests
 
 on:
   pull_request:
+    branches:
+      - main
     types: [opened, synchronize, reopened]
   push:
+    branches:
+      - main
   workflow_dispatch:
 
 jobs:
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@
 *.exe
 source/gen_version.bat
 build*/
-docs/venv/*
+venv/*
 *.ifc
 
 # Visual Studio cache directory


jarzec:
LGTM

jarzec (approved):

.github/workflows/regression-tests.yml
```diff
@@ -2,8 +2,12 @@ name: Regression tests
 
 on:
   pull_request:
+    branches:
+      - main
     types: [opened, synchronize, reopened]
```
gregmarr, line 7: The `types` should not be needed here.

>if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

gregmarr:
That does mean what you described.  Whether or not that is what is actually desirable is debatable.  Since you're the only one working directly in this repo, and you only work on `main`, then I think that does make sense.  If there were other people working on other branches, then you'd want to include those other branches as well.

hsutter:
Thanks!

Note that I copied this from the compiler test workflow, so this is making those two workflows consistent.

The immediate issue I'm trying to debug why MkDocs's workflow (which I've also called `docs`, maybe I'm creating an overload problem?) isn't working in CI when it works fine locally on my box, and although I suspect a different Pages workflow might be interfering I'm trying to eliminate this as a possible cause. Secondarily, even though the `docs` branch currently has a copy of all the code, I'm only making `/docs` changes there so it's kind of a waste to run regressions there too.

jarzec:
You can in fact just exclude the `docs` branch (if that is what is intended) as follows:
```
on:
  pull_request:
    branches-ignore:
      - docs
  push:
    branches-ignore:
      - docs
```

jarzec:
> You can in fact just exclude the `docs` branch (if that is what is intended) as follows:
> 
> ```
> on:
>   pull_request:
>     branches-ignore:
>       - docs
>   push:
>     branches-ignore:
>       - docs
> ```

@hsutter Would it be fine for you to go with my proposition of excluding `docs`? @gregmarr has a point, e.g. when preparing PRs it is very useful to easily run regression tests on branches other than `main` in forked repos. I did that a lot.

hsutter:
@jarzec OK, done with 0d40c688c3dc127e8c6bb180a34b293cabf5921e -- please check?

jarzec:
@hsutter The update seems to work. However you still have regression tests running on your `docs` branch. The reason is that the workflow file on that branch is still an old one. I would suggest rebasing `docs` on current `main`. I tried this in my fork with the expected result - regression tests are not run on [my rebased version of `docs`](https://github.com/hsutter/cppfront/compare/main...jarzec:cppfront:docs) anymore.
There was just one trivial conflict in `.gitignore` during the rebase.

hsutter:
Fixed, thanks!

PR #980, opened 2024-02-16
CI: Unify execution conditions of test workflows

jarzec:
Unifies the execution conditions of the two testing workflows:
 - `.github/workflows/build-cppfront.yaml`
 - `.github/workflows/regression-tests.yml`

Adds exclusion of the `docs` branch instead of limiting the tests to the `main` branch only.
Adds the [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) event for the `.github/workflows/build-cppfront.yaml` (already active in `.github/workflows/build-cppfront.yaml`). This allows [manual execution of the workflow](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) through the GitHub interface as soon as the workflow definition is included in the `main` branch.

After the proposed change both testing workflows will execute automatically on development branches used e.g. for preparing PRs.

This change is another follow-up of https://github.com/hsutter/cppfront/pull/975.

### DIFF
--- a/.github/workflows/build-cppfront.yaml
+++ b/.github/workflows/build-cppfront.yaml
@@ -1,11 +1,13 @@
 name: Multi-platform Build of cppfront
 on:
-  push:
-    branches:
-      - main
   pull_request:
-    branches:
-      - main
+    branches-ignore:
+      - docs
+  push:
+    branches-ignore:
+      - docs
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks!

PR #983, opened 2024-02-17
Another round of docs updates

gregmarr:


DyXel (approved):

docs/cpp2/types.md
```diff
@@ -213,6 +218,8 @@ mytype: type
         print();
     }
 
+    // TODO skipping the default memberwise assignment for later setting
```
hsutter, line 221: Thanks! My only question is, what does this comment mean?

Otherwise this all looks good, thanks again.

hsutter:
Thanks!

docs/cpp2/types.md
gregmarr: @hsutter Including an example of using the `member = _;` to skip the memberwise assignment so that you can assign to it later in the function.

hsutter: Ah, got it -- I'll re-add the TODO comment, thanks.

PR #984, opened 2024-02-19
Update CTAD description in hello-world.md

bluetarpmedia:
The original text seemed a bit circular:
>Declaring `words` uses CTAD (...) to declare its `words` variable.

So I've updated it to hopefully reflect the intention.


hsutter:
Thanks!

PR #993, opened 2024-02-22
Update source paths in README

danra:


README.md
```diff
@@ -74,15 +74,15 @@ Cppfront builds with any major C++20 compiler.
 
 #### MSVC build instructions
 
-    cl cppfront.cpp -std:c++20 -EHsc
+    cl source\cppfront.cpp -std:c++20 -EHsc
```
gregmarr, line 77: backslash instead of forward slash

danra:
It's probably preferrable for the binary to be built outside of the source directory. So just specifiying the path makes sense to me. It's also less "explaining" and more "just works" than the alternative of instructing to change into the source folder.

gregmarr:
On the other hand, I'd hate to imply that you HAVE to build it from the root of the install if it can be built from inside the directory.

danra:
> On the other hand, I'd hate to imply that you HAVE to build it from the root of the install if it can be built from inside the directory.

I don't think that's implied... these are instructions for first-time comers to the repo. Having it "just work" makes it just a bit easier, probably saving approx. a minute.

hsutter:
Thanks!

PR #998, opened 2024-02-26
Changes to functions.md

bluetarpmedia:
- Fix small typos (singular/plural)
- Suggest wording changes
- Added a new example (to appear first) for functions with multiple/named return values

The existing `insert` example declares a member function in a `set` type, but the reader may not have encountered types yet if they're following the guide in order.


hsutter:
Thanks!

PR #1005, opened 2024-02-29
Update aliases.md

bluetarpmedia:
- Fix small typos
- Suggested change `assert` to `static_assert`. In these examples I think the programmer would prefer a compile-time assert, but I understand if you prefer to highlight Cpp2's violation handler behaviour.


hsutter:
Thanks!

PR #1008, opened 2024-03-07
used std::variant instead of union in cpp2::out

ARG-Helper:
I performed a blind replacement of the union in cpp2::out with a variant in order to demonstrate a bug in the original implementation. This branch should cause some tests to fail

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -754,7 +754,6 @@ class out {
     //  Actually using std::variant here
     std::variant<T*,deferred_init<T>*> vt;
     out<T>* ot = {};
-    bool has_t;
 
     //  Each out in a chain contains its own uncaught_count ...
     int  uncaught_count   = Uncaught_exceptions();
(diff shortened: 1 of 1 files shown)


ARG-Helper:
example code to demonstrate bug:
f1: (out x: _) = {
    f2(out x);
    x = 42;
}
f2: (out x: _) = {
    x = 13;
}

main: ()->int = {
    x:i32;
    f1(out x);
    return x;

}

ARG-Helper:
my old example does not actually work new example
inline void f1(cpp2::out<int> in){
    in.construct(1)
    in.construct(2)
}
int main(){
cpp2::deferred_init<int> x; 
f1(x);
return x.value();
}

hsutter:
Thanks! Adding xref to this comment: https://github.com/hsutter/cppfront/issues/1012#issuecomment-1986138995

I don't think this bug can happen unless the types are used directly. Cppfront never generates those code patterns.

Does that help answer the question?

Again, thanks!

hsutter:
Thanks again. I think the code is fine as long as it's used from cppfront-generated code, which we know will never call `construct` twice.

However, this has helped me see that I need to clarify that some things in `cpputil.h`) are, and others aren't, intended for users to use directly. So I've put the latter into an `impl` namespace in this commit: 7fd706de88b57857af4d5a31c8dc5429535d367f

Thanks!

PR #1009, opened 2024-03-07
Fixed cpp2::out

ARG-Helper:
example fix for the bug demonstrated in #1008


ARG-Helper:
fix does not work if `alignof(T)<alignof(bool)`

PR #1014, opened 2024-03-08
removed harmfull check in out<T>

### LINKED ISSUE #1012: [BUG] read from inactive union member in cpp2::out.construct()
**Describe the bug**
cpp2::out.construct() may read from the inactive member of its anonymus union when called more than once

**To Reproduce**
`int main(){
   cpp2::deferred_init<int> a;
   cpp2::out o{&a};
   o.construct(1);
   o.construct(2);
}`

**Additional context**
the variable initialisation rules of cpp2 ensure that construct is only ever called once for each cpp2::out chain

turning `cpp2::deferred_init<T>*` into a `T*` to it's contained T is doable in a standart compliant way using `reinterpret_cast<T*>()` followed by `std::launder()`

ARG-Helper:
fixes #1012 by removing the check that ignored the invariant 
`called_construct()` implies `dt->init`

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -788,7 +788,7 @@ public:
     }
 
     auto construct(auto&& ...args) -> void {
-        if (has_t || called_construct()) {
+        if (has_t) {
             if constexpr (requires { *t = T(CPP2_FORWARD(args)...); }) {
                 Default.enforce( t );
                 *t = T(CPP2_FORWARD(args)...);


hsutter:
Thanks! I think the check is necessary... the reason it exists is because of the case when an uninitialized variable is initialized via a chain of calls through `out` parameters. It's possible I could restructure the code to use just the flag already inside `deferred_init`, in the future.

However, this has helped me see that I need to clarify that some things in `cpputil.h`) are, and others aren't, intended for users to use directly. So I've put the latter into an `impl` namespace in this commit: 7fd706de88b57857af4d5a31c8dc5429535d367f

Thanks!

PR #1019, opened 2024-03-13
Correct the feature test macro for including `syncstream` header

bluetarpmedia:
Verified that `__cpp_lib_syncbuf` is the correct macro by checking:
- Repro on [Godbolt](https://godbolt.org/z/rvajMMYPK) for 3 major compilers
- https://en.cppreference.com/w/cpp/feature_test


hsutter:
Thanks!

PR #1036, opened 2024-03-20
Fix switched columns in common.md

6he:


hsutter:
Thanks for pointing this out! Actually that whole row is an error because there's no binary `~`, that was a thinko. Fixed via a separate commit.

PR #1044, opened 2024-03-29
Create CITATION.cff

dutkalex:
Adds a CITATION.cff file to the repo to ease citing in research papers and technical reports


hsutter:
Thanks! Sorry for the delay, this came in just before I started a gone-dark month of travel/training/other commitments in April and I'm starting to catch up now.

PR #1045, opened 2024-03-29
Fixes for arm-none-eabi-g++

matgla:
Typeid cannot be used when -fno-rtti is used.
Fixed compilation of generated sources on arm-none-eabi-g++ 13.1.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -654,7 +654,7 @@ constexpr auto type_name() -> std::string_view {
 
 [[noreturn]] auto Throw(auto&& x, [[maybe_unused]] char const* msg) -> void {
 #ifdef CPP2_NO_EXCEPTIONS
-    auto err = std::string{"Throw exception (-fno-exception) with type \""};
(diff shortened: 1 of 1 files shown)


matgla:
Hi, 
I just wanted to raise issue that fix is necessary for -fno-rtti + -fno-exception builds. 
Usage of __PRETTY_FUNCTION__ and MSVC __FUNCSIG__ is a little bit hacky I think, so maybe it's better to print just exception without type. 

arm-none-eabi-g++ also fails on std::string to const char* conversion. 

Gotbolt link with example usage compile time type_name : https://godbolt.org/z/azoY9Mc55

hsutter:
Thanks! That's quite cute.

It looks like this may be your first contribution to cppfront. I've sent the Contributor License Agreement (CLA) to your email, and once it's signed I can look in more detail at your pull request. Thanks again for your contribution.

hsutter:
Looks, good, thanks!

PR #1053, opened 2024-04-15
Extension of is() part1: free function predicates, fix implicit cast for build-in and custom is operator, refactoring

filipsajdak:
This is the first part of `is()` rework originally done in #701 .

It contains:

* handling of free function predicates,
* fix of implicit casts for build-in is() and custom is() operator,
* rework of matching variable to template - adjust to new style,


hsutter:
I just took a pass over it and fixed a couple of small things -- looks good! Thanks.

PR #1064, opened 2024-05-11
Fix regressions tests

jarzec:
Fix the regression tests by updating the content the tests data files for a number of tests for `apple-clang-14`, `apple-clang-15` and `clang-12`, `clang-15` and `gcc-13`.

In addition, excludes `cppfront` build on macos-11 in the **Multi-platform Build of cppfront**.
`cppfront` stopped building on that runner due to a number of features missing in the `std` implementation of that system.

The PR fixes all CI test as of ecc62d7e9f32825f385b6e58077e02243c990712.


hsutter:
Thanks!

PR #1070, opened 2024-05-15
Bugfix for nonconst has in generated flag_enums.

MaxSagebaum:
In flag_enums the `has` function is marked as `inout` and not as `in`.


MaxSagebaum:
@hsutter This is a small fix for the flag_enum.

hsutter:
Thanks!

PR #1077, opened 2024-05-22
[CI] Fix `clang++-15` regression test job

bluetarpmedia:
Updated expected results for `clang++-15` on `ubuntu-latest`.

### DIFF
--- a/regression-tests/test-results/clang-15/pure2-assert-expected-not-null.cpp.output
+++ b/regression-tests/test-results/clang-15/pure2-assert-expected-not-null.cpp.output
@@ -5,7 +5,7 @@ pure2-assert-expected-not-null.cpp2:7:10: error: no member named 'expected' in n
     std::expected<int,bool> ex {4}; 
     ~~~~~^~~~~~~~
          unexpected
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks! But a question: Clang-15 no longer works with formatted interpolation?

bluetarpmedia:
> Thanks! But a question: Clang-15 no longer works with formatted interpolation?

That's because it's now using the libstdc++ from gcc 12. Previously it was using libstdc++ from gcc 13. So it's not really a clang compiler issue but the version of the standard library that it uses.

Something changed with the GitHub Action ubuntu images which means gcc 13 is no longer available on `ubuntu-latest`, so clang now picks up the highest version of gcc that is installed (12) and uses that version of libstdc++.

I figured the best option for now is to fix this regression test job, but I've made a note to improve things:

1. Stop using `ubuntu-latest` since over time this "pointer" will change, and we won't have consistent results. Instead I think it's better to use explicit versions

2. I think we need to add tests for clang and libc++ (the LLVM standard library), instead (or in addition to) clang with libstdc++.

hsutter:
Sounds good, thanks for paying attention to this! Appreciated.

PR #1078, opened 2024-05-22
[CI] Replace `ubuntu-latest` with an explicit ubuntu version

bluetarpmedia:
`ubuntu-latest` is an alias which over time changes and points at newer ubuntu images. Different ubuntu images have different compiler versions installed and newer images drop older compiler versions.

So to prevent the CI jobs from failing when the `ubuntu-latest` alias changes I've modified the workflow to use a specific ubuntu version for the `g++-10` and `clang++-15` jobs, and this `ubuntu-22.04` image is [documented](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) to have gcc 10 and clang 15 installed.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks!

PR #1079, opened 2024-05-22
[Suggestion] Add Ubuntu tests for GCC 14 and Clang 18

jarzec:
Now that Ubuntu 24.04 LTS is out and a respective GitHub runner is already available it seems to make a lot of sense to add GCC 14 and Clang 18 to the test suite. Both compilers are installed by default on the `ubuntu-24.04` GitHub runner.

With C++26 in sight it might also be the time to drop tests (not yet done in this PR):
 - on `ubuntu-20.04`,
 - for GCC 10,
 - for Clang 12.

Thoughts anyone?


.github/workflows/regression-tests.yml
```diff
@@ -19,13 +19,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
-        compiler: [g++-10, clang++-15]
```
hsutter, line 23: I think this is where GCC 10 support is removed? I'd like to keep it please (see longer comment in main thread).

hsutter:
Thanks! Quick check:

> With C++26 in sight it might also be the time to drop tests (not yet done in this PR):
> 
> * on `ubuntu-20.04`,
> * for GCC 10,
> * for Clang 12.

Actually this PR does remove GCC 10 for the regression tests, right?

I'd like to keep GCC 10 and Clang 12 as a baseline, because right now they do work (with only a few exceptions that should all have diagnostics), and I'd like to know when we come to the point where something in the code gen causes generated code to no longer be compatible with them. (FWIW, I do my own local testing with GCC 10 and 13, and Clang 12.)

.github/workflows/regression-tests.yml
jarzec: This is just changing the defaults. GCC 10 is kept on Ubuntu 20.04 in new lines 29 - 30. It might be possible to install that version on the newer Ubuntu, but I don't this there will be much difference in the test execution. I can check this, though.

jarzec:
> Thanks! Quick check:
> 
> > With C++26 in sight it might also be the time to drop tests (not yet done in this PR):
> > 
> > * on `ubuntu-20.04`,
> > * for GCC 10,
> > * for Clang 12.
> 
> Actually this PR does remove GCC 10 for the regression tests, right?
> 
> I'd like to keep GCC 10 and Clang 12 as a baseline, because right now they do work (with only a few exceptions that should all have diagnostics), and I'd like to know when we come to the point where something in the code gen causes generated code to no longer be compatible with them. (FWIW, I do my own local testing with GCC 10 and 13, and Clang 12.)

As I mentioned in the earlier comment I did keep all the compilers used so far. You can confirm that in the list of checks ([GCC 10](https://github.com/hsutter/cppfront/actions/runs/9198621472/job/25301721278?pr=1079) and [Clang 12](https://github.com/hsutter/cppfront/actions/runs/9198621472/job/25301721012?pr=1079)).

.github/workflows/regression-tests.yml
hsutter: D'oh! Right you are, sorry. I didn't look at the rest of the changes carefully enough.

hsutter:
Thanks! I overlooked it, sorry.

PR #1080, opened 2024-05-23
[CI] Regression test jobs now include the C++ standard version and (if applicable) the standard library to link with

bluetarpmedia:
* Jobs now specify the C++ standard version to compile for, and (if applicable) the standard library to link with
* Add a new job to test clang with `libc++` on ubuntu. (All other ubuntu jobs for gcc & clang link with `libstdc++`.)
* Add a new msvc job so that we now test both `c++latest` and `c++20`.
* Reorder jobs so that GCC are first, then clang on ubuntu & macOS, and then msvc on Windows

Job names now have this format:
`<short os name> | <compiler> | <c++ std> | <std library> | <full os name>`

The summary job list shows them like this (which unfortunately still crops the end):
image

Full name when you click into a job:
image

The `test-results` directories now include the C++ standard and, only for libc++ on ubuntu, the standard library name.

image


jarzec (approved):
LGTM
Cool, thanks! I was thinking of adding std lib version but never had time.
BTW as usual https://github.com/hsutter/cppfront/pull/1080/files is not very responsive for big diffs and freezes after some time.
I suggest reviewing in VSCode through https://github.dev/hsutter/cppfront/pull/1080/files.

regression-tests/test-results/msvc-2022-c++20/run-tests-msvc-2022.bat
```diff
@@ -1,9 +1,9 @@
-REM This is intended to be run in the /test-results/msvc-2022 directory
+REM This is intended to be run in the /test-results/msvc-2022-c++20 directory
 REM in a VS2022 Command Prompt (which sets the path and other environment)
 REM
 REM After installing VS2022, to finish setting up std modules in your environment, run:
-REM     cl /std:c++latest /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.ixx"
-REM     cl /std:c++latest /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.compat.ixx"
+REM     cl /std:c++20 /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.ixx"
+REM     cl /std:c++20 /EHsc /nologo /W4 /MD /c "%VCToolsInstallDir%\modules\std.compat.ixx"
```
jarzec, line 6: I overlooked this change in my first review. It is OK with me, but @hsutter please confirm this is something you want. I guess you use this script a lot, don't you?

bluetarpmedia: That change is in the new msvc `c++20` directory. The original bat file is unchanged but its directory now has `c++latest` in its name.

JarekGlobus: Sorry, my bad.

jarzec: Sorry, my bad.

PR #1088, opened 2024-05-29
Default-define out-of-line dtors for types with fwd-declared unique_ptr members

jarzec:
Some of the node types in `parse.h` have `std::unique_ptr` of forward-declared types as members without having destructors deferred to when the pointed-to types are defined. If the default deleter is used this is [IFNDR](https://en.cppreference.com/w/cpp/language/ndr). Despite that, all compilers seem to have been compiling such code without problem. Only recently Clang >=15 with C++23 enabled started to raise errors in such situations. This is likely due to the added `constexpr` in the default deleter, which seemed to have change changed implementation decisions.

This PR adds out-of-line defaulted destructors for such cases to fix the possible compiler errors. It also enables a build test for Clang 18 with C++23 to ensure the solution works.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -141,6 +141,9 @@ struct primary_expression_node
     // Cache to work around <https://github.com/llvm/llvm-project/issues/73336>.
     bool expression_list_is_fold_expression = false;
 
+    // Out-of-line definition of the dtor is necessary due to the forward-declared
+    // type(s) used in a std::unique_ptr as a member
+    ~primary_expression_node();
 
     //  API
     //
@@ -234,6 +237,10 @@ struct prefix_expression_node
     std::vector<token const*> ops;
     std::unique_ptr<postfix_expression_node> expr;
 
+    // Out-of-line definition of the dtor is necessary due to the forward-declared
+    // type(s) used in a std::unique_ptr as a member
+    ~prefix_expression_node();
+
     //  API
     //
     auto is_fold_expression() const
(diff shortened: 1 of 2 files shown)


source/parse.h
```diff
@@ -1090,6 +1101,18 @@ struct template_argument
         std::unique_ptr<type_id_node>
     > arg;
 
+    // The type needs to be movable
+    // The copy ctor+operator are implicitly deleted due to the std::unique_ptr member
+    // Because a forward-declared type is used in a std::unique_ptr as a member an out-of-line dtor is necessary
+    // Because of the OOL dtor together with the fact that the copy ctor+operator are deleted
+    // the move ctor+operator need to be explicitly defaulted
+    // As a result the default constructor also needs to be explicitly defaulted
```
jarzec, line 1109: This is an interesting side-effect of the changes.

source/parse.h
```diff
@@ -4389,6 +4428,26 @@ struct translation_unit_node
     }
 };
 
+// Definitions of out-of-line dtors for nodes with unique_ptr members of forward-declared types
```
jarzec, line 4431: I've decided to add the deferred dtor definitions here, at the point where (to the best of my knowledge) all node types are already defined.

jarzec:
@hsutter The PR with the discussed updates is ready. For some reason, when I created it GitHub runners had a bad time and some testes were cancelled/skipped. The changes are OK as you can see from the [tests of the same commit in my own fork](https://github.com/jarzec/cppfront/actions/runs/9291136161). You can also rerun the failed steps ones manually for the PR in the Actions section.

jarzec:
@hsutter I have rebased this branch on your current main to get all the CI tests to pass.

source/parse.h
hsutter: Good old SMF defaults 😁 Thanks!

hsutter:
Thanks!

PR #1092, opened 2024-05-31
Default initialize function argument, closes #1090

### LINKED ISSUE #1090: [BUG] Cannot value initialise a function argument with equivalent of C++ `{}`
**Describe the bug**
I can't find a Cpp2 syntax to value initialise an argument passed to a function using its default constructor.

**To Reproduce**
Run cppfront on this code:
```cpp
log: (message: std::string_view,
      context: std::string,
      event_id: int) =
{
    std::println("{} {} {}", message, context, event_id);
}

main: () = {
    log("event", (), 123);  // (1) Error in C++ compiler
    log("event", _, 123);   // (2) Error in C++ compiler
    log("event", {}, 123);  // (3) Error in cppfront
}
```

(1) produces this error:
```
main.cpp2:10:19: error: expected expression
   10 |     log("event", (), 123);
      |                   ^
```

(2) produces this error:
```
main.cpp2:10:18: error: use of undeclared identifier '_'
   10 |     log("event", _, 123);
      |                  ^
```

(3) produces a cppfront syntax error.

[Repro](https://cpp2.godbolt.org/z/MM5zcMP56) on Godbolt

**Additional context**
I was [translating](https://cpp2reference.com/algorithm/ranges/find_end/) the `ranges::find_end` [example](https://en.cppreference.com/w/cpp/algorithm/ranges/find_end) from cppreference and have used explicit arguments like this:

```cpp
found4:= std::ranges::find_end(
    secret, "SWORD"sv,
    std::ranges::equal_to(),
    std::identity(),
    :(c: char) std::tolower(c));
```
instead of the original:
```cpp
const auto found4 = std::ranges::find_end(secret, "SWORD"sv,
        {},
        {},
        [](char c) { return std::tolower(c); });
```

sookach:
@hsutter @bluetarpmedia 
This patches #1090 by allowing defaulting of arguments in a function call by use of '()' (akin to '{}' in C++).


MaxSagebaum:
Looks good, a regression test would be nice.

sookach:
> Looks good, a regression test would be nice.

Yeah, it passed on my fork
https://github.com/sookach/cppfront/tree/pr/default_arg

but it looks like the tests don't run on their own in this repo?

gregmarr:
> but it looks like the tests don't run on their own in this repo?

https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
> By default, all first-time contributors require approval to run workflows.

MaxSagebaum:
>>  Looks good, a regression test would be nice.
>
>Yeah, it passed on my fork
>https://github.com/sookach/cppfront/tree/pr/default_arg
>
> but it looks like the tests don't run on their own in this repo?

Ah, what I meant, is that it would be nice if you add a regression test for the new feature. Sorry for the misunderstanding.

sookach:
Cool, added a regression test. Let me know what you think.

source/parse.h
```diff
@@ -5899,8 +5903,12 @@ class parser
                 //  Next should be an expression-list followed by a ')'
                 //  If not, then this wasn't a call expression so backtrack to
                 //  the '(' which will be part of the next grammar production
+		        is_inside_call_expr = true;
 
                 term.expr_list = expression_list(term.op, lexeme::RightParen);
+				
+		        is_inside_call_expr = false;
+
```
MaxSagebaum, line 5911: I thought, that it might be better to add `is_inside_call_expr` as an argument. But I think from the geeneral design it is not possible.  So the state modification is ok.  In order to make the more prominent, I would changed it to:

```suggestion

                is_inside_call_expr = true;  // Flag this expression list parse.
                term.expr_list = expression_list(term.op, lexeme::RightParen);
                is_inside_call_expr = false;

```

MaxSagebaum (commented):
Nice, thanks for adding the test.  It looks good to me. I just have a minor comment.

source/parse.h
sookach: That makes sense. Done.

MaxSagebaum (approved):

source/parse.h
hsutter: Right, it's either use out-of-band state, or wire a parameter through the call chain. Sometimes I do the former, sometimes the latter.

hsutter:
Thanks!

PR #1095, opened 2024-06-05
Update test-results for ubuntu clang 12, gcc 13 and gcc 14

bluetarpmedia:
Update test-results for ubuntu clang 12, gcc 13 and gcc 14 after changes to `cpp2util.h` from d13615317734b611042dcba5af623611bcc054f2


jarzec (approved):

hsutter:
Thanks!

PR #1101, opened 2024-06-09
Add `..` syntax to select a member function only

hsutter:
With this PR, `x.f()` is still UFCS, but `x..f()` will now find only a member function.

For now I like that the default is UFCS, but as we get more experience I'm open to changing the default so that `.` is member selection and `..` is UFCS (which would be a breaking change, but a mechanical one).

Also: Remove the old version of `get_declaration_of` now that the new lookup seems stable.

### DIFF
--- a/source/sema.h
+++ b/source/sema.h
@@ -419,26 +419,6 @@ public:
     ) const
         -> declaration_sym const*
     {
-        //--------- START TEMPORARY REGRESSION TEST CODE FOR G_D_O OPTIMIZATION VERIFICATION ---------
-        auto result_old = static_cast<declaration_sym const*>(nullptr);
-        if (flag_internal_debug)
-        {
-            auto debug = false; // t == "xxx" && t.position().colno > 20;
-            if (debug)
-            {
-                std::cout << "\nlooking up " << t.as_string_view() << " (" << static_cast<void const*>(&t) << ", " << look_beyond_current_function << ", " << include_implicit_this << ") from " << t.position().to_string();
-                std::cout << " ... symbol table state is:\n";
-                debug_print(std::cout);
-            }
-
-            result_old = get_declaration_of_old(t, look_beyond_current_function, include_implicit_this );
-
-            if (debug) {
-                std::cout << "\n  found -> " << static_cast<void const*>(result_old);
-            }
-        }
-        //--------- END TEMPORARY REGRESSION TEST CODE FOR G_D_O OPTIMIZATION VERIFICATION -----------
-
         //  Calculate result using declaration_of[]
         auto result = static_cast<declaration_sym const*>(nullptr);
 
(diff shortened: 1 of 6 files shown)


jarzec:
@hsutter It looks like builds/regression tests on some platforms fail due to warnings that become errors. Fixing the problems should not be difficult. Would you like some help with that?
In any case, I would suggest merging #1095 first as it fixes some pre-existing regression test issues.

source/to_cpp1.h
```diff
@@ -272,6 +272,9 @@ class positional_printer
     )
         -> void
     {
+if (s.starts_with("..")) {
+    s == s;
```
jarzec, line 276: This is causing "unused" warnings which lead to errors in the tests on some platforms. Is the `if` statement actually necessary?

jarzec: I've just checked that this is the reason for all failures among the multi-platform build tests.
The failures in the regression tests seem to be mainly due to necessary line updates in error messages. This should be quick to fix, but first the `s == s` needs to be fixed as this might further change line numbers.

hsutter: Ah, I should remove that... that was so I could set a debug breakpoint on 276 that would fire only when `..` was happening, to investigate why `..` wasn't yet lowering to `.` correctly.

jarzec: OK, so let me get the tests running. I will create a pull request into this branch.
Please give me a couple of minutes.

jarzec: Sorry, it took a bit longer than a couple of minutes, but here is the PR into this one: #1103.

source/sema.h
```diff
@@ -502,228 +482,36 @@ class sema
...
 
-            debug_print(out);
+        //    debug_print(out);
 
-            exit(EXIT_FAILURE);
-        }
-        //--------- END TEMPORARY REGRESSION TEST CODE FOR G_D_O OPTIMIZATION VERIFICATION -----------
+        //    exit(EXIT_FAILURE);
+        //}
+        ////--------- END TEMPORARY REGRESSION TEST CODE FOR G_D_O OPTIMIZATION VERIFICATION -----------
```
jarzec, line 510: I guess this was commented out for testing and can be removed now?

```suggestion
```

source/to_cpp1.h
hsutter: Thanks! This PR shouldn't affect any regression test output I think? The test generation should be identical, except only the one test I updated to add a `..` test case which has expanded `.cpp` output but should not have any output changes because it doesn't print anything.

Re line# output: Usually the only kind of change that affects regression test line# output is when `cpp2util.h` changes cause the assertion line numbers to change for the handful of tests that a given older compiler doesn't fully support (typically GCC 10, sometimes Clang 12). So I don't think there should be line# regression test impact for this PR... I hope it's a pure extension!

source/sema.h
hsutter: Thanks! Yes I just forgot to remove that part. It shouldn't affect regression tests though. 👍

source/to_cpp1.h
jarzec: Ah, true, you are right.

jarzec: It seems some of the issues solved in #1103 are pre-existent (except the update of `regression-tests/test-results/pure2-ufcs-member-access-and-chaining.cpp`).

jarzec: Done. The tests on #1103 will be failing until you merge it into this branch. You can also merge this one with failures on `pure2-ufcs-member-access-and-chaining.cpp` and then #1103 to fix those.

jarzec (approved):

jcanizales:
> but `x..f()` will now find only a member function.

Hi, what is the use case this is addressing, if you don' t mind me asking?

gregmarr:
> > but `x..f()` will now find only a member function.
> 
> Hi, what is the use case this is addressing, if you don' t mind me asking?

Quite a bit of discussion at https://github.com/hsutter/cppfront/issues/1004

jarzec:
Is the PR ready to be merged?

hsutter:
> Is the PR ready to be merged?

Pretty much, I just wanted to test build+regression with gcc and Clang before merging, as I usually do.

Longer story: My GCC and Clang environments have been disrupted and it's been taking me over a week to get them back. I had been using those compilers via Ubuntu and Fedora running locally on my Windows laptop via WSL2, which was really convenient. However, those guest OSes were recently disabled by a company group policy, and googling around I discovered there were security issues which presumably was the reason for that. So I've been trying to keep using WSL2 Linux distros by recreating those test environments in a cloud-based VM, to keep it separate from my own machine... but it's been quite an experience to figure out the incantations to create that VM correctly... I've kept encountering roadblocks of the form "you're almost there but actually you needed to set option G back when you created the VM and you can't change it now so go start over" after several hours' investment in setting up the VM, and I just started my fourth do-over attempt a few minutes ago (it's provisioning right now). Hopefully I'll get past this hurdle soon and be back to normal productivity. Sigh! I really want to test routinely with various GCC and Clang builds.

And so that's why I was doing this update as a PR... so that I could pull the PR in the VM to do my GCC and Clang testing there.

That said, I should probably just merge this so you don't have to keep waiting, and besides because it is a PR I can see the multi-platform builds are working which is half of what I wanted to test. 👍

`</ramble>`

JarekGlobus:
> > Is the PR ready to be merged?
> 
> Pretty much, I just wanted to test build+regression with gcc and Clang before merging, as I usually do.
> 
> Longer story: My GCC and Clang environments have been disrupted and it's been taking me over a week to get them back. I had been using those compilers via Ubuntu and Fedora running locally on my Windows laptop via WSL2, which was really convenient. However, those guest OSes were recently disabled by a company group policy, and googling around I discovered there were security issues which presumably was the reason for that. So I've been trying to keep using WSL2 Linux distros by recreating those test environments in a cloud-based VM, to keep it separate from my own machine... but it's been quite an experience to figure out the incantations to create that VM correctly... I've kept encountering roadblocks of the form "you're almost there but actually you needed to set option G back when you created the VM and you can't change it now so go start over" after several hours' investment in setting up the VM, and I just started my fourth do-over attempt a few minutes ago (it's provisioning right now). Hopefully I'll get past this hurdle soon and be back to normal productivity. Sigh! I really want to test routinely with various GCC and Clang builds.
> 
> And so that's why I was doing this update as a PR... so that I could pull the PR in the VM to do my GCC and Clang testing there.
> 
> That said, I should probably just merge this so @MaxSagebaum doesn't have to keep waiting, and besides because it is a PR I can see the multi-platform builds are working which is half of what I wanted to test. 👍
> 
> `</ramble>`

A relatively simple solution to running Linux on Windows without security concerns is using Docker containers. It is not difficult to set up such a solution. Docket is very popular so there is a lot of online material on using it.

hsutter:
(deleted some comments with extraneous info)

@JarekGlobus wrote:
> A relatively simple solution to running Linux on Windows without security concerns is using Docker containers. It is not difficult to set up such a solution. Docket is very popular so there is a lot of online material on using it.

I'll take a look. Thanks!

PR #1103, opened 2024-06-10
Fix tests after addition of .. support

jarzec:
Updates test after the support for `..` for non-UFCS was added: the generated `pure2-ufcs-member-access-and-chaining.cpp` updated after the changes in the .cpp2.


hsutter:
Thanks!

hsutter:
Oops, I didn't notice that this PR was against the #1101 branch, and merged them in the wrong order. Sorry, will fix with separate commit...

PR #1116, opened 2024-06-16
[CI] Fix regression tests (add expected results for `mixed-default-arguments` test)

bluetarpmedia:
Add Ubuntu & Mac gcc & clang execution test results for `mixed-default-arguments` test


hsutter:
Thanks!

PR #1117, opened 2024-06-17
Reactivate CI regression tests for Windows + improve CI

jarzec:
- Reactivate CI regression tests on Windows. The test were accidentally disabled (always succeeding without an tests run) though an update of the CI workflow. Update test files:
  - `regression-tests/test-results/msvc-2022-c++20/pure2-assert-expected-not-null.cpp.output` after the inclusion of `..` for non-UFCS,
  - for `mixed-default-arguments.cpp2` after the default initialization of function arguments.
- Add a CI job at the end of the regression tests workflow to aggregate patches generated for each failing run in a single zip fie.
- Temporarily exclude lines with `C:\` from git diff. The reason is that on the GitHub runners the MSVC paths include strange characters that cause `git diff` to spuriously flag differences. I will raise an issue with with the GitHub team responsible for the Windows runners.


regression-tests/run-tests.sh
```diff
@@ -55,17 +57,21 @@ check_file () {
...
+        # Lines includng Windows paths are excluded from diff
+        # This is necessary due to characters in those paths on GitHub runners
+        # that cause git diff to spuriously flag them
         report_diff "$file" \
-            "--ignore-cr-at-eol" \
             "Non-matching $description" \
-            "$patch_file"
+            "$patch_file" \
+            --ignore-cr-at-eol \
+            --ignore-matching-lines="C:\\\\"
```
jarzec, line 74: [Explanation] This option is excluding lines with Windows paths. On the GitHub runners those paths may include characters that seem to cause trouble with `git diff`.
This is hopefully a temporary solution.

regression-tests/test-results/msvc-2022-c++20/pure2-assert-expected-not-null.cpp.output
```diff
@@ -1,20 +1,22 @@
...
+predefined C++ types (compiler internal)(347): note: see declaration of 'std'
 pure2-assert-expected-not-null.cpp2(14): error C2062: type 'int' unexpected
 pure2-assert-expected-not-null.cpp2(14): error C2143: syntax error: missing ';' before '{'
 pure2-assert-expected-not-null.cpp2(14): error C2039: 'unexpected': is not a member of 'std'
-predefined C++ types (compiler internal)(348): note: see declaration of 'std'
-pure2-assert-expected-not-null.cpp2(14): error C3861: 'unexpected': identifier not found
+predefined C++ types (compiler internal)(347): note: see declaration of 'std'
+pure2-assert-expected-not-null.cpp2(14): error C2660: 'unexpected': function does not take 1 arguments
+C:\Program Files\Microsoft Visual Studio�2nterprise\VC\Tools\MSVC
.40.33807\includeh.h(33): note: see declaration of 'unexpected'
```
jarzec, line 17: [Explanation] This path was always indicated as different despite the test file being updated. My conjecture is that this is due to the strange characters in the path.

.github/workflows/regression-tests.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         bash run-tests.sh -c ${{ matrix.compiler }} -s ${{ matrix.cxx_std }} -d ${{ matrix.stdlib }} -l ${{ matrix.os }}
 
     - name: Run regression tests - Windows version
-      if: matrix.os == 'windows-latest'
+      if: startsWith(matrix.os, 'windows')
```
jarzec, line 95: [Explanation] This line was not updated in a change of this workflow. I've noticed that somewhat accidentally looking at a change that should have caused the test run to fail (e.g. in [here](https://github.com/hsutter/cppfront/actions/runs/9504287442/job/26196600829?pr=1103)) while in fact it showed as successful.

hsutter:
Thanks!

Issue #1123, opened 2024-06-18
[BUG] Assert failed, file to_cpp1.h, line 6722.


MarekKnapek:
**Describe the bug**
File `to_cpp1.h`.
Line `6722`.
Expression `assert(n.initializer);`.

**To Reproduce**
1. Download source code of cppfront.
2. Compile cppfront.
2. Run cppfront on test.cpp2 file.
3. Assert triggers inside cppfront.

**Additional context**
Test.cpp2 file's content is only 3 bytes long and it is: `r:;`.

gregmarr:
Confirmed on cpp2.godbolt.org

MaxSagebaum:
> Confirmed on cpp2.godbolt.org

Why not post the short link?

sookach:
> > Confirmed on cpp2.godbolt.org
> 
> Why not post the short link?

https://cpp2.godbolt.org/z/r9KfYb5jE

gregmarr:
> Why not post the short link?

It's literally three characters in the source, seemed like a waste to have it create and store the short link.

PR #1124, opened 2024-06-18
to_cpp1: Remove missing initializer assert

sookach:
Patches #1123

The program is ill formed, and semantic analysis detects the issue, but because errors are checked after lowering, it hits an assert when spitting out the cpp code, causing the crash. We could, of course, check for errors generated by semantic analysis before lowering, but this pr simply swaps out the assert for a return, so as to maintain the current structure.

### DIFF
--- a/source/to_cpp1.h
+++ b/source/to_cpp1.h
@@ -6719,7 +6719,9 @@ public:
 
             //  Emit "auto" for deduced types (of course)
             if (type->is_wildcard()) {
-                assert(n.initializer);
+                if(!n.initializer) {
+                    return;
+                }
                 emit( *type, n.position() );
             }
             //  Otherwise, emit the type


source/to_cpp1.h
```diff
@@ -6719,7 +6719,9 @@ class cppfront
 
             //  Emit "auto" for deduced types (of course)
             if (type->is_wildcard()) {
-                assert(n.initializer);
+                if(!n.initializer) {
+                    return;
```
gregmarr, line 6723: Should there be an error recorded here, like at line 6735?

sookach: there would be a double reporting of the error since sema already records it.

sookach: good eye though

hsutter:
Thanks! I tried to push an alternate solution to this branch but I don't have permission, so I'll apply the solution on `main` separately instead of taking this PR, and reference this PR. Again, thanks for this.

PR #1127, opened 2024-06-20
Fix shadowing issue in sema.h

jarzec:
Fixes an issue causing the multiplatform build test to fail on Windows runners on GitHub.
The MSVC compiler generates a warning about a local variable shadowing a member. The warning is treated as an error  in accordance with the test settings.

The Windows regression test is fixed by #1127.

### DIFF
--- a/source/sema.h
+++ b/source/sema.h
@@ -1511,11 +1511,15 @@ public:
     )
         -> bool
     {
-        std::vector<error_entry> devnull;
-        auto& errors = emit_errors ? this->errors : devnull;
+        const std::function emit_error = [this](source_position pos, std::string_view msg) {
+            errors.emplace_back(pos, msg);
+        };
+        const std::function skip_error = [](source_position, std::string_view){};
+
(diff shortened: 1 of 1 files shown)


source/sema.h
```diff
@@ -1511,11 +1511,15 @@ class sema
...
         -> bool
     {
-        std::vector<error_entry> devnull;
-        auto& errors = emit_errors ? this->errors : devnull;
+        const std::function emit_error = [this](source_position pos, std::string_view msg) {
+            errors.emplace_back(pos, msg);
+        };
+        const std::function skip_error = [](source_position, std::string_view){};
+
+        const auto handle_error = emit_errors ? emit_error : skip_error;
```
jarzec, line 1519: I chose to use two lambdas to make the `emit_errors == false` more explicitly no-op.

hsutter:
Thanks! Now I see it in my local build, I must have forgotten to locally test self-build before checking that in. Thanks for the catch.

Issue #594, opened 2023-08-16
Question: how to generate C++ headers from cppfront?


vladimir-kraus:
Please correct me if I am mistaken or let me know whether I am lagging behind the status-quo of cppfront design.

I believe that the success of any new "C++ successor" language will be determined not only how easy and safe it is to write **new code** but also by how easy it will be gradually translate **existing code** from C++ to cppfront. The simplest approach of transforming existing code would be to rewrite classes (typically with declaration in *.h and implementation in *.cpp) one by one from C++ to cppfront. This would exactly represent the way of language adoption as expressed by smooth ramp-up line as presented by Herb at the latest CPP conference. 

I have read that cppfront is probably aiming to ditch the idea of C++ headers altogether and rely only on modules. I am no expert i modules (so maybe there is some magical solution...) but I think that ditching C++ headers completely may harm the process of cppfront gradual adoption. Simply because existing C++ code expects to include class headers. Majority of existing code is written without modules in mind.

So in my opinion, cppfront should have a way to generate headers alongside with *.cpp files. I do not know how this can be done now. I learned that there i something like *.h2 files. So I experimented with this a bit. I took a very naive (and wrong) approach...

```
// file widget.h2

Widget: type = {
    x : int = 0;
    y : int = 0;
    sum : (this) -> int;  // I naively attempted something like a method declaration here. It does not compile. I know it is wrong.
}
```

```
// file widget.cpp2

// I am naively trying something like a method definition. It does not compile. I know it is wrong.
sum : (this : Widget) -> int = {
    return x + y;
}
```

```
// file main.cpp2

#include "widget.h2"

main: (args) = {
    w := Widget();
    std::cout << w.sum();
}
```
I know this is wrong approach but given the documentation does not exist, I did not discover any other solution which would do the same and would work. I would expect that cppfront would generate `widget.h`, `widget.cpp` and `main.cpp` which would be exactly what a programmer would write in hand with C++, but with cppfront it will be much safer and concise. The benefit would be that other existing C++ code could also include the generated `widget.h` if it needs it.

Another possible approach would be writing just `widget.cpp2` but also containing method implemantations but some magical @... directive added to the class  would cause that a *.h file would be generated alongside with *.cpp. The header would contain class definition and method declarations, and cpp file would contain method implementations.

However as I wrote above, maybe some other and better solution already exists in cppfront and I am not aware of it. In that case I would love to learn about how to solve the issue above.

Addendum:

[...]

PR #1141, opened 2024-06-24
Feature to allow multiple translation unit compilations of generated pure headers

DrizztDoUrden:
The purpose of this PR is to allow including pure headers into multiple translation units without ODR violations.

First and foremost it splits `phase2_func_defs` into `phase2_inline_defs` and `phase3_regular_defs`. Produced .h file gets all declarations (as before), alias definitions and anything template-related, while the rest go to .hpp.

A flag `-tu-compatible-h2` is added that stops generated hpp files from including other hpps as it is not necessary to produce a buildable source, but the default behaviour is unchanged. When using the flag it is important to include the resulting .hpp at least somewhere as it is never included automatically.

With a simple-ish cmake functions it makes it possible to produce automatically both header with declarations and inline definitions and a cpp with non-inline definitions. As an example, in my hobby project cmake parses h2s, finds #cppinclude lines and adds them to a new .cpp as regular #include, than includes the generated .hpp.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -3131,6 +3131,43 @@ public:
     auto parent_is_polymorphic() const -> bool
         { return  parent_declaration && parent_declaration->is_polymorphic(); }
 
+    auto is_inline() const -> bool
+    {
+        return is_alias()
+            || is_constexpr
+            || template_parameters != nullptr;
+    }
+
+    auto parent_is_inline() const -> bool
+    {
+        for (auto p = parent_declaration; p; p = p->parent_declaration)
+        {
+            if (p->is_inline())
+                return true;
+        }
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


hsutter:
Thanks for the suggestion and investing the time to provide code! I may have cycles to take a look at this but for now I'm putting it in bucket 4 in the current priorities (#1287) as I'm taking a pass to triage the outstanding PRs.

vincent-hui:
@DrizztDoUrden  I would like to try this feature, would you mind sharing an example project with me? Thanks.

MatthieuHernandez:
Will this feature avoid having to do this?
https://github.com/hsutter/cppfront/issues/594#issuecomment-2766300571

PR #1150, opened 2024-07-06
CI update interpolation test results

jarzec:
Update the execution outputs of pure2-interpolation.cpp2 regression test.


hsutter:
Thanks!

Issue #1152, opened 2024-07-08
[BUG]  Cppfront can not parse "\x{62}"


MaxSagebaum:
Cppfront produces the error:
```
main.cpp2(3,22): error: string literal "\" is missing its closing "
main.cpp2(5,32): error: local variable c cannot be used in its own initializer
```
while parsing `""\x{62}blub""`

A reproducer on compiler explorer:
https://cpp2.godbolt.org/z/h6e8vc1rv

One possible patch is:
```
diff --git a/source/lex.h b/source/lex.h
index bab314a..ba3be0e 100644
--- a/source/lex.h
+++ b/source/lex.h
@@ -939,10 +939,16 @@ auto lex_line(
         if (
             peek(  offset) == '\\'
             && peek(1+offset) == 'x'
-            && is_hexadecimal_digit(peek(2+offset))
+            && (is_hexadecimal_digit(peek(2+offset))
+              || (peek(2+offset) == '{' && is_hexadecimal_digit(peek(3+offset)))
             )
+        )
         {
+            bool has_bracket = peek(2+offset) == '{';
             auto j = 3;
+
+            if (has_bracket) { ++j; }
+
             while (
                 peek(j+offset)
                 && is_hexadecimal_digit(peek(j+offset))
@@ -950,6 +956,14 @@ auto lex_line(
             {
                 ++j;
             }
+
+            if (has_bracket) {
+                if (peek(j+offset) == '}') {
+                    ++j;
+                } else {
+                    return 0;
+                }
+            }
             return j;
         }
         return 0;
```

It seems that there is currently no regression test for the escapes.

MaxSagebaum:
Created a pull request https://github.com/hsutter/cppfront/pull/1153.

hsutter:
Thanks! TIL:

> can not parse `"\x{62}"`

I wasn't familiar with that extension, and after some spelunking discovered it was added for C++23 by paper [P2290R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2290r3.pdf).

It seems to be accepted by GCC 13.1 or higher and Clang 14 or higher (and they backported it all the way to `-std=c++98` mode). It does not seem to work at all on MSVC though.

I think there is an even better alternative fix though: C++23 `"\x{62}blub"` has the identical meaning as C++98 `"\x62" "blub"`, so if I (finally) supported string literal concatenation, that would give you a way to write this. Would that solve your problem?

MaxSagebaum:
[...]

PR #1153, opened 2024-07-08
Fix for hexadecimal escapes of the from `\x{62}`.

MaxSagebaum:
Closes https://github.com/hsutter/cppfront/issues/1152.

Missing things:
 - [x] Review
 - [x] Regression tests if required.

### DIFF
--- a/source/lex.h
+++ b/source/lex.h
@@ -930,21 +930,27 @@ auto lex_line(
         return 0;
     };
 
+    //G simple-hexadecimal-digit-sequence:
+    //G     hexadecimal-digit
+    //G     simple-hexadecimal-digit-sequence hexadecimal-digit
+    //G
     //G hexadecimal-escape-sequence:
     //G     '\x' hexadecimal-digit
     //G     hexadecimal-escape-sequence hexadecimal-digit
+    //G     '\x{' simple-hexadecimal-digit-sequence '}'
     //G
     auto peek_is_hexadecimal_escape_sequence = [&](int offset)
     {
         if (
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks, looks good! I'm going to push another commit to refine this:

- Also support `\u{...}` for universal character names (I've avoided doing octal so I'll ignore that `\o{...}` part of P2290)
- Add `//G` grammar comments
- Run regression tests

source/lex.h
```diff
@@ -930,26 +930,51 @@ auto lex_line(
...
     };
 
+    //G simple-hexadecimal-digit-sequence:
+    //G     hexadecimal-digit
+    //G     simple-hexadecimal-digit-sequence hexadecimal-digit
+    //G
     //G hexadecimal-escape-sequence:
     //G     '\x' hexadecimal-digit
     //G     hexadecimal-escape-sequence hexadecimal-digit
+    //G     '\x{' simple-hexadecimal-digit-sequence '}'
```
gregmarr, line 940: Would it make sense to reuse `simple-hexadecimal-digit-sequence` for both cases?

    //G hexadecimal-escape-sequence:
    //G     '\x' simple-hexadecimal-digit-sequence
    //G     '\x{' simple-hexadecimal-digit-sequence '}'

hsutter: I wondered that too. I followed the adopted paper's wording, which didn't change the original productions.

However, now I see the that since then the standard did exactly what you suggest, so I'll do that too. Thanks!

PR #1154, opened 2024-07-08
CI Add Clang-18 C++23 regression tests

jarzec:
As stated in the tile. After #1088 cppfront compiles using Clang-18 with C++23 selected. For this to work an up-to-date libc++ has to be used.
- The `build-cppfornt.yml` workflow file is modified as Clang-18 accepts `c++23` as language standard option.
- The `regression-tests.yml` is modified to include the new compiler options.
- Reference files are added.
- `run-tests.sh` is modified to make the label option (`-l`) mandatory. If it is not provided error reporting fails.

The added compiler option is the only one that allows for all tests to compile on Ubuntu. Out of all the other options only `mac-14 | clang++ | c++2b | default` and `win-22 | cl.exe | c++latest | default` correctly lower and compile all the regression test.


hsutter:
Thanks!

PR #1164, opened 2024-07-15
CI Remove empty test reference files

jarzec:
- Update the `run-tests.sh`.
  - Handle empty generated files as correct without the need for committing them.
  - Remove such files generated by the test.
  - Report empty reference files as errors to maintain the status quo.
- Add a `rm-empty-files.sh` utility script for removing all empty files in current directory.
- Remove all empty reference test files (>1500 empty files less in the repository).

Note: This PR is one step towards improvements of the kind proposed in #1137. It should make further improvement simpler.
An iterative approach should be easier to handle.

### DIFF
--- a/regression-tests/rm-empty-files.bat
+++ b/regression-tests/rm-empty-files.bat
@@ -1,0 +1,13 @@
+@echo off
+setlocal enabledelayedexpansion
+
+:: Loop through all files in the current directory and subdirectories
+for /r %%f in (*) do (
+    :: Check if the file size is 0 bytes
+    if %%~zf==0 (
+        :: Remove the empty file
+        del "%%f"
+    )
+)
+
+endlocal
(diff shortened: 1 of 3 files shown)


jarzec:
The test failures will be fixed by #1165.

hsutter:
Thanks!

jarzec:
@hsutter  My last update of this PR was missing some changes from main and is now causing test failures.
I am working on a fix in #1176 - it should become all green soon.

hsutter:
Hmm... one thing I meant to ask before merging this (but forgot, sorry) was:

Early on, the reason I was storing files even if empty was because otherwise I couldn't distinguish between "test ran successfully and there was no output" and "test crashed and failed to generate any output."

Are we still going to catch the "empty" vs "crashed" case somehow so that we don't start getting silent regressions due to crashes for test cases that would have empty output anyway?

jarzec:
That is a good point. The test script currently silently removes the generated empty file. I will add a check that the file actually got created.

PR #1172, opened 2024-07-21
Initial range operators implementation

hsutter:
Examples: `v.begin()...end()` and `1..=10`

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -2107,7 +2107,7 @@ inline auto make_args(int argc, char** argv) -> args
 
 //-----------------------------------------------------------------------
 //
-//  range: a half-open range of [first, last)
+//  range: a range of [begin, end) or [first, last]
 //
 //-----------------------------------------------------------------------
 //
@@ -2117,7 +2117,7 @@ struct range
     range(
         T const& f, 
         T const& l, 
-        bool include_last = false
+        bool     include_last = false
     ) 
         : first{ f }
         , last{ l } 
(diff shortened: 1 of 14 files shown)


include/cpp2util.h
```diff
@@ -2084,24 +2087,90 @@ struct args_t
...
+        T first;
+        T last;
+        T curr;
+    };
+
+    auto begin()  const -> iterator { return iterator{ first, last, first }; }
+    auto end()    const -> iterator { return iterator{ first, last, last }; }
+    auto cbegin() const -> iterator { return begin(); }
+    auto cend()   const -> iterator { return end(); }
+    auto size()   const -> std::size_t { return /*cpp2::unsafe_narrow<std::size_t>*/(ssize()); }
```
jarzec, line 2161: [Question] Do you want to eventually remove the commented part or keep it aligned with the `args::iterator` code in line 2090?

include/cpp2util.h
```diff
@@ -2084,24 +2087,90 @@ struct args_t
...
+        T const& l, 
+        bool include_last = false
+    ) 
+        : first{ f }
+        , last{ l } 
+    { 
+        if (include_last) { 
+            ++last; 
+        }
+    }
```
jarzec, line 2128: [Suggestion] How about something like (I am not great at choosing names):
```suggestion
    enum class type {
        ExcludeLast,
        IncludeLast
    };

    range(
        T const& f, 
        T const& l, 
        type t = type::ExcludeLast
    )
        : first{ f }
        , last{ l } 
    { 
        if (t = type::IncludeLast) { 
            ++last; 
        }
    }
```
The point here is that calls for closed intervals would change from `range(f, l, true)` to `range(f, l, type::IncludeLast)`, which expresses intent better.

This could even go even further to something of the form:
```
    template<range_type Type>
    range { /*...*/ };
```
including `if constexpr`.

include/cpp2util.h
hsutter: Thanks, good catch -- I had commented that out while debugging something, which I guess I've now fixed so I can restore that code. 👍

include/cpp2util.h
hsutter: Thanks! This is in the generated code, but still I'll take a look at making it more readable. (I do use a fair number of `bool` flags in cppfront...)

Re `if constexpr`: Right now the only place that flag is used is in the constructor, it's never needed later. So I think there's no opportunity for `if constexpr` to make simpler code paths, there's already a single logic for all ranges?

source/parse.h
```diff
@@ -1802,7 +1806,10 @@ auto postfix_expression_node::visit(auto& v, int depth)
             x.id_expr->visit(v, depth+1);
         }
         if (x.expr_list) {
-            x.expr_list->visit(v, depth+1);
+            x.expr_list->visit(v, depth + 1);
+        }
+        if (x.last_expr) {
+            x.last_expr->visit(v, depth+1);
```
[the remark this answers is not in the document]
hsutter: Thanks! An autoformat I didn't undo... :)

include/cpp2util.h
```diff
@@ -2084,24 +2087,90 @@ struct args_t
...
+inline auto make_args(int argc, char** argv) -> args
 {
-    return args_t{argc, argv};
+    return args{argc, argv};
 }
 
 
+//-----------------------------------------------------------------------
+//
+//  range: a half-open range of [first, last)
```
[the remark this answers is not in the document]
hsutter: Thanks! Will fix comment.

I'm reluctant to make this more complicated (e.g., bifurcate into two types, use a policy parameter), because the current design seems to work well with all the code identical except only for one additional test in the constructor only?

source/parse.h
```diff
@@ -5928,6 +5938,7 @@ class parser
     //G     postfix-expression '(' expression-list? ','? ')'
     //G     postfix-expression '.' id-expression
     //G     postfix-expression '..' id-expression
+    //G     postfix-expression '...' primary-expression
```
gregmarr, line 5941: Do you need `..=` here too?

hsutter: Yes I do! Thanks, will fix.

PR #1173, opened 2024-07-21
CI Update tests after recent changes

jarzec:
As stated in the title:
- Line number updates in error messages.
- Files missing likely after #904.


hsutter:
Thanks!

PR #1175, opened 2024-07-21
Remove forward slash as command line switch

### LINKED ISSUE #1174: [BUG] Absolute paths no longer work as input
**Describe the bug**
After the following commit, absolute paths are misinterpreted as unrecognized flags.

https://github.com/hsutter/cppfront/commit/16686c1ce15df4efb0802e96865e39c86be877cb

**To Reproduce**

```
~$ git clone https://github.com/hsutter/cppfront
~$ cd cppfront
~/cppfront$ clang++-19 -o cppfront -O2 -std=c++20 source/cppfront.cpp
~/cppfront$ ./cppfront "$PWD/regression-tests/pure2-enum.cpp2"
/home/alex/cppfront/regression-tests/pure2-enum.cpp2 - unknown compiler flag name (try /help)
```

**Additional context**
N/A

sookach:
Fixes #1174. According to the docs, forward slashes are not used for any flags (and I can't think of a good reason for them to be, given the problems it would cause, like the mentioned issue), so this pr just removes forward slashes for command line options.

### DIFF
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -63,7 +63,7 @@ auto main(
     {
         if (
             arg.text.starts_with("-")
-            || arg.text.starts_with("/")
+            || (arg.text.starts_with("/") && arg.text.find('.') != std::string::npos)
             )
         {
             auto ambiguous = cmdline.flags_starting_with(arg.text.substr(1));
--- a/source/cppfront.cpp
+++ b/source/cppfront.cpp
@@ -63,7 +63,7 @@ auto main(
     {
         if (
             arg.text.starts_with("-")
-            || (arg.text.starts_with("/") && arg.text.find('.') != std::string::npos)
+            || (arg.text.starts_with("/") && !contains(arg.text, '.'))
             )
         {
             auto ambiguous = cmdline.flags_starting_with(arg.text.substr(1));


jarzec:
I think this was added for compatibility with Windows where `/` is pretty much the default way to provide command line arguments.
I agree that the everlasting choice to use of `\` in paths and `/` for command line arguments on Windows is... breakneck (i.e. is a pain in the... neck 😉)

hsutter:
Thanks! Yes, the issue is Windows familiarity.

Maybe I'll just emit that warning if there's no `.` character in the name!

sookach:
> Thanks! Yes, the issue is Windows familiarity.
> 
> 
> 
> Maybe I'll just emit that warning if there's no `.` character in the command-line argument!

That's a good thought but I'm not too sure if that will work. The linked issue is from a user trying to use an absolute file path, so it shouldn't have a '.'

bluetarpmedia:
> That's a good thought but I'm not too sure if that will work. The linked issue is from a user trying to use an absolute file path, so it shouldn't have a '.'

If the input file is `filename.cpp2` or `filename.h2` then it will, though I'm not sure if that's a requirement for the source file to have an extension.

sookach:
> > That's a good thought but I'm not too sure if that will work. The linked issue is from a user trying to use an absolute file path, so it shouldn't have a '.'
> 
> If the input file is `filename.cpp2` or `filename.h2` then it will, though I'm not sure if that's a requirement for the source file to have an extension.

Ahh, my bad, I misunderstood Herb's original comment. I thought he meant a leading dot. All good then.

sookach:
Cool, I updated the pr to skip over arguments that begin with a forward slash and contain a dot.

hsutter:
Thanks!

PR #1177, opened 2024-07-23
feat: Support function types - an attempt to rebase #526 with current main

hsutter:
Hand-merged #526 into current main code

Some basic examples compile, but `reflect.h2` doesn't yet, needs investigation

My main additional change so far to the #526 diffs is that I've commented out some of the new error message emissions on what appeared to be trial parses(?), because the presence of the error message stopped further processing and caused programs like `f: (x: int) { }` to fail

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -370,6 +370,17 @@ using _uchar     = unsigned char;    // normally use u8 instead
 
 //-----------------------------------------------------------------------
 //
+//  fn_t<R(ArgTypes...)>       For emitted Cpp2 function types
+//
+//-----------------------------------------------------------------------
+//
+template<typename T>
+    requires std::is_function_v<T>
+using fn_t = T;
+
+
+//-----------------------------------------------------------------------
+//
 //  String utilities
 //
 
--- a/source/parse.h
+++ b/source/parse.h
@@ -1325,6 +1325,8 @@ struct qualified_id_node
 };
 
 
+struct function_type_id_node;
+
 struct type_id_node
 {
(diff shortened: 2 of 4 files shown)


hsutter:
Thanks again #526, and deep apologies for the delay as I haven't kept up with PRs.

This now builds, but isn't quite working yet because I haven't yet fully grokked how this affects parsing normal function types, which are failing in `reflect.h2` for example which is my current test case. I suspect it's something fairly simple... @JohelEGP I know you're busy with other things now, but if you do have a chance to take a look I'd appreciate it and can prioritize my time for this now. Thanks again for this contribution!

JohelEGP:
I'm happy to see this take off.

> caused programs like `f: (x: int) { }`

That seems wrong, because `f` is a function, and not a function type.
Some condition that guards `undo_function_parse` might be wrong.

hsutter:
Interim update: I've spent the last couple of days on this but haven't been able to make more progress in getting it working. Today I tried doing a fresh from-scratch parallel implementation with a different implementation approach, but only got basic `std::function` cases working and have hit some barriers on that path.

I'll need to set this aside for a few days and then maybe try to take a run at it again...

hsutter:
OK, I haven't been able to get this branch working, but I'm making progress on an independent implementation, will create a new PR for that...

hsutter:
OK, I got my second try in #1183 working, so I'll merge that in favor of this attempt to rebase the original PR. Thanks again though for #526 and your many other contributions already merged, @JohelEGP !

PR #1187, opened 2024-07-30
[CI] Fix clang-18 C++23 build of cppfront

bluetarpmedia:
Some constructors and destructors have to be defaulted/written out-of-line in order to compile successfully with clang 18 in C++23 mode, due to forward declaring types and using them with `unique_ptr`.

Fixes [this build error](https://github.com/hsutter/cppfront/actions/runs/10153992510/job/28078384021).

Similar issue as solved by this previous PR: https://github.com/hsutter/cppfront/pull/1088

EDIT: This only fixes the CI build of cppfront; I'll create a separate PR to update the regression tests.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -2273,7 +2273,9 @@ struct parameter_declaration_node
 
     std::unique_ptr<declaration_node> declaration;
 
-    parameter_declaration_node(parameter_declaration_list_node const* my) : my_list{my} { }
+    // Out-of-line definition of the ctor is necessary due to the forward-declared
+    // type(s) used in a std::unique_ptr as a member
+    parameter_declaration_node(parameter_declaration_list_node const* my);
 
     //  API
     //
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks! But this is only needed for the destructor, right? Not the constructor?

bluetarpmedia:
It's needed for both in this case.

Here's a [repro](https://godbolt.org/z/17K1jK785) showing the ctor issue:

```cpp
#include <memory>

struct MyType;

struct Owner
{
    std::unique_ptr<MyType> mine_;

    Owner() {}
};

struct MyType
{
    int value = 0;
};

int main()
{
    Owner o;
}
```

Relevant hint from the error:
```
<source>:9:5: note: in instantiation of member function 'std::unique_ptr<MyType>::~unique_ptr' requested here
    9 |     Owner() {}
```
I _think_ it needs to be able to destruct the `unique_ptr` in case the ctor throws?

Fixed [version](https://godbolt.org/z/hGMKxd8q3):
```cpp
#include <memory>

struct MyType;

struct Owner
{
    std::unique_ptr<MyType> mine_;

    Owner();
};

struct MyType
{
    int value = 0;
};

Owner::Owner() {}

int main()
{
    Owner o;
}
```

hsutter:
> I _think_ it needs to be able to destruct the `unique_ptr` in case the ctor throws?

That reasoning makes sense to me. Then I wondered, 'why do other types in parse.h only have to put the dtor out of line for this?' I looked, and I think the answer is because those types don't have a user-declared constructor... bingo.

So I think this makes sense. Thanks!

PR #1192, opened 2024-07-30
[CI] Update regression tests for clang-15 on macOS and clang-18 on ubuntu

bluetarpmedia:
I must've somehow missed these changes in my previous PR, sorry!


hsutter:
Thanks!

PR #1195, opened 2024-07-31
Documentation for regular expressions.

MaxSagebaum:
Basic documentation for regular expressions.
 - Added section for regular expressions in the metafunction documentation. This section explains how to use them and the public interface.
 - Added a list of all supported features. I could not add this as a subpage for the metafunction documentation. Therefore I created a category `other`.


MaxSagebaum:
@hsutter Here is the promised documentation for the regular expressions.

docs/cpp2/metafunctions.md
```diff
@@ -372,6 +372,67 @@ A `cpp1_rule_of_zero` type is one that has no user-written copy/move/destructor
 > This is known as "the rule of zero".
 > — Stroustrup, Sutter, et al. (C++ Core Guidelines)
 
+#### `regex`
+
+Replaces fields in the class with regular expression objects. Each field starting with `regex` is replaced with a regular expression of the same type.
```
gregmarr, line 377: Is it "starting with `regex`" or is it "named `regex` or starting with `regex_`"?

docs/other/regex_status.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Current status and planned on doing
+
+### Modifiers
+```
+ - [x] i                Do case-insensitive pattern matching. For example, "A" will match "a" under /i.
+ - [x] m                Treat the string being matched against as multiple lines. That is, change "^" and "$" from matching the start of the string's first line and the end of its last line to matching the start and end of each line within the string.
+ - [x] s                Treat the string as single line. That is, change "." to match any character whatsoever, even a newline, which normally it would not match.
+ - [x] x and xx         Extend your pattern's legibility by permitting whitespace and comments. Details in "/x and /xx"
+ - [x] n                Prevent the grouping metacharacters () from capturing. This modifier, new in 5.22, will stop $1, $2, etc... from being filled in.
```
gregmarr, line 13: Probably don't need "new in 5.22," here.

docs/cpp2/metafunctions.md
gregmarr: Currently it's "starting with `regex`" and it doesn't care what comes after it, it can be `regexes` if you want.

MaxSagebaum: Yes, until now I did not think about this corner case.  I will do an update that checks for `regex` and `regex_*`.

The long term goal would be to apply the metafunction to the variable definition, e.g.:
```
name_matcher: @regex = "regex here";
```

docs/other/regex_status.md
MaxSagebaum: I will have another run on this and remove the perl version reference. For me this was ok, but I can understand if this throws people off.

MaxSagebaum:
Thanks for the notes, sorry for the mail spam. For some reason github does not allows to start an combined answer.

jcanizales:
> For some reason github does not allows to start an combined answer.

There's a workaround way that I use, which is to start your own review of your code, and send all your answers together as a "comment" review.

hsutter:
Thanks! This is looking good.

I just took a pass over it... the main change was to turn the code lists into formatted tables.

Could you please sanity-check what I did? Then I think this is ready to merge, thanks again for all the great work and detail!

MaxSagebaum:
Thanks for the updates. You could have also told me to convert the regex status file to tables. Would have been no problem. 

I had a look at it and fixed some small things.

hsutter:
> You could have also told me to convert the regex status file to tables. Would have been no problem.

No worries! I didn't know if it would look right until I tried doing it myself, so at first it was an experiment. Then, once I liked how it looked, my fingers had the hang of it so it was easy to do the rest. Also, doing manual formatting helps me read text more carefully.

MaxSagebaum:
I rebased the branch. It should be ready for merging now.

hsutter:
Great, thanks again for this.

Issue #1197, opened 2024-08-01
[SUGGESTION] Add the equivalent of C++23 "deducing this"


bluetarpmedia:
For the same motivations as in [P0847](https://wg21.link/p0847), I'd like to be able to write the equivalent of the "deducing this" feature in Cpp2.

Here's one example where the feature would prevent duplicating code:
```cpp
my_optional: <T> type = {
    has_value: bool = false;
    value:     T;

    get: (in this)    -> forward T = { assert(has_value); return value; }  // return const-ref
    get: (inout this) -> forward T = { assert(has_value); return value; }  // return mutable-ref
    get: (move this)  -> move    T = { assert(has_value); return value; }
}
```

Instead I'd like to write the `get` function once, with some kind of equivalent "deducing this" syntax that takes care of the `this` parameter and also the return type, e.g. along the lines of:

```cpp
get: <Self>(Self _ this) -> _ = { assert(has_value); return value; }
```

The lowered C++ could duplicate the functions if it's targeting C++20, since the C++23 "deducing this" syntax won't be available.

If, in the future, cppfront supports some kind of `cpp_standard` flag (like proposed in #942) then the lowered code for C++23 could use the real "deducing this" feature.

**Will your feature suggestion eliminate X% of security vulnerabilities of a given kind in current C++ code?**
No

**Will your feature suggestion _automate or eliminate_ X% of current C++ guidance literature?**
Yes, in the same way that "deducing this" does for C++.

SebastianTroy:
Instead of making special syntax, would

get: (forward this) -> forward T = {...}

Work instead?


On 1 August 2024 07:43:48 Neil Henderson ***@***.***> wrote:

For the same motivations as in P0847<https://wg21.link/p0847>, I'd like to be able to write the equivalent of the "deducing this" feature in Cpp2.

Here's one example where the feature would prevent duplicating code:

my_optional: <T> type = {
    has_value: bool = false;
    value:     T;

    get: (in this)    -> forward T = { assert(has_value); return value; }  // return const-ref
    get: (inout this) -> forward T = { assert(has_value); return value; }  // return mutable-ref
    get: (move this)  -> move    T = { assert(has_value); return value; }
}

Instead I'd like to write the get function once, with some kind of equivalent "deducing this" syntax that takes care of the this parameter and also the return type, e.g. along the lines of:

get: <Self>(Self _ this) -> _ = { assert(has_value); return value; }

[...]

PR #1198, opened 2024-08-01
Fix some sign conversion warnings emitted by clang in `cpp2regex.h`

bluetarpmedia:
This PR adds some `size_t` casts when accessing the `match_context` groups `std::array` to prevent some clang sign-conversion warnings.

This Cpp2:
```cpp
matcher: @regex type = {
    regex_number:= R"([0-9]*)";
}

main: () -> int = {
    m: matcher = ();
    res:= m.regex_number.match("0123");
    std::print("matched: {} {}", res.matched, res.group(0));
    return 0;
}
```
causes clang to emit some warnings & notes (edited for brevity, the repro has the full details):
```
warning: implicit conversion changes signedness: 'const int' to 'size_type' (aka 'unsigned long') [-Wsign-conversion]
```
[Repro](https://cpp2.godbolt.org/z/qbzdK8G9a)

The diff for the regenerated `cpp2regex.h` is very large due to new lines inserted into the `h2`.

CC: @MaxSagebaum 

I've left this as a draft for now. I noticed that the GCC regression tests had some failures (unrelated to regex) in my repo because GCC has used different `auto` identifiers:

```
-mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:257&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:257 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
+mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:259&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:259 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
```
So I'm waiting to see if that changes again.

### DIFF
--- a/include/cpp2regex.h2
+++ b/include/cpp2regex.h2
@@ -90,38 +90,42 @@ match_context: <CharT, Iter, max_groups: int> type =
 
     //  Getter and setter for groups
     //
-    get_group: (in this, group) groups[group];
+    get_group: (in this, group) groups[get_group_idx(group)];
 
     get_group_end:    (in this, group) -> int = {
-        if group >= max_groups || !groups[group].matched {
+        idx:= get_group_idx(group);
+        if group >= max_groups || !groups[idx].matched {
             return 0;
         }
-        return cpp2::unsafe_narrow<int>( std::distance(begin, groups[group].end) );
+        return cpp2::unsafe_narrow<int>( std::distance(begin, groups[idx].end) );
     }
     get_group_start:  (in this, group) -> int = {
-        if group >= max_groups || !groups[group].matched {
+        idx:= get_group_idx(group);
+        if group >= max_groups || !groups[idx].matched {
             return 0;
         }
-        return cpp2::unsafe_narrow<int>( std::distance(begin, groups[group].start) );
+        return cpp2::unsafe_narrow<int>( std::distance(begin, groups[idx].start) );
     }
     get_group_string: (in this, group) -> std::string = {
-        if group >= max_groups || !groups[group].matched {
+        idx:= get_group_idx(group);
+        if group >= max_groups || !groups[idx].matched {
             return "";
         }
-        return std::string(groups[group].start, groups[group].end);
+        return std::string(groups[idx].start, groups[idx].end);
     }
     
     set_group_end: (inout this, group, pos) = {
-        groups[group].end = pos;
-        groups[group].matched = true;
+        idx:= get_group_idx(group);
+        groups[idx].end = pos;
+        groups[idx].matched = true;
     }
 
     set_group_invalid: (inout this, group) = {
-        groups[group].matched = false;
+        groups[get_group_idx(group)].matched = false;
     }
 
     set_group_start: (inout this, group, pos) = {
-        groups[group].start = pos;
+        groups[get_group_idx(group)].start = pos;
     }
 
     size: (in this) max_groups;
@@ -130,6 +134,13 @@ match_context: <CharT, Iter, max_groups: int> type =
     //
     fail: (in this)            match_return<Iter>(false, end);
     pass: (in this, cur: Iter) match_return<Iter>(true, cur);
+
+    //  Private functions
+    //
+    get_group_idx: (in this, group) -> size_t = {
+        assert(group >= 0);
+        return group as size_t;
+    }
 }
 
 
--- a/include/cpp2regex.h2
+++ b/include/cpp2regex.h2
@@ -137,7 +137,7 @@ match_context: <CharT, Iter, max_groups: int> type =
 
     //  Private functions
     //
-    get_group_idx: (in this, group) -> size_t = {
+    private get_group_idx: (in this, group) -> size_t = {
         assert(group >= 0);
         return group as size_t;
     }
--- a/include/cpp2regex.h
+++ b/include/cpp2regex.h
@@ -27,117 +27,117 @@ template<typename Iter> class match_return;
 #line 77 "cpp2regex.h2"
 template<typename CharT, typename Iter, int max_groups> class match_context;
 
-#line 138 "cpp2regex.h2"
+#line 149 "cpp2regex.h2"
 class true_end_func;
 
-#line 146 "cpp2regex.h2"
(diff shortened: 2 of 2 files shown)


bluetarpmedia:
@jarzec Have you seen this before with GCC? The 3 GCC jobs are showing different numbers for the `auto` parameters.

E.g. here's the diff from the g++-14 job:
```
-mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:257&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:257 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
+mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:259&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:259 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
```

I'm not sure what those numbers are or why GCC puts them in. This is the equivalent line for the clang output:
```
mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto &, cpp2::impl::in<int>, cpp2::impl::in<int>) [rng:auto = std::vector<int>]: Bounds safety violation
```

include/cpp2regex.h2
```diff
@@ -130,6 +134,13 @@ match_context: <CharT, Iter, max_groups: int> type =
     //
     fail: (in this)            match_return<Iter>(false, end);
     pass: (in this, cur: Iter) match_return<Iter>(true, cur);
+
+    //  Private functions
+    //
+    private get_group_idx: (in this, group) -> size_t = {
```
gregmarr, line 140: I thought "since this is always used to access an element of `groups`, can we make a function that handles that".  I ended up with this:
```
    private get_group: (inout this, group) -> forward match_group<Iter> = {
        assert(group >= 0);
        return groups[group as size_t];
    }
```
and then used like this:
```
    get_group(group).start = pos;
```

I discovered that in order to get a modifiable `match_group` back, the function needed `inout this` and a return type of `forward match_group<Iter>`.

@hsutter  That's a little ugly.  I realize that you want to eliminate references as anything but input parameters to functions, but I hadn't really thought about what that would mean for trying to write functions like this.   Is your expectation that we'd do this?

```
    private get_group: (inout this, group) -> *match_group<Iter> = {
        assert(group >= 0);
        return groups[group as size_t]&;
    }
```
```
        get_group(group)*.start = pos;
```

hsutter: > to get a modifiable `match_group` back, the function needed `inout this` and a return type of `forward match_group<Iter>`.

In general, if you're going to return a `forward`ed modifiable parameter, that parameter needs to be modifiable == `inout`:

```cpp
f: (inout x: int) -> forward int = x;

main: () = {
    xx := 42;
    f(xx) = 43;  // ok, changes xx to 43
    std::cout << xx;  // prints 43
}
```

Does this example answer what you were asking?

gregmarr: > Does this example answer what you were asking?

No, I'm talking about writing the Cpp2 equivalent of `T& vector<T>::operator[](size_t index);`.

jarzec:
> @jarzec Have you seen this before with GCC? The 3 GCC jobs are showing different numbers for the `auto` parameters.
> 
> E.g. here's the diff from the g++-14 job:
> 
> ```
> -mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:257&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:257 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
> +mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto:259&, cpp2::impl::in<int>, cpp2::impl::in<int>) [with auto:259 = std::vector<int>; cpp2::impl::in<int> = const int]: Bounds safety violation
> ```
> 
> I'm not sure what those numbers are or why GCC puts them in. This is the equivalent line for the clang output:
> 
> ```
> mixed-bounds-safety-with-assert.cpp2(11) void print_subrange(const auto &, cpp2::impl::in<int>, cpp2::impl::in<int>) [rng:auto = std::vector<int>]: Bounds safety violation
> ```

This is new to me as well. I might have some time soon to have a look at it.

include/cpp2regex.h2
bluetarpmedia: > I thought "since this is always used to access an element of `groups`, can we make a function that handles that".

Yes, that was exactly my thinking at first too! 😃 

Because the `match_context` type has both const and non-const member functions that would need to call this new helper function, it meant I had to write two overloads of the same getter function that returns either a const-ref or a mutable-ref to the element in the groups array, with duplicate implementations.

```cpp
private get_group: (in this, group_idx) -> forward match_group<Iter> = {
    assert(group_idx >= 0);
    return groups[group_idx as size_t];
}

private get_group: (inout this, group_idx) -> forward match_group<Iter> = {
    assert(group_idx >= 0);
    return groups[group_idx as size_t];
}
```
These lower to (slightly edited for clarity):
```cpp
auto get_group(auto const& group_idx) const& -> match_group<Iter> const&;
auto get_group(auto const& group_idx) &      -> match_group<Iter>&;
```

I'm not sure if this is the intended way to write functions that return const-ref or mutable-ref but, just like you found Greg, that was the only way I could achieve it.

(In the real version of the above, there's already a public `get_group` function that returns the element by value, so I chose a different name.)

I didn't like this duplication so I went with the `get_group_idx` helper function instead.

The duplication also made me realise I really want C++23's "deducing this" in Cpp2, so I've created https://github.com/hsutter/cppfront/issues/1197 for that.

hsutter:
I wasn't expecting the PR to be closed -- was this resolved some other way, or are you waiting for `forward this` parameters or another issue/PR?

gregmarr:
@hsutter Replaced by #1212

bluetarpmedia:
Yes, #1212 covers this and future cases like it, but I forgot to add a note here to explain why I closed this one!

PR #1204, opened 2024-08-04
is(): correct is() for types

filipsajdak:
This change fixes the inspection of variables.

* correct inspection of polymorphic types,
* correct inspection of pointers,
* add support for nullptr_t

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1651,7 +1651,10 @@ auto is( X const& x ) -> bool {
         return Dynamic_cast<C const*>(&x) != nullptr;
     }
     else if constexpr (
-        requires { *x; X(); }
+        (
+            std::is_same_v<X, std::nullptr_t>
+            || requires { *x; X(); }
+        )
         && std::is_same_v<C, empty>
     )
     {
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1644,11 +1644,11 @@ auto is( X const& x ) -> bool {
         return true;
     }
     else if constexpr (
-            std::is_polymorphic_v<C>
-            && std::is_polymorphic_v<X>
+        std::is_polymorphic_v<C>
+        && std::is_polymorphic_v<X>
     )
     {
-            return Dynamic_cast<C const*>(&x) != nullptr;
+        return Dynamic_cast<C const*>(&x) != nullptr;
     }
     else if constexpr (
         requires { *x; X(); }
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1657,6 +1657,16 @@ auto is( X const& x ) -> bool {
     {
         return x == X();
     }
+    else if constexpr (
+        std::is_pointer_v<C>
+        && std::is_pointer_v<X>
+    )
+    {
+        if (x != nullptr) {
+            return is<std::remove_pointer_t<C>>(*x);
+        }
+        return false;
+    }
     else {
         return false;
     }
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -1635,13 +1635,13 @@ constexpr auto is( X&& ) {
 //  Types
 //
 template< typename C, typename X >
-auto is( X const& x ) -> bool {
+constexpr auto is( X const& x ) -> auto {
     if constexpr (
         std::is_same_v<C, X>
         || std::is_base_of_v<C, X>
     )
     {
-        return true;
+        return std::true_type{};
     }
     else if constexpr (
         std::is_polymorphic_v<C>
@@ -1666,12 +1666,12 @@ auto is( X const& x ) -> bool {
     )
     {
         if (x != nullptr) {
-            return is<std::remove_pointer_t<C>>(*x);
+            return bool{is<std::remove_pointer_t<C>>(*x)};
         }
         return false;
     }
     else {
-        return false;
+        return std::false_type{};
     }
 }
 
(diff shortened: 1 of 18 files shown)


include/cpp2util.h
```diff
@@ -1629,39 +1629,44 @@ constexpr auto is( X&& ) {
...
 template< typename C, typename X >
-auto is( X const& x ) -> bool {
+constexpr auto is( X const& x ) -> auto {
     if constexpr (
         std::is_same_v<C, X>
         || std::is_base_of_v<C, X>
     )
     {
-        return true;
+        return std::true_type{};
```
hsutter, line 1638: What are the benefits of returning `std::true_type` instead of `bool`?

My familiarity with `std::true_type` and `std::false_type` is with TMP, and in general the compile-time work we're doing here tries to avoid TMP.

include/cpp2util.h
```diff
@@ -1629,39 +1629,44 @@ constexpr auto is( X&& ) {
...
     }
     else if constexpr (
-        std::is_base_of_v<X, C>
-        || (
-            std::is_polymorphic_v<C>
-            && std::is_polymorphic_v<X>
-            )
+        std::is_polymorphic_v<C>
+        && std::is_polymorphic_v<X>
```
hsutter, line 1642: Is the `is_base_of` alternative redundant? Or is this fixing the semantics of a certain use case?

include/cpp2util.h
```diff
@@ -1629,39 +1629,44 @@ constexpr auto is( X&& ) {
...
+        && std::is_pointer_v<X>
+    )
+    {
+        if (x != nullptr) {
+            return bool{is<std::remove_pointer_t<C>>(*x)};
+        }
         return false;
     }
+    else {
+        return std::false_type{};
```
hsutter, line 1668: As above: Can you summarize the benefit of `false_type` over `bool`?

Also, I haven't fully checked, just asking: Are the choices of `[true|false]_type` vs `bool` paired so that the former or the latter is consistently used for a given kind of successful/failed test?

include/cpp2util.h
```diff
@@ -1629,39 +1629,44 @@ constexpr auto is( X&& ) {
...
+        && std::is_polymorphic_v<X>
     )
     {
-        if constexpr (std::is_pointer_v<X>) {
-            return Dynamic_cast<C const*>(x) != nullptr;
-        }
-        else {
-            return Dynamic_cast<C const*>(&x) != nullptr;
-        }
+        return Dynamic_cast<C const*>(&x) != nullptr;
```
hsutter, line 1645: Ah, because `is_pointer_v<X>` has to be `false` in this branch so it's dead code?

include/cpp2util.h
gregmarr: I believe that `true_type/false_type` means that the result is known at compile time because it's based solely on the types involved, and that `true/false` means that it's deferred to runtime because it depends on the value of the variable.

include/cpp2util.h
gregmarr: I commented on the same question below.

include/cpp2util.h
filipsajdak: Using `false_type`/`true_type` over `bool` gives the possibility of using the result in the compile-time. So, the result can be used in the `if constexpr`s or `requires` clause.

That means we know the result of some `is()` calls at compile time, which is useful for inspecting the types.

include/cpp2util.h
filipsajdak: In the following code:
```
        if constexpr (std::is_pointer_v<X>) {
             return Dynamic_cast<C const*>(x) != nullptr;
         }
```
the dead code is `return Dynamic_cast<C const*>(x) != nullptr;`

As `std::is_pointer_v<X>` will always be `false` just because to get to this `if`, we need to fulfill the following expression:
```
           (
            std::is_polymorphic_v<C>
            && std::is_polymorphic_v<X>
             )
```
That means `X` is polymorphic, and that means it is not a pointer (as a pointer does not fulfill the `is_polymorphic_v` type traits (see: https://godbolt.org/z/3j73cfWYe).

include/cpp2util.h
filipsajdak: Answered here: https://github.com/hsutter/cppfront/pull/1204#discussion_r1712342613

include/cpp2util.h
filipsajdak: The issue with `std::is_base_of_v<X, C>` is that it will also match non-polymorphic types for which dynamic_cast is undefined.

see: https://godbolt.org/z/GMvfrGzEh

filipsajdak: Ah, the following check
```
std::is_polymorphic_v<C>
        && std::is_polymorphic_v<X>
```
is enough to know that we can use `dynamic_cast` and it works for all combinations of polymorphic types.

See: https://godbolt.org/z/fhTbE31Ke

filipsajdak: I want to let you know that this PR is just the beginning of a series of PRs that correct or improve the `is()` and `as()`. The upcoming PRs (some of which I already requested) will improve/correct some of the cases that are still in this code.

I intend to make these changes digestive.

hsutter:
Thanks! I think all our review questions are resolved -- are there any more commits I should wait for before merging, or can I go ahead and merge now?

filipsajdak:
@hsutter please merge. I will rebase other PRs on top of this PR. thank you!

hsutter:
Thanks!

PR #1208, opened 2024-08-05
Update regression tests after recent changes

bluetarpmedia:


hsutter:
I hand-resolved that MSVC regex-10.execution file, it might not be quite right but for some reason it was sticking in my local GitHub environment's throat (it appeared to be a change I couldn't discard), so I just committed and pushed it and hand-resolved the conflict in this merge.

MaxSagebaum:
@bluetarpmedia Since you always do the regression updates. I just wanted to ask if there is any easy way to get all the diffs. Checking every version, searching for diff and then copying it out, is quite cumbersome. Is there an automated way to get all the diffs of all runners?

bluetarpmedia:
Yes, @jarzec modified the CI pipeline to create a handy aggregate of all the failed diffs.

Eg scroll down on this Summary page:
https://github.com/hsutter/cppfront/actions/runs/10327301935

Download the aggregate zip file, unzip into your cppfront directory and then `git apply` each diff file.

DyXel:
You probably can automate further with the Github CLI tool and/or a combination of scripts that uses Github APIs, probably something to look for in the future if more toolchains/targets are added.

MaxSagebaum:
Ah thank. That helped alot.

PR #1210, opened 2024-08-05
is(): refactor is() for std::optional to new design

filipsajdak:


filipsajdak:
I will organize these PRs differently

PR #1212, opened 2024-08-07
Disable some clang conversion warnings in `cpp2util.h` and cppfront itself

bluetarpmedia:
This PR disables some clang conversion warnings in 2 places:

- `cpp2util.h` for Cpp2 users
- cppfront translation unit when compiling cppfront

The warnings appear when compiling with clang with `-Wconversion` (as recommend by [Jason Turner](https://youtu.be/rWCvk4KZuV4?si=jNg9iXiz3_F1MeAl&t=412)).

`cpp2util.h` disables `Wsign-conversion` (only in that header) to allow warning-free signed integer types for indices and sizes.

The cppfront translation unit (via `common.h`) disables the above and also `Wstring-conversion` to enable implicit conversions of string literals to bools for `assert(!"message")`.

Some regression tests are updated because the line numbers have changed in `cpp2util.h`.

---

Note: there are some other 64-bit to 32-bit precision loss conversion warnings in cppfront itself (which this PR doesn't address) but I think we should fix them via casts rather than disabling the warning, since that warning may find other issues in the future.

```
$ clang++ source/cppfront.cpp -std=c++20 -o cppfront -Wall -Wconversion
source/cppfront.cpp:116:64: warning: implicit conversion loses integer precision: 'size_type' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
  116 |                         << std::right << std::setw(total_lines.size())
      |                                          ~~~       ~~~~~~~~~~~~^~~~~~
source/cppfront.cpp:119:64: warning: implicit conversion loses integer precision: 'size_type' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
  119 |                         << std::right << std::setw(total_lines.size())
      |                                          ~~~       ~~~~~~~~~~~~^~~~~~
source/cppfront.cpp:148:67: warning: implicit conversion loses integer precision: 'size_type' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
  148 |                             << std::right << std::setw(total_time.size())
      |                                              ~~~       ~~~~~~~~~~~^~~~~~
3 warnings generated.
```


hsutter:
Thanks! I've taken a pass and actually just removed the `!"string literal"` uses so that we don't need to disable the second Clang warning. And I've also addressed the three `.size()`-to-`int` cases you mentioned above so they're also addressed.

This seems to build clean and pass regression, but I don't have Clang 18+ installed -- before merging, please confirm that this looks right to you and still builds clean on the newer Clang with high warnings?

bluetarpmedia:
Looks good, no warnings with clang 18.

PR #1219, opened 2024-08-09
Separation of metafunction and runtime regex code.

### LINKED ISSUE #1199: [SUGGESTION] Only include `cpp2regex.h` when `@regex` is used to improve compile time
The generated `cpp2regex.h` file is ~170 KB and so I wondered what the impact on compile times are.

With clang 18 on ubuntu, I tested with and without including `cpp2regex.h` from inside `cpp2util.h` when compiling an empty `main` function lowered from Cpp2:

```cpp


#include "cpp2util.h"



    return 0; 
}
```

**With cpp2regex.h**
```
$ time clang++ -std=c++23 -stdlib=libc++ -fexperimental-library -Wall -Wextra -Wconversion source.cpp

real    0m2.628s
user    0m1.953s
sys     0m0.098s
```

**Without cpp2regex.h**
```
$ time clang++ -std=c++23 -stdlib=libc++ -fexperimental-library -Wall -Wextra -Wconversion source.cpp

real    0m1.138s
user    0m1.074s
sys     0m0.045s
```

I repeated the test a few times and the results were relatively stable. The real (wall clock) time is approx 2X more when bringing in `cpp2regex.h`, so I wondered if cppfront can detect whether it's needed and then conditionally include it?

MaxSagebaum:
Closes #1199.

This MR separates `include/cpp2regex.h2` into `include/cpp2regex.h2` and `source/regex.h2`. `cpp2regex.h2` contains all "runtime" functions for the regular expression matching. `regex.h2` contains the parsing and generation functionality.


hsutter:
Thanks! I've taken a pass at this to simplify (I hope), by moving all the metafunction stuff directly into `./source/reflect.h2` instead of into a new `./source` file.

Before I go further: Does that first step look okay?

Then my second step would be to remember a global flag if `@regex` is ever used, and then have `to_cpp1` emit a flag to enable including the regex runtime support, only in that case.

MaxSagebaum:
Looks good. 

The global flag is probably the simplest way to implement this. But when we want to allow users, to write their own metafunctions. This functionality is probably required.

hsutter:
Good point... better than a global flag is a function by which a metafunction can add a runtime `#include` or `import`. I'll pursue that direction...

hsutter:
OK, I've now finished adding to this PR, and I've updated the description to "closes #1199" which I think is accurate.

I think this is ready to merge. Could you please take a look to see if you spot anything more, and if you don't find anything I'll merge it. Thanks again!

MaxSagebaum:
Looks good. Adding the includes looks really simple. Would have needed quite some time to figure this out.

hsutter:
> Looks good. Adding the includes looks really simple. Would have needed quite some time to figure this out.

You share credit for that! This line in the PR to solve the problem in #1199:

```cpp
t.add_runtime_support_include( "cpp2regex.h" );
```

... was directly motivated by your comment above:

> But when we want to allow users, to write their own metafunctions. This functionality is probably required.

Thanks!

PR #1224, opened 2024-08-13
Update for sanitizing regex results and matcher strings.

### LINKED ISSUE #1223: [BUG] New clone of repo shows modification to `pure2-regex_10_escapes.cpp.execution`
Tested on Windows 11 and Ubuntu 22.04:

A new, fresh clone of the cppfront repo shows a change to `pure2-regex_10_escapes.cpp.execution`.

Repro:
```
$ git clone https://github.com/hsutter/cppfront.git
$ cd cppfront
$ git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   regression-tests/test-results/msvc-2022-c++latest/pure2-regex_10_escapes.cpp.execution

no changes added to commit (use "git add" and/or "git commit -a")
```

This message also appears in CI regarding the file:
```
warning: in the working copy of 'regression-tests/test-results/gcc-14-c++2b/pure2-regex_10_escapes.cpp.execution', CRLF will be replaced by LF the next time Git touches it
```

However, I'm not convinced the problem is with the line endings because other `.execution` files appear to have CRLF endings.

Instead I'm wondering if there's something in the file itself which is causing a problem. The GitHub preview stops printing the line numbers after [line 38](https://github.com/hsutter/cppfront/blob/main/regression-tests/test-results/msvc-2022-c%2B%2Blatest/pure2-regex_10_escapes.cpp.execution):

image

I had a look in Notepad++ and it shows the `\f` Form Feed, `\a` Bell and `\e` Esc characters, so maybe that confuses the GitHub preview.

image

MaxSagebaum:
The matcher strings and the result output for regular expression are now sanitized. Special characters like newline are replaced with the escaped version.

closes #1223


MaxSagebaum:
Letting the test suite run, so that I can update the test results.

MaxSagebaum:
@hsutter ready to merge

The last regression problem is about a temporary directory. I created a bug report for that https://github.com/hsutter/cppfront/issues/1225.

hsutter:
Thanks again for this! 🎉

hsutter:
Hmm, I just noticed that apparently as of this PR we broke a five regression test cases / compiler combinations?

mixed-type-safety-1: on GCC 10, Clang 18

pure2-type-safety-1: on GCC 14, Clang 18

pure2-requires-clauses: on GCC 10

I might be misreading and it might not have been due to this PR, but if I'm reading it correctly these five files disappeared as of this PR?

hsutter:
Ah, I see @filipsajdak has explained the reason in this comment: https://github.com/hsutter/cppfront/pull/1203#issuecomment-2295173063

It seems to be a requires-clause thing, rather than something about this PR. I'll go take a look.

Thanks again for this!

filipsajdak:
@hsutter here is the solution to above issues: #1250

PR #1230, opened 2024-08-17
Fix: cpp1 genertion for is-constraints on wildcarded params

### LINKED ISSUE #1229: [BUG] Bad generated cpp1 for is constraints on wildcarded parameters
67c665c introduces the possibility of constraining the wildcarded type with concepts. It works correctly for concepts taking one argument but fail for concepts taking two (or more) arguments.

E.g. for the following code:
```cpp
fun: (x: _ is std::convertible_to<bool>) = {
    if x {
        std::cout << "true" << std::endl;
    } else {
        std::cout << "false" << std::endl;
    }
}

main: () = {
    fun(true);
}
```

cppfront generates the following code (clean-cpp1 version):
```cpp


#include "cpp2util.h"


auto fun(auto const& x) -> void
CPP2_REQUIRES (std::convertible_to<bool><CPP2_TYPEOF(x)>) ;

auto main() -> int;
auto fun(auto const& x) -> void
requires (std::convertible_to<bool><CPP2_TYPEOF(x)>) {
    if (x) {
        std::cout << "true" << std::endl;
    }else {
        std::cout << "false" << std::endl;
    }
}

auto main() -> int{
    fun(true);
}
```

The bug is in the generated require clause, which is generated in the following way:
```cpp
requires (std::convertible_to<bool><CPP2_TYPEOF(x)>) {
```
And it should be generated in the following way:
```cpp
requires (std::convertible_to<CPP2_TYPEOF(x), bool>) {
```

filipsajdak:
Closes #1229

### DIFF
--- a/source/to_cpp1.h
+++ b/source/to_cpp1.h
@@ -4613,9 +4613,11 @@ public:
                 auto name = n.declaration->identifier->get_token();
                 assert(name);
                 auto req = print_to_string(*type_id.constraint);
-                req += "<CPP2_TYPEOF(";
-                req += *name;
-                req += ")>";
+                if (auto pos = req.find('<'); pos != req.npos) {
+                    req.insert(pos+1, "CPP2_TYPEOF(" + name->to_string() + "), ");
+                } else {
+                    req.append("<CPP2_TYPEOF("+ name->to_string() +")>");
+                }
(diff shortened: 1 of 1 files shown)


hsutter:
Good fix, thanks!

PR #1251, opened 2024-08-20
is()/as(): refactor of is() and as() for std::any and std::optional to new design (part 3 of n)

filipsajdak:
Aligning implementation of `is()` and `as()` to the rest of the functions.

Keep them as separate functions to make them examples of how to extend `is()`/`as()` with custom-made functions.

This PR is based on #1203


filipsajdak:
This PR has an experimental continuation here: https://github.com/hsutter/cppfront/pull/1267 that uses `type_safety` to handle failures.

hsutter:
Thanks!

PR #1253, opened 2024-08-21
Initial clang-tidy configuration

farmerpiki:
This adds some configuration to clang tidy, more to come in the future as well as more fixes to the code to go along with it.

Thanks.


hsutter:

farmerpiki:
Signed.

hsutter:
Got it, thanks Radu!

source/common.h
```diff
@@ -65,7 +65,7 @@ struct source_line
 {
     std::string text;
 
-    enum class category { empty, preprocessor, comment, import, cpp1, cpp2, rawstring };
+    enum class category : std::uint8_t { empty, preprocessor, comment, import, cpp1, cpp2, rawstring };
```
jarzec, line 68: Why is this suggested by clang-tidy? Is this meant to optimize for size? If that is the case it looks a bit like over-optimization.
It might be personal taste but the original seems nicer to read.

source/common.h
```diff
@@ -880,8 +880,8 @@ class cmdline_processor
         -> void
     {
         help_requested = true;
-        std::string_view a = __DATE__;
-        std::string_view b = __TIME__;
+        const std::string_view a = __DATE__;
+        const std::string_view b = __TIME__;
```
jarzec, line 884: How about:
```suggestion
        constexpr auto a = __DATE__;
        constexpr auto b = __TIME__;
```

source/common.h
```diff
@@ -880,8 +880,8 @@ class cmdline_processor
...
         std::unordered_map<std::string_view, char> m = { {"Jan",'1'}, {"Feb",'2'}, {"Mar",'3'}, {"Apr",'4'}, {"May",'5'}, {"Jun",'6'}, {"Jul",'7'}, {"Aug",'8'}, {"Sep",'9'}, {"Oct",'A'}, {"Nov",'B'}, {"Dec",'C'} };
```
jarzec, line 885: `m` also smells like it cold be `const`?

source/cppfront.cpp
```diff
@@ -72,7 +72,7 @@ auto main(
             }
             else {
                 std::cerr << arg.text << " - ambiguous compiler flag name, did you mean one of these?\n";
-                for (auto a : ambiguous) {
+                for (auto const &a : ambiguous) {
```
jarzec, line 75:
```suggestion
                for (auto &&a : ambiguous) {
```

source/common.h
gregmarr: Yes, it is meant to optimize for size.  https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html

source/reflect.h
```diff
@@ -2116,9 +2116,10 @@ declaration::declaration(declaration const& that)
         out_this_move_that.construct(declared.out_this_move_that != nullptr);
         inout_this_in_that.construct(declared.inout_this_in_that != nullptr);
         inout_this_move_that.construct(cpp2::move(declared).inout_this_move_that != nullptr);
-    return  { std::move(out_this_in_that.value()), std::move(out_this_move_that.value()), std::move(inout_this_in_that.value()), std::move(inout_this_move_that.value()) }; }
+        return  { std::move(out_this_in_that.value()), std::move(out_this_move_that.value()), std::move(inout_this_in_that.value()), std::move(inout_this_move_that.value()) }; // NOLINT(performance-move-const-arg)
```
jarzec, line 2119: This warning looks like it should rather be fixed than silenced.

source/reflect.h
```diff
@@ -3342,22 +3343,22 @@ std::string str {s}; return from_string(cpp2::string_util::replace_all(cpp2::mov
...
         token_vec new_list {}; 
         std::swap(new_list, cur_match_list);
         post_process_list(new_list);
         static_cast<void>(alternate_match_lists.insert(alternate_match_lists.end(), CPP2_UFCS_TEMPLATE(cpp2_new<regex_token_list>)(cpp2::shared, cpp2::move(new_list))));
     }
 
-#line 1680 "reflect.h2"
-    auto parse_context_group_state::swap(parse_context_group_state& t) & -> void{
+#line 1681 "reflect.h2"
+    auto parse_context_group_state::swap(parse_context_group_state& t) & -> void{// NOLINT(performance-noexcept-swap)
```
jarzec, line 3355: Can this method be marked `noexcept` instead?

source/common.h
filipsajdak: It might, but `operator[]` is not const in `std::unordered_map` (https://en.cppreference.com/w/cpp/container/unordered_map/operator_at).

`m` might be a const array with custom lookup... but I perceive it as premature optimization as the impact of this should be first measured.

gregmarr: Definitely a premature optimization for a function that is only ever called during command line processing, and thus likely only once per process.  If anything, this would be a pessimization in general because it might cause the compiler to create this at compile time and thus increase the size of the binary unnecessarily.

source/reflect.h
```diff
@@ -3626,19 +3627,20 @@ parse_context_branch_reset_state::parse_context_branch_reset_state(){}
...
-#line 1966 "reflect.h2"
+#line 1967 "reflect.h2"
     [[nodiscard]] auto parse_context::has_next() const& -> bool{return cpp2::impl::cmp_less(pos,regex.size()); }
 
-#line 1968 "reflect.h2"
+#line 1969 "reflect.h2"
     [[nodiscard]] auto parse_context::grab_until_impl(cpp2::impl::in<std::string> e, cpp2::impl::out<std::string> r, cpp2::impl::in<bool> any) & -> bool
     {
-        auto end {pos}; 
+        auto end {[&]() mutable -> auto{
```
jarzec, line 3636: The instantly evaluated lambda seems a bit of a mouthful here. The following should work:
```suggestion
        const auto end = any ? regex.find_first_of(e, pos) : regex.find(e, pos);
```

source/reflect.h2
```diff
@@ -1676,8 +1677,8 @@ parse_context_group_state: @struct type =
         _ = alternate_match_lists..insert(alternate_match_lists..end(), shared.new<regex_token_list>(new_list));
     }
 
-    //  Swap this state with the other one.
-    swap: (inout this, inout t: parse_context_group_state) = {
+    //  Swap this state with the other one. NOLINTNEXTLINE(performance-noexcept-swap)
+    swap: (inout this, inout t: parse_context_group_state) = { // NOLINT(performance-noexcept-swap)
```
jarzec, line 1681: This smells like it might be `noexcept`.

source/reflect.h
gregmarr: @jarzec probably better to review the `.h2` instead of the `.h`.

source/reflect.h
gregmarr: I had a similar thought.

source/reflect.h2
```diff
@@ -1967,13 +1968,14 @@ parse_context: type =
 
     private grab_until_impl: (inout this, in e: std::string, out r: std::string, any: bool) -> bool = 
     {
-        end:= pos;
+        end:= : () -> _ = {
```
jarzec, line 1971: Is there a ternary operator in cpp2?

source/to_cpp1.h
```diff
@@ -4882,9 +4882,9 @@ class cppfront
         function_type_node const& n,
         bool                      is_main                    = false,
         bool                      is_ctor_or_dtor            = false,
-        std::string               suffix1                    = {},
+        std::string const&        suffix1                    = {},
```
jarzec, line 4885: Default value for a const ref argument always made me cringe, but again, this might be personal taste 😉.
Nevertheless, wouldn't `std::string_view` passed by value work instead?

farmerpiki: definitely better with a std::string_view, I was just trying to do minimal cleanup with minimal changes

There's a lot more places in the code where a std::string_view could replace a string, it's on my TODO list... for both constness and performance reasons.

source/reflect.h2
farmerpiki: Not that I know of, this is the only way I could think of at the moment for fixing this lint, of course I could disable it instead, or we could fix the bug because the variable is initialized in both branches in the original `cpp2` code, however I couldn't satisfy `cppfront` by just declaring it, maybe there is a way... and I just don't know it yet.

source/reflect.h2
farmerpiki: Definitely, tell me how to write it and I will change it, it should be noexcept.

source/reflect.h
farmerpiki: the stuff in reflect.h is generated code, please provide alternative syntax for reflect.h2 and I will update it!

source/reflect.h
farmerpiki: the stuff in reflect.h is generated code, please provide alternative syntax for reflect.h2 and I will update it!

source/reflect.h
farmerpiki: should I write it up as a bug? can we even fix that, I don't think we know all the types of all variables and which of those are trivially copyable?

source/cppfront.cpp
farmerpiki: I'm not sure moving from std::string is in any way better, due to small string optimizations, but I'm not opposed to this change either.

source/common.h
farmerpiki: The `constexpr` I'll add.
`auto` would make for more extensive code changes, will change type to `char *`, which doesn't support `.substr`. Was going for minimal impact changes.

source/common.h
farmerpiki: It doesn't make an impact in any of the places I put it, I can turn the lint off and undo these changes if you'd prefer it that way.
I tried to keep as many of the checks enabled, only turning off the most noisy ones.
I'm totally ok with turning this off as well.

source/reflect.h2
farmerpiki:
```
        end: size_t;
        if any {
            end = regex..find_first_of(e, pos);
        }
        else {
            end = regex..find(e, pos);
        }
```

this should work in theory, but I get

reflect.h2...
reflect.h2(1973,13): error: local variable r must be initialized before end (local variables must be initialized in the order they are declared)

I'm not sure if this is a bug or not, I could move the `r = ""`; at the top instead but that will cause r to be initialized twice if the expression is found which I think is a pessimization.

I consider that it should work no matter the initialization order, or we will force people to write functions with akwardly placed parameters, so IMHO it's a bug! Do you also feel the same, should I open a bug?

source/cppfront.cpp
jarzec: Again, I don't see it as optimization, more as making it more flexible. I think in the current form the binary result will be identical.

source/reflect.h
jarzec: If this is generated shouldn't it be excluded from git?

gregmarr: No, because otherwise `cppfront` isn't self-hosting, you need a copy of `cppfront` to build `cppfront`.

source/reflect.h2
hsutter: > reflect.h2(1973,13): error: local variable r must be initialized before end (local variables must be initialized in the order they are declared)

That's as intended, because `r` is an `out` parameter and so for initialization purposes is treated the same as an uninitialized local variable right at the start of the function.

The main intent of requiring uninitialized locals to be initialized in declaration order is to ensure that their lifetimes still correctly nest when they're destroyed in reverse order, so that nested structured lifetime is preserved.

But you're right to point out that doesn't really apply to `out` parameters because those outlive the function anyway, so I could relax this.

hsutter (approved):
Thanks! Here are some comments, and I've already gone and applied them myself.

hsutter:
Looks good, we just have the reflect.h merge conflict... usually the line numbers changes don't cause a conflict, and maybe the simplest workaround is to move the reflect.h/h2 changes to a separate PR?

source/to_cpp1.h
```diff
@@ -632,7 +632,7 @@ class positional_printer
             && !pcomments
             && "ICE: tried to call .open twice"
         );
-        cpp1_filename = cpp1_filename_;
+        cpp1_filename = std::move(cpp1_filename_);
```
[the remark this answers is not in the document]
farmerpiki: yours seems better, it's a single copy, mine was copy + move

farmerpiki:
> Looks good, we just have the reflect.h merge conflict... usually the line numbers changes don't cause a conflict, and maybe the simplest workaround is to move the reflect.h/h2 changes to a separate PR?

I hid the initialization thing for now cause some didn't like the immediately invoked lambda expression..., without that the line number changes are no longer an issue

I think I'll leave it until you relax the rules for out parameters.

source/to_cpp1.h
farmerpiki: will have them in a follow-up pass, I rebased the whole thing on top of current master, with the suggested changes.

source/reflect.h2
farmerpiki: looking forward to the relaxation

source/reflect.h2
farmerpiki: that would be a welcome addition, I think

source/cppfront.cpp
farmerpiki: applied the & next to const as in the rest of code

source/reflect.h2
```diff
@@ -1967,13 +1968,14 @@ parse_context: type =
 
     private grab_until_impl: (inout this, in e: std::string, out r: std::string, any: bool) -> bool = 
     {
-        end:= pos;
```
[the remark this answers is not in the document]
farmerpiki: well, that was the thing, I added nolint for now instead as this was causing all the line changes

farmerpiki:
how can I get those workflows approved?

gregmarr:
I'm sure Herb is busy with CppCon right now.  The conference opens tomorrow and he has the opening keynote on Monday morning.

hsutter:
Thanks!

PR #1254, opened 2024-08-22
Fix doc typo: Updated `..<` instead of `...` for range operations in doc common.md

tsoj:


docs/cpp2/common.md
```diff
@@ -216,11 +216,11 @@ Postfix notation lets the code read fluidly left-to-right, in the same order in
 | `#!cpp --` | `#!cpp iter--` | `#!cpp --iter` |
 | `(` `)` | `#!cpp f( 1, 2, 3)` | `#!cpp f( 1, 2, 3)` |
 | `[` `]` | `#!cpp vec[123]` | `#!cpp vec[123]` |
-| `...` | `#!cpp v.begin()...v.end()` | `#!cpp std::ranges::subrange(v.begin(), v.end())` |
+| `..<` | `#!cpp v.begin()..<v.end()` | `#!cpp std::ranges::subrange(v.begin(), v.end())` |
```
gregmarr, line 219: @hsutter Should these be in the binary operators table instead of the unary operators table?

hsutter: Good point, I'll look at that post-merge

hsutter:
Thanks!

docs/cpp2/common.md
gregmarr: @hsutter 
> Note: The `(` `)`, `[` `]`, `..<`, and `..=` operators are treated as postfix unary operators, though they can take additional arguments.

I can kinda see how `()` and `[]` as pairs are considered postfix since they apply their contents to the thing to the left.  Can you explain how `..<` and `..=` are treated as postfix unary?  I don't understand what that means.

PR #1258, opened 2024-08-23
Fix: `int` variable casted with as to `size_t` does not fail on compile time - closes #1257

### LINKED ISSUE #1257: [BUG] int variable casted with as to size_t does not fail on compile time
In the current implementation of cppfront (065a993), the following code compiles and runs:
```cpp
i : int = 123;
std::cout << (i as size_t) << std::endl;
```
produces
```
123
```
and should fail to compile with the following error:
```
cppfront/include/cpp2util.h:3382:9: error: static assertion failed due to requirement 'program_violates_type_safety_guarantee<unsigned long, int>': 'as' does not allow unsafe possibly-lossy narrowing conversions - if you're sure you want this, use 'unsafe_narrow<T>' to explicitly force the conversion and possibly lose information
        static_assert(
        ^
narrowing.cpp2:10:31: note: in instantiation of function template specialization 'cpp2::impl::as_<unsigned long, int>' requested here
    std::cout << (cpp2::impl::as_<size_t>(cpp2::move(i))) << std::endl;
                              ^
1 error generated.
```

A potential error will be caught in runtime, e.g., when the `i` is negative, the following error will be produced:
```
cppfront/include/cpp2util.h(2486) decltype(auto) cpp2::impl::as(auto &&, std::source_location) [C = unsigned long, x:auto = int]: Type safety violation: dynamic lossy narrowing conversion attempt detected
libc++abi: terminating
```

It will save us from the trouble... but a user should be informed to use `cpp2::unsafe_narrow<size_t>(i)` instead.

The bug is related to the badly defined `is_narrowing_v` - I have the patch already and will deliver it in next PR.

filipsajdak:
The current implementation of `is_narrowing_v` does not handle type conversions correctly. For example, `cpp2::impl::is_narrowing_v<std::size_t, int>` returns `false`, indicating that it mistakenly perceives the conversion from a signed `int` to an unsigned `size_t` as potentially non-narrowing.

The new implementation is based on curly braces initialization on the cpp1 side which does all the required checking correctly. Examples here: https://godbolt.org/z/EdEK75E7z 

Closes #1257

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -741,6 +741,9 @@ concept valid_custom_is_operator = predicate_member_fun<X, F, &F::op_is>
                         || brace_initializable_to<X, argument_of_op_is_t<F>> 
                       );
 
+template <typename T>
+concept arithmetic = std::is_arithmetic_v<std::remove_cvref_t<T>>;
+
 //-----------------------------------------------------------------------
 //
 //  General helpers
@@ -1759,20 +1762,7 @@ inline constexpr auto is( X const& x, bool (*value)(X const&) ) -> bool {
 //  If it's confusing, we can switch this to <From, To>
 
 template< typename To, typename From >
-inline constexpr auto is_narrowing_v =
-    // [dcl.init.list] 7.1
-    (std::is_floating_point_v<From> && std::is_integral_v<To>) ||
-    // [dcl.init.list] 7.2
-    (std::is_floating_point_v<From> && std::is_floating_point_v<To> && sizeof(From) > sizeof(To)) ||
-    // [dcl.init.list] 7.3
-    (std::is_integral_v<From> && std::is_floating_point_v<To>) ||
-    (std::is_enum_v<From> && std::is_floating_point_v<To>) ||
-    // [dcl.init.list] 7.4
-    (std::is_integral_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    (std::is_enum_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    // [dcl.init.list] 7.5
-    (std::is_pointer_v<From> && std::is_same_v<To, bool>)
-    ;
+concept is_narrowing_v = arithmetic<To> && arithmetic<From> && !brace_initializable_to<From,To>;
 
 template< typename To, typename From >
 inline constexpr auto is_unsafe_pointer_conversion_v =
--- a/source/reflect.h
+++ b/source/reflect.h
@@ -3661,9 +3661,9 @@ parse_context_branch_reset_state::parse_context_branch_reset_state(){}
 #line 1993 "reflect.h2"
     [[nodiscard]] auto parse_context::grab_n(cpp2::impl::in<int> n, cpp2::impl::out<std::string> r) & -> bool
     {
-        if (cpp2::impl::cmp_less_eq(pos + cpp2::impl::as_<size_t>(n),regex.size())) {
-            r.construct(regex.substr(pos, cpp2::impl::as_<size_t>(n)));
-            pos += (cpp2::impl::as_<size_t>(n)) - 1;
+        if (cpp2::impl::cmp_less_eq(pos + cpp2::unsafe_narrow<size_t>(n),regex.size())) {
+            r.construct(regex.substr(pos, cpp2::unsafe_narrow<size_t>(n)));
+            pos += (cpp2::unsafe_narrow<size_t>(n)) - 1;
             return true; 
         }
         else {
@@ -3844,7 +3844,7 @@ parse_context_branch_reset_state::parse_context_branch_reset_state(){}
 
 #line 2178 "reflect.h2"
     auto generation_function_context::remove_tabs(cpp2::impl::in<int> c) & -> void{
-        tabs = tabs.substr(0, (cpp2::impl::as_<size_t>(c)) * 2);
+        tabs = tabs.substr(0, (cpp2::unsafe_narrow<size_t>(c)) * 2);
     }
 
     generation_function_context::generation_function_context(auto const& code_, auto const& tabs_)
--- a/source/reflect.h2
+++ b/source/reflect.h2
@@ -1992,9 +1992,9 @@ parse_context: type =
 
     grab_n: (inout this, in n: int, out r: std::string) -> bool = 
     {
-        if pos + n as size_t <= regex..size() {
-            r = regex..substr(pos, n as size_t);
-            pos += (n as size_t) - 1;
+        if pos + unsafe_narrow<size_t>(n) <= regex..size() {
+            r = regex..substr(pos, unsafe_narrow<size_t>(n));
+            pos += (unsafe_narrow<size_t>(n)) - 1;
             return true;
         }
         else {
@@ -2176,7 +2176,7 @@ generation_function_context: @struct type = {
     }
 
     remove_tabs: (inout this, c: int) = {
-        tabs = tabs..substr(0, (c as size_t) * 2);
+        tabs = tabs..substr(0, (unsafe_narrow<size_t>(c)) * 2);
     }
 }
 
--- a/regression-tests/test-results/apple-clang-15-c++2b/mixed-bounds-check.cpp.execution
+++ b/regression-tests/test-results/apple-clang-15-c++2b/mixed-bounds-check.cpp.execution
@@ -1,1 +1,1 @@
-../../../include/cpp2util.h(1007) decltype(auto) cpp2::impl::assert_in_bounds(auto &&, std::source_location) [arg = 5, x:auto = std::vector<int>]: Bounds safety violation: out of bounds access attempt detected - attempted access at index 5, [min,max] range is [0,4]
+../../../include/cpp2util.h(1010) decltype(auto) cpp2::impl::assert_in_bounds(auto &&, std::source_location) [arg = 5, x:auto = std::vector<int>]: Bounds safety violation: out of bounds access attempt detected - attempted access at index 5, [min,max] range is [0,4]
--- a/regression-tests/test-results/apple-clang-15-c++2b/mixed-bounds-safety-with-assert.cpp.execution
+++ b/regression-tests/test-results/apple-clang-15-c++2b/mixed-bounds-safety-with-assert.cpp.execution
@@ -1,1 +1,1 @@
-../../../include/cpp2util.h(819) : Bounds safety violation
+../../../include/cpp2util.h(822) : Bounds safety violation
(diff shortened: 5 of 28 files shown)


include/cpp2util.h
```diff
@@ -1736,20 +1739,7 @@ inline constexpr auto is( X const& x, bool (*value)(X const&) ) -> bool {
...
-    // [dcl.init.list] 7.2
-    (std::is_floating_point_v<From> && std::is_floating_point_v<To> && sizeof(From) > sizeof(To)) ||
-    // [dcl.init.list] 7.3
-    (std::is_integral_v<From> && std::is_floating_point_v<To>) ||
-    (std::is_enum_v<From> && std::is_floating_point_v<To>) ||
-    // [dcl.init.list] 7.4
-    (std::is_integral_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    (std::is_enum_v<From> && std::is_integral_v<To> && sizeof(From) > sizeof(To)) ||
-    // [dcl.init.list] 7.5
-    (std::is_pointer_v<From> && std::is_same_v<To, bool>)
```
gregmarr, line 1751: Doesn't this lose the pointer to bool test?  What about the enum handling?

source/reflect.h
```diff
@@ -3661,9 +3661,9 @@ parse_context_branch_reset_state::parse_context_branch_reset_state(){}
 #line 1993 "reflect.h2"
     [[nodiscard]] auto parse_context::grab_n(cpp2::impl::in<int> n, cpp2::impl::out<std::string> r) & -> bool
     {
-        if (cpp2::impl::cmp_less_eq(pos + cpp2::impl::as_<size_t>(n),regex.size())) {
-            r.construct(regex.substr(pos, cpp2::impl::as_<size_t>(n)));
-            pos += (cpp2::impl::as_<size_t>(n)) - 1;
+        if (cpp2::impl::cmp_less_eq(pos + cpp2::unsafe_narrow<size_t>(n),regex.size())) {
```
DyXel, line 3664: these days you can just write `unsafe_narrow` (without the `cpp2::`), I think.

include/cpp2util.h
filipsajdak: Hm... yes, this is lost. Interestingly, in our tests and in cppfront, there is no such call to cast ptr to bool with as().
If needed, I can bring it back.

gregmarr: It would be good to have tests for things which we know should fail.

DyXel: `as bool` seems nicer than `!= nullptr`, although the latter seems more explicit.

filipsajdak: On the other hand, it is perceived by the compiler as `narrowing`, so maybe it should require special treatment with `unsafe_narrow<bool>(ptr)`
```
<source>:59:19: warning: narrowing conversion of '& i' from 'int*' to 'bool' [-Wnarrowing]
   59 |     auto b = bool{&i};
```

source/reflect.h
filipsajdak: I don't know, I will try. This code should be generated by cppfront... I will check how it should be generated.

include/cpp2util.h
filipsajdak: I might have misunderstood you. In my current implementation, you cannot cast a pointer to bool. It ends up with an error:
```
cppfront/include/cpp2util.h:3515:9: error: static assertion failed due to requirement 'program_violates_type_safety_guarantee<bool, int *>': No safe 'as' cast available - please check your cast
        static_assert(
        ^
narrowing.cpp2:15:31: note: in instantiation of function template specialization 'cpp2::impl::as_<bool, int *>' requested here
    std::cout << (cpp2::impl::as_<bool>(cpp2::move(pi))) << std::endl;
```

The previous implementation also considers it as narrowing, so there is no change in behavior.

Check here: https://godbolt.org/z/5KzMPevrE

source/reflect.h
filipsajdak: cppfront generates it in that way:
```cpp
        if (cpp2::impl::cmp_less_eq(pos + cpp2::unsafe_narrow<size_t>(n),regex.size())) {
            r.construct(regex.substr(pos, cpp2::unsafe_narrow<size_t>(n)));
            pos += (cpp2::unsafe_narrow<size_t>(n)) - 1;
            return true; 
```

include/cpp2util.h
gregmarr: As long as `is_narrowing_v` is now only ever used for arithmetic types, then you're good with what you have, but then you don't really need `arithmetic<To> && arithmetic<From> &&` and `!brace_initializable_to<From,To>` is sufficient, and in that case, do you even need `is_narrowing_v`?  You can just use `!brace_initializable_to`.

I was assuming that `is_narrowing_v` was used for things other than arithmetic due to all the checks that were there before.

gregmarr: Now that I've gone looking at where this is used, I think the current behavior is actually correct.  Conversion from `int` to `size_t` is a runtime check:

```
    //  representable in the target type C is flagged as a type_safety contract violation
    else if constexpr (
        std::is_integral_v<C> &&
    )
    {
        const C c = static_cast<C>(CPP2_FORWARD(x));
        type_safety.enforce(   // precondition check: must be round-trippable => not lossy
            "dynamic lossy narrowing conversion attempt detected" CPP2_SOURCE_LOCATION_ARG_AS
        );
        return CPP2_COPY(c);
    }
```

filipsajdak: Yes, you are right. I try to minimize the change to make it easier for @hsutter to merge (I have prepared big PRs in the past that didn't go through). The current `is_narrowing_v` is only for arithmetic types. If you remove it, then the `as_()` function will also match for non-arithmetic types that will need additional treatment unrelated to arithmetic types. 
```cpp
template< typename C >
inline constexpr auto as_( auto&& x ) -> decltype(auto)
{
    if constexpr (is_narrowing_v<C, CPP2_TYPEOF(x)>) {
        static_assert(
            program_violates_type_safety_guarantee<C, CPP2_TYPEOF(x)>,
            "'as' does not allow unsafe possibly-lossy narrowing conversions - if you're sure you want this, use 'unsafe_narrow<T>' to explicitly force the conversion and possibly lose information"
        );
    }
    else if constexpr (is_unsafe_pointer_conversion_v<C, CPP2_TYPEOF(x)>)
//...
```

I will refactor more `is()` and `as()` so I will get there soon.

I have my big PR on my local branch, which I am introducing here step by step :) (adjusting it to @hsutter vision).

filipsajdak: > Now that I've gone looking at where this is used, I think the current behavior is actually correct. Conversion from `int` to `size_t` is a runtime check:

No, it is not correct. When you use `as` cast, it should fail as soon as possible. Just because `int` is signed and `size_t` is unsigned, the conversion might fail. We do similar checks for ints of various sizes. E.g. you cannot cast std::int16_t` to `std::int8_t`.

Check the following example:
```cpp
i: std::int16_t = 12;
j := i as std::int8_t;
```
it ends with:
```cpp
cppfront/include/cpp2util.h:3515:9: error: static assertion failed due to requirement 'program_violates_type_safety_guarantee<signed char, short>': 'as' does not allow unsafe possibly-lossy narrowing conversions - if you're sure you want this, use 'unsafe_narrow<T>' to explicitly force the conversion and possibly lose information
        static_assert(
        ^
narrowing.cpp2:18:25: note: in instantiation of function template specialization 'cpp2::impl::as_<signed char, short>' requested here
    auto j {cpp2::impl::as_<std::int8_t>(cpp2::move(i))}; 
                        ^
1 error generated.
```

source/reflect.h
DyXel: My bad, I meant to put the review inside the `.h2` file.

source/reflect.h2
```diff
@@ -1992,9 +1992,9 @@ parse_context: type =
 
     grab_n: (inout this, in n: int, out r: std::string) -> bool = 
     {
-        if pos + n as size_t <= regex..size() {
-            r = regex..substr(pos, n as size_t);
-            pos += (n as size_t) - 1;
+        if pos + cpp2::unsafe_narrow<size_t>(n) <= regex..size() {
```
DyXel, line 1995: In Cpp2 code you can use just `unsafe_narrow` these days. Cppfront will add the `cpp2::` when lowering.

filipsajdak: OK, will fix that.

filipsajdak: Fixed!

include/cpp2util.h
filipsajdak: It works differently when you use numeric literal. The following code is correct:
```cpp
s := 123 as size_t;
```
We can check if `123` literal fits in `size_t` at compile-time, so it works.
and because of that the following code:
```cpp
s := -123 as size_t;
```
will fail on compile time with message:
```
cppfront/include/cpp2util.h:3529:13: error: static assertion failed due to requirement 'program_violates_type_safety_guarantee<unsigned long, int>': 'as' does not allow unsafe possibly-lossy narrowing conversions - if you're sure you want this, use `unsafe_narrow<T>()` to explicitly force the conversion and possibly lose information
            static_assert(
            ^
narrowing.cpp2:19:25: note: in instantiation of function template specialization 'cpp2::impl::as_<unsigned long, -123>' requested here
    auto s {cpp2::impl::as_<size_t, -123>()};
```

gregmarr: > We do similar checks for ints of various sizes. E.g. you cannot cast std::int16_t to std::int8_t.

Yes, larger to smaller is compile time, but according to the code I quoted, when casting between different signed-ness and not going larger to smaller, it's runtime.

`int` to `size_t` should hit this code:

https://github.com/hsutter/cppfront/blob/main/include/cpp2util.h#L1834-L1871

Oh, there's another `as_` later which uses this.  What's the difference between `as` and `as_`?  I can't find any uses for `as_`.

filipsajdak: Static assert should be triggered whenever "from" type can represents values that cannot be represented by "to" type. That is why u16 to u8 will trigger assertion (it could be also runtime check but we decided to trigger it as soon as possible - so compile time). 

`as_` is generated by cppfront where you use `as` (check generated cpp file). It is old issue: https://github.com/hsutter/cppfront/pull/106#issuecomment-1373962682

gregmarr: > as_ is generated by cppfront where you use as (check generated cpp file).  I tried

gregmarr: >Static assert should be triggered whenever "from" type can represents values that cannot be represented by "to" type. That is why u16 to u8 will trigger assertion (it could be also runtime check but we decided to trigger it as soon as possible - so compile time).

That doesn't explain the explicit runtime check in the code for "a type with different unsignedness but same or larger size".  Is that code not actually being used?

hsutter:
Thanks! Looks good except for the conflicts -- can you take a pass through to resolve those? Then this should be ready to merge, thanks again.

gregmarr:
@hsutter Have you reviewed the discussion in #1257 about whether this change is actually wanted, or if other things should be updated to the current behavior here?

hsutter:
@gregmarr No, thanks for the reminder!

hsutter:
Re #1257: Thanks for the discussion! I'm going to stick with status quo on this for now, but please feel free to reopen if over time you find this keeps causing problems or user surprise.

PR #1263, opened 2024-08-27
CI Update default arg tests

jarzec:
As stated in the title.


regression-tests/test-results/gcc-14-c++2b/pure2-default-arguments.cpp.execution
```diff
@@ -1,4 +1,4 @@
 calling: int main(int, char**)
 012
-a newer compiler
+an older compiler
```
jarzec, line 3: Is this intended? 14 is the most recent version of GCC  released only a few months back.

gregmarr: I think this is a bug in https://github.com/hsutter/cppfront/blob/2e2359772a13fb707d039f9289deaf4683288ffb/include/cpp2util.h#L367 and they should be `>=` rather than `>`.

jarzec: It does smell like it. @hsutter do you confirm, should I fix it in this PR?

filipsajdak: This test does not test whether gcc-14 is modern but whether you can test its version at compile time.

jarzec: I agree, but it is still a bit strange to see a very recent compiler marked as old.
To me the name `gcc_clang_msvc_min_versions` suggest that if the actual version is equal to the value stored internally the version should be accepted.

include/cpp2util.h
```diff
@@ -364,7 +364,7 @@ constexpr auto gcc_clang_msvc_min_versions(
     auto msvc
 )
 {
-    return gcc_ver > gcc || clang_ver > clang || msvc_ver > msvc;
+    return gcc_ver >= gcc || clang_ver >= clang || msvc_ver >= msvc;
```
jarzec, line 367: After https://github.com/hsutter/cppfront/pull/1263#discussion_r1733509413 I've updated the code to accept the provided value to better fit the name (`gcc_clang_msvc_min_versions`).

regression-tests/test-results/gcc-14-c++2b/pure2-range-operators.cpp.execution
```diff
@@ -34,3 +34,6 @@ true
 true
 false
 false
+
+Make sure views::take works:
```
JarekGlobus, line 38: Now that GCC 14 is considered newer compiler it is allowed to do this as well.

jarzec, line 38: Now that GCC 14 is considered newer compiler it is allowed to do this as well.

regression-tests/test-results/clang-18-c++20/pure2-bugfix-for-ufcs-arguments.cpp copy.execution
jarzec, line 1: An accidentally committed file.

regression-tests/test-results/clang-18-c++23-libcpp/pure2-bugfix-for-ufcs-arguments.cpp copy.execution
jarzec, line 1: An accidentally committed file.

jarzec:
@hsutter If you agree with the outcome of the discussions this test update is ready to go.
I will address other test failures on the main branch in a separate PR.

hsutter:
Thanks!

Issue #1265, opened 2024-08-29
[BUG] cppfont 0.7.4 still showing 0.7.3 as version number


yann64:
**Describe the bug**

**To Reproduce**
compile cppfront from released **0.7.4** source package
run : 

> ~> cppfront -version
> 
> cppfront compiler **v0.7.3**   Build 9825:1501
> Copyright(c) Herb Sutter   All rights reserved
> 
> SPDX-License-Identifier: CC-BY-NC-ND-4.0
>   No commercial use
>   No forks/derivatives
>   Note: This license emphasizes that this is a personal
>         experiment; it will be upgraded if that changes

hsutter:
Thanks! Now fixed in `main`.

PR #1266, opened 2024-08-29
Update version.info - parity with release version number

yann64:
Correct version number
Solves #1265


hsutter:
Thanks! I made this change as part of my other commits so this has been done. Still, thanks for the catch!

PR #1267, opened 2024-08-30
as(): make as() cast to use type_safety.enforce

filipsajdak:
The current implementation of `as()` for `std::any`, `std::optional`, and `std::variant` throws exceptions in case of tunability to cast. This behavior is inconsistent with runtime checked `as()` for signed/unsigned integral cast where it calls `terminate()`.

This implementation switch runtime checked as() to use `type_safety`, which allows the setup of one approach for all `type_safety` issues.

The current behavior is to terminate, but it can be changed, e.g., to throw an exception. The following example is an excellent example of how it can be used.

```cpp
void throw_error(CPP2_MESSAGE_PARAM msg CPP2_SOURCE_LOCATION_PARAM) {
    throw std::runtime_error(std::string("Type safety exception: ") + msg);
}

make_throwable: (inout cg : cpp2::contract_group) -> _ = {
    h := cg.get_handler();
    sh := :(pcg : *cpp2::contract_group) = {
        pcg*.set_handler(h$);
    };
    cg.set_handler(throw_error);
    return std::unique_ptr<cpp2::contract_group, decltype(sh)>(cg&, sh);
}

void expect_no_throw(auto&& fun) try {
    fun();
} catch(std::exception const& e) {
    std::cout << e.what() << std::endl;
} catch(...) {
    std::cout << "Unknown exception!" << std::endl;
}

main: () = {
    o : std::optional<int> = ();

    (_ := make_throwable(cpp2::type_safety)) {
        expect_no_throw( :() = {
            std::cout << (o$ as int) << std::endl; // that will throw
        });
    }

    std::cout << (o as int) << std::endl; // that will terminate
}
```
and the result is the following:
```
Type safety exception: 'as' cast failed for 'std::optional'
Type safety violation: 'as' cast failed for 'std::optional'
libc++abi: terminating
```

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -2083,27 +2083,21 @@ auto as(X&& x CPP2_SOURCE_LOCATION_PARAM_WITH_DEFAULT_AS) -> decltype(auto)
 
 //  is Type
 //
-template<typename T, typename X>
-    requires (std::is_same_v<X,std::any> && !std::is_same_v<T,std::any> && !std::is_same_v<T,empty>)
-constexpr auto is( X const& x ) -> bool
-    { return x.type() == Typeid<T>(); }
-
-template<typename T, typename X>
-    requires (std::is_same_v<X,std::any> && std::is_same_v<T,empty>)
-constexpr auto is( X const& x ) -> bool
-    { return !x.has_value(); }
-
+template<typename T, std::same_as<std::any> X>
+constexpr auto is( X const& x ) -> bool{
+    if (!x.has_value()) {
+        return std::is_same_v<T,empty>;
+    }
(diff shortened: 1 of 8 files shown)


hsutter:
Hi! Just checking in as I clean up the PRs with the relicensing (see #1322), is still intended to be still a draft PR? If so perhaps I'll close it for now and you can reopen this or a new PR once it's ready to look at?

As described in #1322, this didn't make release 0.8, so I will need a new one-time CLA that will cover all your future contributions. I've emailed you the new CLA, and one it's completed I can look at PRs.

hsutter:
Hi! I haven't heard back so I'll just close this for now, but please reopen or replace with a new PR if you pick this up again. Thanks Filip!

PR #1270, opened 2024-09-03
Add a second `cpp2::range` constructor which takes 2 types, and deduce the common type for them (if available)

### LINKED ISSUE #1269: [BUG] `Wshorten-64-to-32` clang warning when using range `0 ..< vector.size()`
**Describe the bug**
Clang produces a `Wshorten-64-to-32` warning with a `cpp2::range` when it is initialised with an integer literal as the first member and an expression producing a `size_t`/`size_type` as the second member.

**To Reproduce**
Run cppfront on the following code and then compile with clang with `-Wconversion`:
```cpp
main: () = {
    v: std::vector = (11, 22, 33);

    for 0 ..< v.size() do (i) {    // Warning
        v[i]++;
    }
    std::println("v: {}", v);

    for 0uz ..< v.size() do (i) {  // No warning with `uz`
        v[i]++;
    }
    std::println("v: {}", v);
}
```
The warning is:
```
warning: implicit conversion loses integer precision: 'size_type' (aka 'unsigned long') to 'const std::type_identity_t<int>' (aka 'const int') [-Wshorten-64-to-32]
```
[Repro](https://cpp2.godbolt.org/z/Gjh89Pboj) on Godbolt

This is because the `cpp2::range` constructor deduces the type from only the first parameter:
```cpp
range(
    T const&                       f,
    std::type_identity_t<T> const& l,
    bool                           include_last = false
)
```

bluetarpmedia:
Use a CTAD deduction guide to deduce the `std::common_type` for both types (if a common one exists).

This allows code like the following to compile without integer precision conversion warnings:
```
for 0 ..< vector.size() do (e) { ... }
```
since the integer literal 0 and the size expression evaluate as different signed/unsigned types.

Also update the regression tests after recent changes.

Fixes #1269


feature-engineer:
Does putting `e - 1 < 0` inside the loop still fails?

hsutter:
Thanks!

@feature-engineer AFAICS this only affects `..<` and `..=` expressions so those would be unchanged.

Issue #1280, opened 2024-09-16
[SUGGESTION] Make main args usable with ranges


jhugard:
While toying with hello world, noticed that `cpp2:args` isn't usable with `std::ranges`.

Suggesting to update definition to satisfy both `std::bidirectional_iterator` and `std::ranges::bidirectional_range`, so that the following is possible:

```cpp2
main: (args) = {
    static_assert(std::bidirectional_iterator<cpp2::args::iterator>);
    static_assert(std::ranges::bidirectional_range<cpp2::args>);
    static_assert(std::ranges::viewable_range<cpp2::args>);

    // Output command line arguments, dropping the program name argc[0].
    for args | std::ranges::views::drop(1) do (arg)
    {
        std::cout
            << arg << std::endl;
    }
}
```

PR inbound...

PR #1281, opened 2024-09-16
Make cpp2::args usable as a std::ranges::bidirectional_range

jhugard:
Satisfy the `std::ranges::bidirectional_range` constraint, so that `cpp2::args` can be used with `std::ranges` adapters, etc.

See #1280.

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -2327,8 +2327,11 @@ struct args
         using difference_type = std::ptrdiff_t;
         using value_type      = std::string_view;
 
-        constexpr iterator() : argc(0), argv(nullptr), curr(0) {}
-        constexpr iterator(const iterator& o) : argc(o.argc), argv(o.argv), curr(o.curr) {}
+        constexpr iterator() = default;
+        constexpr iterator(const iterator& o) = default;
+        constexpr iterator(iterator&& o) = default;
+        constexpr iterator& operator=(const iterator& o) = default;
+        constexpr iterator& operator=(iterator&& o) = default;
 
         constexpr auto operator*() const {
             if (curr < argc) { return std::string_view{ argv[curr] }; }
@@ -2348,9 +2351,9 @@ struct args
         constexpr auto operator<=>(iterator const&) const = default;
 
     private:
-        int    argc;
-        char** argv;
-        int    curr;
+        int    argc = 0;
+        char** argv = nullptr;
+        int    curr = 0;
     };
 
     constexpr auto begin()  const -> iterator       { return iterator{ argc, argv, 0    }; }
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -2324,6 +2324,12 @@ struct args
     public:
         constexpr iterator(int c, char** v, int start) : argc{c}, argv{v}, curr{start} {}
 
+        using difference_type = std::ptrdiff_t;
+        using value_type      = std::string_view;
+
+        iterator() : argc(0), argv(nullptr), curr(0) {}
+        iterator(const iterator& o) : argc(o.argc), argv(o.argv), curr(o.curr) {}
+
         constexpr auto operator*() const {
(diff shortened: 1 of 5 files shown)


hsutter:

include/cpp2util.h
```diff
@@ -2324,6 +2324,12 @@ struct args
     public:
         constexpr iterator(int c, char** v, int start) : argc{c}, argv{v}, curr{start} {}
 
+        using difference_type = std::ptrdiff_t;
+        using value_type      = std::string_view;
+
+        constexpr iterator() : argc(0), argv(nullptr), curr(0) {}
```
gregmarr, line 2330: Should this instead set initializers on the members and then use `= default`?

include/cpp2util.h
```diff
@@ -2324,6 +2324,12 @@ struct args
...
+        constexpr iterator(const iterator& o) : argc(o.argc), argv(o.argv), curr(o.curr) {}
```
gregmarr, line 2331: The compiler-generated version should be enough.

include/cpp2util.h
jhugard: Good call.  Done.

include/cpp2util.h
```diff
@@ -2327,8 +2327,11 @@ struct args
         using difference_type = std::ptrdiff_t;
         using value_type      = std::string_view;
 
-        constexpr iterator() : argc(0), argv(nullptr), curr(0) {}
-        constexpr iterator(const iterator& o) : argc(o.argc), argv(o.argv), curr(o.curr) {}
+        constexpr iterator() = default;
+        constexpr iterator(const iterator& o) = default;
+        constexpr iterator(iterator&& o) = default;
+        constexpr iterator& operator=(const iterator& o) = default;
+        constexpr iterator& operator=(iterator&& o) = default;
```
gregmarr, line 2334: Did it not work without defaulting the copy/move functions?

jhugard: No, it does not.  Apologies for patronizing, but because a constructor (with parameters) was specified, the default ctors are suppressed.

Rule-of-zero, rule-of-five says if you define one, you should define them all: see core guidelines [C.ctor](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-ctor) C.21 & C.22).

gregmarr: Defining any constructor suppresses the default constructor, which is why you need a version with `=default` for that.

However, the default constructor is not part of the rule of 0/5 from C.21.  Those are copy/move construct/assign, and destructor.  They should all still be generated unless you define one of those 5.

https://en.cppreference.com/w/cpp/language/copy_constructor
>If no user-defined copy constructors are provided for a class type, the compiler will always declare a copy constructor as a non-explicit inline public member of its class.

Here's a table form of that information: https://mariusbancila.ro/blog/2018/07/26/cpp-special-member-function-rules/

jhugard: TIL, thank you!  I stand 100% corrected and have apparently misunderstood the rule of five for some time: I had been including the default constructor in that count, rather than the dtor.  Appreciate your link to the table form.  Very clear.

gregmarr: >TIL, thank you!

You're welcome.

> I had been including the default constructor in that count, rather than the dtor.

That's an easy mistake to make.  I'm pretty sure I've included it in the past.  :)

gregmarr:
@jhugard Did you return the CLA?

gregmarr (approved):

jhugard:
> @jhugard Did you return the CLA?

I am still waiting for approval from our legal department, which could take a while.

hsutter:
I'll close this for now, but please reopen if you get approval! Thanks again for your contribution.

jhugard:
No worries, totally understand the cleanup.  Will do if Legal ever gets back to me.  Thanks!

PR #1286, opened 2024-09-20
Fix: using statement in another namespace that pointer to type alias

HungMingWu:
Test case:
``` cpp2
A : namespace = {
        t1: type == int;
}

B : namespace = {
        using A::t1;
}

int main()
{
        return 0;
}
```

### DIFF
--- a/source/to_cpp1.h
+++ b/source/to_cpp1.h
@@ -6057,7 +6057,7 @@ public:
 
         //  Now, emit our own template parameters
         if (
-            n.template_parameters
+            (n.template_parameters && !n.is_alias())
             && (
                 printer.get_phase() <  printer.phase2_func_defs
                 || n.is_object()
--- a/source/to_cpp1.h
+++ b/source/to_cpp1.h
@@ -5685,10 +5685,11 @@ public:
             return;
         }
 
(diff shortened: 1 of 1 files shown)


hsutter:

hsutter:
Thanks! I agree that the above example isn't handled correctly so there is a bug, but when I tried this PR it fixes that example but breaks others and doesn't pass regression.

Here is an example that no longer works with this PR:

```cpp
A : namespace = {
    test: <T> type == std::vector<T>;
}
```

which in this PR lowers to:

```cpp
namespace A {
    template<typename T> using test = std::vector<T>;
}

namespace A {
    template<typename T> 
}

namespace A {

}
```

I didn't see a quick way to resolve the problem myself just now, but if you want to keep trying I'll review this PR again if you have an update that passes regression tests. Thanks!

hsutter:
Hi! I'm sorry I can't get this branch to self-build... rebuilding `reflect.h2` and `cppregex.h2` generates code that doesn't build. I haven't looked to see why (sorry).

For now I'm going to close this, but if you want to try again please reopen a new PR. Thanks again, I do appreciate your interest and contribution!

PR #1299, opened 2024-10-06
CI Update the GitHub checkout action

jarzec:
As stated in the title.
Some GitHub workflows use v3 of the checkout action that causes wirings. The PR updates the version to 4, as used e.g. for regression tests.


hsutter:
Thanks!

PR #1303, opened 2024-10-07
CI Update test outputs

jarzec:
Updates outdated regression test output files.
Test failures to be addressed separately.


JohelEGP:
#1291 addresses one such failure.

jarzec:
> #1291 addresses one such failure.

Well, I was rather thinking about fixing the code than modifying the tests to make the faulty code pass.
There is actually an [LLVM bug](https://github.com/llvm/llvm-project/issues/108095) I've reported that affects all (almost all) the Clang versions we use. Nevertheless an elegant workaround modifying the code itself seems achievable.

JohelEGP:
> Well, I was rather thinking about fixing the code than modifying the tests to make the faulty code pass.
> There is actually an [LLVM bug](https://github.com/llvm/llvm-project/issues/108095) I've reported that affects all (almost all) the Clang versions we use.

I do wonder.

I noticed this while working on #526,
before requiring parameters to be named `_`,
and lowering only the parameter's type.

I'm fairly sure that, at the time,
I checked with the C++ Standard whether Clang was right.
I think I found out Clang was indeed right,
and as a result I opened no compiler bugs.

jarzec:
Outdated, replaced with #1324

PR #1306, opened 2024-10-09
Exclude MSVC workaround on clang and gcc

gregmarr:
Resolves #1304 
On Windows, clang defines `_MSC_VER`, so the MSVC workaround needs to not happen when the clang define is set.

PR #1307, opened 2024-10-09
Fix Clang with MSVC standard library

MatthieuHernandez:
When compiling with Clang and MSVC as the standard library, clang defines the _MSC_VER and _MSC_FULL_VER macros.
Use the __clang_major__ macro to ensure that the code is compiled with MSVC.


MatthieuHernandez:
This is a duplicate of PR https://github.com/hsutter/cppfront/pull/1306.

PR #1319, opened 2024-10-17
Fix pure cpp1 header file generated using -import-std

MatthieuHernandez:
I've noticed that if I execute `cppfront.exe` with the `-import-std` option on a .h file that doesn't contain any cpp2 code, then an `#endif` will be missing at the end of the file.

⚠️  I haven't added any test because I didn't see any tests with .h files.


JohelEGP:
A source without Cpp2 should be pass-through
(i.e., the Cpp2 source and the generated Cpp1 source should compare equal).
So you shouldn't need to print anything extra.
The bug must be elsewhere.

Consider if that final `#endif` in the pure-Cpp1 Cpp2 source file actually had a space in front:
```Cpp2
 #endif
```
This solution would omit that space, and the resulting Cpp1 source wouldn't be pass-through.
I imagine something similar might occur with regards to an ending newline
(or maybe your original reproducer doesn't have it, which is what triggers the bug).

MatthieuHernandez:
The bug is in the following example:

gregmarr:
>A source without Cpp2 should be pass-through
>(i.e., the Cpp2 source and the generated Cpp1 source should compare equal).

Unless you specify the flag to either import or include the standard library, then it adds extra stuff.
https://github.com/hsutter/cppfront/blob/68b716abe3e4b623e875d68a2b9b3f0efcd39b45/source/to_cpp1.h#L1298-L1305

source/to_cpp1.h
```diff
@@ -1497,6 +1497,13 @@ class cppfront
         //
         if (!source.has_cpp2()) {
             assert(ret.cpp2_lines == 0);
+            if (
+                cpp1_filename.back() == 'h' 
+                && flag_import_std
```
gregmarr, line 1502: `(flag_import_std || flag_include_std)` to match https://github.com/hsutter/cppfront/blob/68b716abe3e4b623e875d68a2b9b3f0efcd39b45/source/to_cpp1.h#L1301-L1305

MatthieuHernandez: I force-pushed with your suggestion. Thanks 😊

JohelEGP (approved):

hsutter:
Thanks for your pull request! It looks like this may be your first contribution to cppfront. I've emailed you the Contributor License Agreement (CLA), and once it's completed I can look at your pull request. Thanks again for your contribution.

PR #1329, opened 2024-11-01
Fix: using statement in another namespace that pointer to type alias

HungMingWu:
Reopen PR #1286


hsutter:
Thanks for your pull request! This is your first PR since the project has permanently switched to a standard commercial OSS license, and I've emailed you the new Contributor License Agreement (CLA), which will cover this and all your future contributions. Once it's completed I can look at your PR. For more details see https://github.com/hsutter/cppfront/discussions/1322 .

hsutter:
Thanks!

I am not able to make progress evaluating this because it seems to need testing...

1. Test self-build by:
    a) Building `cppfront.cpp` as usual which builds from the Cpp1 sources
    b) Using that `cppfront.exe` to build `reflect.h2` and `cpp2regex.h2` (this should result in no surprising diffs) -- see the `main` branch's `build_h2.bat`
    c) Building `cppfront.cpp` again (this should work)
2. Test generating regression tests by:
    a) Building the Cpp2 tests into Cpp1 tests -- go to `regression-tests/test-results` and then run `run-tests.bat`
    b) Build the resulting Cpp1 tests using at least one compiler -- go to `regression-tests/test-results/<your preferred configuation>` and then run that directory's `run-tests-...` (a `.bat` or a `.sh`)

I'm currently getting failures at 1(c). I've just committed the results of 1(b) for reference.

HungMingWu:
Seems the main branch have modified on my previous code base commit, I will fix it later.

HungMingWu:
I find the root cause
Here is the example
``` cpp
namespace A {

template<bool> struct t { };
constexpr bool f(const t<true>&) { return true; }
constexpr t<true> o{};

} // namespace ns

A : namespace = {
        t1: <_: t<o.f()>> type == bool;
}

B : namespace = {
        using A::t1;
}

int main()
{
        return 0;
}
```
From main branch, it would become
``` cpp
#line 9 "test.cpp2"
namespace A {

}

namespace B {
        using A::t1;
}


#line 1 "test.cpp2"
namespace A {

template<bool> struct t { };
constexpr bool f(const t<true>&) { return true; }
constexpr t<true> o{};

} // namespace ns
```
`t1` is aliases from stage1, but my code will become
``` cpp
#line 9 "test.cpp2"
namespace A {
 template<t<CPP2_UFCS_NONLOCAL(f)(o)> UnnamedTypeParam1_1> using t1 = bool;
}

namespace B {
        using A::t1;
}


#line 1 "test.cpp2"
namespace A {

template<bool> struct t { };
constexpr bool f(const t<true>&) { return true; }
constexpr t<true> o{};

} // namespace ns
```
Looks it needs to move cpp1 namespace definition before stage0.

PR #1333, opened 2024-11-04
documentation: fix type

ChocolateChipKookie:
I just started reading into cppfront so I might be wrong, but it seems to me that the template type of the vector is incorrect.


hsutter:
Thanks!

PR #1335, opened 2024-11-09
Fix clang-18 compilation issue

jarzec:
Move the definition of `compound_statement_node::add_statement()` after the definition of `statement_node::~statement_node()` to fix compilation issue with clang-18.
Together with #1336 this PR makes all the tests pass on CI.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -1913,20 +1913,10 @@ struct compound_statement_node
         return open_brace;
     }
 
-    auto add_statement( 
+    auto add_statement(
         std::unique_ptr<statement_node>&& statement,
         int                               before_pos
-    )
-        -> bool
-    {
-        //  Adopt this statement into our list of statements
-        statements.insert( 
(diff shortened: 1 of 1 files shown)


source/parse.h
```diff
@@ -1913,20 +1913,10 @@ struct compound_statement_node
...
 
-    auto add_statement( 
+    auto add_statement(
         std::unique_ptr<statement_node>&& statement,
         int                               before_pos
-    )
-        -> bool
-    {
-        //  Adopt this statement into our list of statements
-        statements.insert(
```
jarzec, line 1923: [Explanation] This call causes deletion of the `std::unique_ptr<statement_node>`. As discussed in #1088  this causes trouble for clang-18 if `compound_statement_node::add_statement()` is defiend before `statement_node::~statement_node()`.

jarzec:
I've rebased the branch on current main.
The PR is ready to be merge (the remaining CI test failures are solved in #1336).

hsutter:
Looks good, thanks!

PR #1355, opened 2025-01-14
fix compilation errors with exceptions disabled

jmplanes02:
Currently, compiling with -fno-exceptions will not work because `cpp2util.h` uses try-catch and throw without checking if the macro `CPP2_NO_EXCEPTIONS` is defined.

I've done the following and now compiling with exceptions disabled works fine:

* Give `cpp2::string_util::string_to_int` function an implementation when `CPP2_NO_EXCEPTIONS` is set
* Use `Throw` instead of `throw` inside `cpp2::range` constructor

### DIFF
--- a/include/cpp2util.h
+++ b/include/cpp2util.h
@@ -557,6 +557,7 @@ constexpr bool is_escaped(std::string_view s) {
 }
 
 inline bool string_to_int(std::string const& s, int& v, int base = 10) {
+#ifndef CPP2_NO_EXCEPTIONS
     try {
         v = stoi(s, nullptr, base);
         return true;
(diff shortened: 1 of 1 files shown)


hsutter:
Thanks! It looks like this may be your first contribution to cppfront. I've sent the Contributor License Agreement (CLA) to your email, and once it's signed I can look in more detail at your pull request. Thanks again for your contribution.

jmplanes02:
I've just made my email public on my GitHub account, but in any case, I’ve also sent you my information to your gmail.

hsutter:
Looks good, thanks!

hsutter:
Oh, by the way I inadvertently left in a temporary change from `#ifndef CPP2_NO_EXCEPTIONS` to `#ifdef CPP2_NO_EXCEPTIONS` (as an expedient way to test the other path). I'll change that back as a separate commit.

PR #1365, opened 2025-02-05
1328 Fixed issue comparing different numerical types as equal.

### LINKED ISSUE #1328: [BUG] `is` type for `std::optional` can return unexpected results
**Describe the bug**
An expression testing if a `std::optional` is a type can return unexpected results, depending on whether the `optional`'s type is convertible to the target type. (I'm assuming this behaviour is unintended.)

**To Reproduce**
Run cppfront on the following and then compile with a C++ compiler:
```cpp
main: () -> int = {
    o: std::optional<int> = 5;
    if o is int {
        std::cout << "int!\n";
    }
    if o is bool {
        std::cout << "bool!\n";
    }
    if o is float {
        std::cout << "float!\n";
    }
    if o is std::string {
        std::cout << "std::string!\n";
    }
    return 0;
}
```
The output is:
```
int!
bool!
float!
```
[Repro](https://cpp2.godbolt.org/z/GravGeq8f) on Godbolt.

I think this behaviour is due to this cast:
https://github.com/hsutter/cppfront/blob/ddde3d833bec373081a3c0f419250f8e8b553805/include/cpp2util.h#L2042

CC @filipsajdak

mihaiboros:
Closes #1328


mihaiboros:
I would like to add some regression tests to this change. Any pointers of how to do that? Thanks.

hsutter:
Thanks!

PR #1370, opened 2025-02-16
CI Update regression test results after recent mods

jarzec:
Update test results missed in recent updates:
- new test output.
- modified line numbers in error messages.

Compilation issues with `pure2-regex-general` using Clang 18 remain and will be investigated separately.


hsutter:
Thanks!

PR #1371, opened 2025-02-17
Fix .h2 not converting to .h in the #include section

LT2H:
Remove the (*) since std::string::ends_with does not support wildcards.


hsutter:
Thanks! Nice catch.

This is tiny so I won't ask for a CLA, but if you intend to submit any future PRs please email me (first dot last at gmail) with your full name and I'll send you a one-time CLA to sign that will cover your future PRs.

Issue #1342, opened 2024-11-29
[BUG] Enum member called 's' is not possible


threeifbyair:
**Describe the bug**
When declaring an enumeration, a member called 's' cannot be declared.

**To Reproduce**

https://cpp2.godbolt.org/z/eExnreoaT

`foo : @enum type = {
    s;
}`

gives the following error:
> main.cpp2: error: a type's implementation may not declare a name that is the same as (i.e., shadows) a type scope name - for example, a type scope function's local variable may not have the same as one of the type's members

's' is not mentioned among the illegal member names for an enumeration.

gregmarr:
I believe that this happens due to these functions:

```
public: [[nodiscard]] static auto from_string(cpp2::impl::in<std::string_view> s) -> foo;
public: [[nodiscard]] static auto from_code(cpp2::impl::in<std::string_view> s) -> foo;
```

mihaiboros:
@hsutter This can be closed after commit https://github.com/hsutter/cppfront/commit/bf77f37b78863efa583fd62bf520e304fbe29901

threeifbyair:
Agreed, looks fixed to me, thanks!

PR #1373, opened 2025-02-22
1342 enum member called s is not possible

mihaiboros:
Fix https://github.com/hsutter/cppfront/issues/1342
Not a definitive fix, but it will reduce the chance of name clashes and shadowing.

### DIFF
--- a/regression-tests/pure2-enum.cpp2
+++ b/regression-tests/pure2-enum.cpp2
@@ -25,6 +25,11 @@ file_attributes: @flag_enum<u8> type = {
     cached_and_current := cached | current;
 }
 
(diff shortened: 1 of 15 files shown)


hsutter:
Thanks for pointing out this problem.

In general, this highlights a problem I hadn't noticed before between generated code and user code... the `@enum` metafunction generates a function that has a parameter named `s`, and so the style rule not to shadow a type scope name fired... which highlights that such style rules create a tension between generated code and user-written code.

I'll create a commit remove the style rule instead of taking this fix, so I'll close this PR, but thanks for highlighting the problem!

PR #1375, opened 2025-03-02
CI Update regression tests after recent mods

jarzec:
Another update of regression test results still excluding the SIGSEGVs cause in Cland 18.1 in `pure2-regex-general.cpp2`.


hsutter:
Thanks!

hsutter:
> excluding the SIGSEGVs cause in Cland 18.1 in `pure2-regex-general.cpp2`

Is this something that needs to be fixed in the cppfront regex source, or is it a Clang 18 bug?

jarzec:
> > excluding the SIGSEGVs cause in Cland 18.1 in `pure2-regex-general.cpp2`
> 
> Is this something that needs to be fixed in the cppfront regex source, or is it a Clang 18 bug?

I hope to be able to find what is causing the problem, although that is not evident. Maybe it is possible to modify the test slightly to work around the issue.
In any case, a SIGSEGV inside the compiler is a bug on their side. I was thinking of trying to narrow it down a bit locally, which would help working around it. In addition this would allow for reporting a more concrete bug at https://github.com/llvm/llvm-project/issues, like what I did for [this issue](https://github.com/llvm/llvm-project/issues/108095).

Issue #1376, opened 2025-03-25
[BUG] Type with @polymorphic_base and an "out this" function that is not a constructor causes cppfront to assert


threeifbyair:
**Describe the bug**
When a type defines an "out this" function that is not a constructor, and is annotated with @polymorphic_base, we get into function_type_node::is_constructor() and fire an assertion because the function is not "operator=".

If instead I move the assertion conditions into the if statement, everything compiles fine (both the cpp2 and the generated cpp).

**To Reproduce**
```
Machine: @polymorphic_base <I:int> type = {
    operator%: (out this, _: std::string) = {}
}
```

**Discussion**
I'm doing some fuzzing on cppfront, and this is the first of many crashes I've found. What I'm not sure of is whether a non-constructor with 'out this' is legal. I suspect it isn't, and I'm assuming we should catch it in semantic analysis (in the function_type_node start/end visitor).

I'll put in a PR with a fix and we can go from there.

hsutter:
Fixed by #1377

PR #1377, opened 2025-03-26
#1376: Fix non-constructors with 'out this'

threeifbyair:
As mentioned in #1376, a non-constructor with an 'out this' parameter currently crashes cppfront. This fixes the crash, outlaws non-constructors with 'out this' parameters, and adds a regression test for it.


hsutter:
Looks good, thanks!

Only one super small thing: Test files that result in errors should be named `*-error.cpp2`.

I'm just re-running regression tests now to update the results.

PR #1380, opened 2025-03-26
Create a simple makefile and allow regressions to use it.

threeifbyair:
One of the frustrating things about cppfront is that I always have to remember the precise incantation to build cppfront. In addition, the regression tests always build cppfront whether or not it needs to be built.

This PR adds a simple Makefile. Nothing special and explicitly marked as optional. In addition, it allows the regression tests to be run with a pre-built version of cppfront. This allows us (for instance) to build a sanitized cppfront and run it through the regressions to see if we get any memory or undefined behavior errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,11 @@ cppfront/x64/Debug/microsoft/STL/std.compat.ixx.ifc.dt.module.json.command
 cppfront/x64/Debug/microsoft/STL/std.ixx.ifc.dt.d.json
 cppfront/x64/Debug/microsoft/STL/std.ixx.ifc.dt.module.json
 cppfront/x64/Debug/microsoft/STL/std.ixx.ifc.dt.module.json.command
+
+# Crashes and hangs
+crashes/
+hangs/
+
+# Generated files
+cppfront.exe
+
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,45 @@
+#  Copyright 2022-2025 Herb Sutter
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#  
+#  Part of the Cppfront Project, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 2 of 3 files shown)


jarzec:
> Thanks! @jarzec you're the expert on that, do you think this will improve regression tests, and do you approve merging this PR?

**Some context**
A while back, after your CppCon 22 presentation, Cpp2 caught interest of many people (myself included). This interest resulted in a number of PRs with CMake setup for cppfront. At that stage you closed those as you didn't want to add extra dependencies. That is why I proposed a pure bash sctipt that can be used on MasOS, Linux and Windows (through git bash) without any extra deps installed. This is what I used for the GitHub actions running for the regression tests.
The idea of using a build system for cppfront was not buried, though. In [the Related Projects section of the wiki](https://github.com/hsutter/cppfront/wiki#related-projects) you point to a number of forks that propose a CMake/Mason setup.

**My thoughts**
It turns out GNU Make is installed on the Windows GitHub runners (which I didn't know), so it seems to be a possible way to go.
However, if adding Make for regression tests starts to be an option I would suggest a discussion about a more complete and flexible system like CMake or the like. The source code of cppfront is constantly growing while still being basically a single translation unit with really huge files.
CMake would also greatly facilitate integration of cppfront in other projects.
As I mentioned before, this is might be a good topic for [a discussion](https://github.com/hsutter/cppfront/discussions).

On the other hand, adding a command line option to provide the executable to the `run-test.sh` wouldn't hurt It would e.g. help with developing new tests without having to recompile cppfront every time the test is run.

@hsutter What do you say?

MaxSagebaum:
I also prefer makefiles for a quick setup. Usually they also do not block the path for manual compilation. Since the project is kept quite simple.

Why not have both. A makefile and cmake file can exist next to each other. Use the simple makefile for the ci and cmake can be used to integrate it into other projects.

jarzec:
@hsutter I think you can merge this one.

PR #1381, opened 2025-03-27
Fix missing explicit for single-argument constructors

LT2H:
We should count `out this` as a parameter as well to ensure single-argument constructors are prefixed with `explicit`.


JohelEGP:
See also <https://github.com/hsutter/cppfront/commit/9901b017936678148457eea9af5313d348a6fe5c#r150292780>.

PR #1383, opened 2025-03-30
Fix a bunch of crashes discovered by fuzzing cppfront

### LINKED ISSUE #1382: [BUG] Method followed by inheritance crashes cppfront
**Describe the bug**
I've been finding little subtle bugs in cppfront by fuzzing, and I have a number of little subtle fixes that I'm going to put in a PR once I have more of them, but this one is trickier and I don't really know what the best fix is, so I'm submitting it as a bug and would love to hear suggestions!

**To Reproduce**
```
crash_m0b: type = {
}

crash_m0c: type = {
  name: i32;
  get_name: (this) -> i32 = { return name; }
  this: crash_m0b;
}
```

When lowering to cpp1, any data members (like `name`) of a type that are listed before its superclass get moved into a little class of their own, and that class becomes another superclass of the type. But any functions listed before the superclass (like `get_name`) just cause cppfront to segfault.

I _think_ the right thing to do is to do a semantic analysis pass and outlaw functions listed before a superclass, but I'm not certain and I'd love to hear better ideas!

threeifbyair:
Here are fixes for several cppfront crashes discovered from fuzzing.

(1) Errors found in metafunctions cause cppfront to end with a signal instead of exiting cleanly.
(2) Two #else operations cause an assertion instead of an error.
(3) $ is allowed in preconditions and assertions but crashes cppfront when it tries to emit cpp1.
(4) 'unsigned char' close to the end of the file causes an assertion.
(5) Comments near the end of the file cause an assertion.
(6) If the initializer of a parameter is not an expression, cppfront segfaults.
(7) If a type defines a function before stating a superclass, cppfront crashes. (Closes #1382)
(8) Some invalid aliases crash cppfront.
(9) 'decltype' in the wrong place crashes cppfront.

Of these, only (5) was found on legal C++2 code. The rest should have generated (and now do generate) errors.


hsutter:
Thanks! This is a nice set of fixed, appreciated. Fuzzing FTW 🏅

PR #1391, opened 2025-05-01
CI: Remove unsupported Ubuntu 20.04 runs

jarzec:
GitHub runners based on Ubuntu 20.04 are no longer available.

This PR removes respective runs from the regression test workflow.


hsutter:
Thanks!

PR #1393, opened 2025-05-14
Make all the regression tests pass on CI

jarzec:
Make all the regression tests pass on CI:
- Update Clang 18 to 19 for the regression tests to avoid failures due to bugs in Clang 18.
- Fix failures of pure Cpp2 regression tests on Windows GitHub runners.
  - The tests fail due to:
     ```
     cpp2util.h(564): error C2065: 'EXIT_FAILURE': undeclared identifier`.
     ```
  - `#include <cstdlib>` for the same tests leads to errors of the kind:
    ```
    C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\corecrt_math.h(444): error C2995: 'bool isfinite(_Ty) noexcept': function template has already been defined
    C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\corecrt_math.h(444): note: see declaration of 'isfinite'
    C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\corecrt_math.h(450): error C2995: 'bool isinf(_Ty) noexcept': function template has already been defined
    C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\corecrt_math.h(450): note: see declaration of 'isinf'
    ...
    ```
  - The temporary solution to allow all tests to pass on CI before a better approach is found is simply:
    ```
    #ifndef EXIT_FAILURE
        #define EXIT_FAILURE 1
    #endif
    ```
 - Update regression tests data accordingly:
    - Update line numbers within `cpp2util.h` for compilation failure messages.
    - Remove error message files for test that started to pass with Clang 19.
    - Add binary execution output files for the tests that started to pass.


hsutter:
Thanks, including for the `EXIT_FAILURE` catch!

PR #1394, opened 2025-05-23
Added meson build script

KaruroChori:
At the moment there is no easy way to build and install cppfront. 
This PR adds a meson build file to do that.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,30 @@
+project(
+  'cppfront',
+  ['cpp'],
+  version: files('./source/version.info'),
+  meson_version: '>= 1.1',
+  default_options: ['cpp_std=c++20'],
+)
+
+threads_dep = dependency('threads', required : true)
+
+executable('cppfront', './source/cppfront.cpp',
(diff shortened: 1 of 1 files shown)


DyXel:
There's a recipe by Jussi Pakkanen to integrate cppfront within an existing project: https://nibblestew.blogspot.com/2022/10/using-cppfront-with-meson.html . Maybe you could try exposing that functionality in this script (via declare_dependency and co.) so people can just add a subproject and start using cppfront immediately.

KaruroChori:
Done :). This way it should be possible to either use it as a subproject dependency or to install it.

hsutter:
Thanks! However I'm trying hard not to support any particular build system... the build instructions for cppfront aim to be [just one short command](https://hsutter.github.io/cppfront/welcome/overview/#how-do-i-get-and-build-cppfront).

If you host this somewhere though I could add a link [on the wiki home page](https://github.com/hsutter/cppfront/wiki) "related projects" section.

PR #1405, opened 2025-06-10
Add runtime check for modulo by zero

### LINKED ISSUE #1399: [SUGGESTION] Add runtime checks for C++ integer modulo by zero
This is #1184 but for modulo... sounds like it was forgotten that you can not mod by zero either.

There is also a case of division and modulo of `std::numeric_limits<T>::min() / -1` which is UB aswell. Not sure if you want to check it too - it requires to check both operands.

LT2H:
Checks for modulo and modulo-assignment by zero at runtime. Based on a suggestion in #1399.
Resolves #1399.

### DIFF
--- a/regression-tests/test-results/mixed-initialization-safety-3.cpp
+++ b/regression-tests/test-results/mixed-initialization-safety-3.cpp
@@ -68,5 +68,5 @@ auto print_decorated(auto const& x) -> void{
 }
 
 #line 32 "mixed-initialization-safety-3.cpp2"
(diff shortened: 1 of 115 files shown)


hsutter:
Thanks!

hsutter:
> @hsutter Did you intend to switch from importing std to including std when running the tests?

That's probably a good idea. @jarzec was maintaining and updating the test runner (thanks again!) -- do those use MSVC modules today which would have required that modules be built somehow (I vaguely recall seeing something like that)? If they do they should probably also switch to use `-in` to include std headers instead of importing the std module, which is what I did in this commit.

PR #1406, opened 2025-06-22
Fix C++1 generation of function pointers.

### LINKED ISSUE #1397: [BUG] Can not declare functions that return function pointers
**Describe the bug**
I copy pasted an example from wiki (https://github.com/hsutter/cppfront/wiki/Design-note:-Postfix-operators) to play with and it does not parse.

**To Reproduce**
Sample code:
```cpp
f: (i: int) -> * (j: int) -> string = { return nullptr; }
```
Result:
```
main.cpp2...
main.cpp2(1,25): error: missing ';' at end of declaration or '=' at start of initializer (at '->')
main.cpp2(1,1): error: unexpected text at end of Cpp2 code section (at 'f')
main.cpp2(1,0): error: parse failed for section starting here
```

Run as `./source/cppfront main.cpp2 -p` with cppfront built from commit 5aa32aef7c74679994d6da39e6d0cf9b9714e1ee.

**Additional context**
Would be nice if it got fixed with a unit test for something like this:
```cpp
// C stdlib and POSIX function
void (*signal(signum int, void (*handler)(int)))(int)
// expected C++2 syntax
signal: (signum: int, handler: *(_: int) -> void) -> *(_: int) -> void
```

threeifbyair:
Fix #1397. Function pointers (but not functions) can now be return types.

It turns out that even if we correctly parse a function pointer type, we drop the pc_qualifiers when we generate the C++1 code. So I fixed that too.

### DIFF
--- a/source/parse.h
+++ b/source/parse.h
@@ -7266,7 +7266,8 @@ private:
     auto type_id(
         bool allow_omitting_type_name = false,
         bool allow_constraint         = false,
-        bool allow_function_type      = false
+        bool allow_function_type      = false,
+        bool allow_qualified_function_type = true
     )
         -> std::unique_ptr<type_id_node>
     {
@@ -7339,7 +7340,9 @@ private:
             assert (n->id.index() == type_id_node::unqualified);
         }
         else if (std::unique_ptr<function_type_node> id = {};
-            allow_function_type
+            (allow_function_type || 
+                (allow_qualified_function_type && 
+                !n->pc_qualifiers.empty()))
             && (id = function_type({})) != nullptr
             )
         {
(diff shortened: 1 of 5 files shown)


MaxSagebaum:
Looks good.

source/parse.h
```diff
@@ -7266,7 +7266,8 @@ class parser
     auto type_id(
         bool allow_omitting_type_name = false,
         bool allow_constraint         = false,
-        bool allow_function_type      = false
+        bool allow_function_type      = false,
+        bool allow_qualified_function_type = true
```
hsutter, line 7270: This new parameter doesn't seem to be used - it always has its default value. Should it just be removed?

source/parse.h
```diff
@@ -7339,7 +7340,9 @@ class parser
             assert (n->id.index() == type_id_node::unqualified);
         }
         else if (std::unique_ptr<function_type_node> id = {};
-            allow_function_type
+            (allow_function_type || 
+                (allow_qualified_function_type &&
```
hsutter, line 7344: And here?

hsutter (changes requested):
Thanks! I have a question about the new parameter...

source/parse.h
threeifbyair: Yes. Done.

PR #1407, opened 2025-06-23
Bugfix: Update for regex 20 lookbehind test.

MaxSagebaum:
Removed cpp file and added cpp2 file.


hsutter:
Thanks! I pushed a commit that re-runs the regression tests to keep everything in sync.

While doing that, I noticed there seems to be a problem for MSVC here with `cpp2regex.h2(591): error C2668: 'std::make_reverse_iterator': ambiguous call to overloaded function`. Here's 591:

    r := func(make_reverse_iterator(cur), make_reverse_match_context(ctx), true_end_func());

It seems the similar line 577 should also get this error, but doesn't because maybe that one is never invoked?

Conjecture: Maybe

   - lines 577 and 591 make*iterator calls should be std:: qualified, and

   - the make*iterator helpers on lines 184-189 should just be removed as they don't seem to be used?

But I'm not sure so I just pushed the update to regression tests in without any code changes.

MaxSagebaum:
I will have a look during the week.

MaxSagebaum:
Seems to be clean from a regex point of view. There are still some errors from MSVC at other places.

PR #1413, opened 2025-08-14
Feature/autodiff

MaxSagebaum:
Additional features for AD through metafunctions. 

 - Multiplication
 - Addition
 - Loops
 - Ifs
 - Function calls (only in same context yet)
 - Intermediate variables
 - Higher order derivatives.


hsutter:
Thanks, Max!

PR #1418, opened 2025-09-02
Wide scanners

jnorthrup:


MaxSagebaum:
Why do you even open up the pull requests? Usually you have to do several clicks to open one. Are you using some automation?

PR #1428, opened 2025-10-03
[POC] Named function arguments using @name

RPeschke:
# [POC] Named function arguments using `@name`

**Status:** This is a **proof of concept** only. It does **not** meet the quality, testing, or design bar for merging into the main repository. The goal is to explore the idea and see whether there is interest in this concept at all.

## Summary

Many modern programming languages provide **named function arguments**. This PR experiments with bringing a similar feature into the cppfront/C++ world. In this proposal, the `@` symbol introduces a **name** which can then be used, for example, as a named function argument.

### Syntax (POC)

* `@name = value` introduces a name-value pair usable at call sites and in lightweight record-like objects.

## Example file

`examples/named_function_arguments.cpp2` demonstrates calling a function with named arguments in any order:

```cpp
variadic_template( 1, 2.5, @argument4 = std::string("five"), @argument3 = 12 );
```

It also shows using the same mechanism to create variadic objects:

```cpp
nt1 := nt::ntuple{
    @a = 1,
    @b = 2.5,
    @c = std::string("three"),
    @d = '4'
};

std::cout << nt1.a << std::endl;  // prints 1
std::cout << nt1 << std::endl;    // prints | a: 1 | b: 2.5 | c: three | d: 4 |
```

## Question

Would the authors be interested in seeing this concept developed further (with proper design notes, diagnostics, tests, and performance considerations), or should it remain a one-off experiment?

### DIFF
--- a/examples/named_function_arguments.cpp2
+++ b/examples/named_function_arguments.cpp2
@@ -1,0 +1,35 @@
+#include "ntuples/full.hh"
+
+
+
+
+template <typename ...Args> 
+auto variadic_template(Args&& ...args) -> void{
+        auto t0 = bind_args(
+            @argument1 = 1, 
+            @argument2 = 15, 
+            @argument3 = nt::required<int>(), 
+            @argument4 = std::string("hello"))(
+            CPP2_FORWARD(args)...);
+
+        std::cout << t0 << std::endl;
+}
+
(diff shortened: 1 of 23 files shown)


RPeschke:
Ah ok thanks for the link

PR #1443, opened 2026-02-01
CI Add singleton test files missing for some compilers

jarzec:
Adds missing test files for some compilers.

NOTE: The compilation of the C++ code fails for some versions of GCC.
This will be investigated separately.


hsutter:
> NOTE: The compilation of the C++ code fails for some versions of GCC.
> This will be investigated separately.

BTW if you mean the unused return value in the singleton test file, I see the bug -- it's a bug in the test, I'll fix it in my next commit.

PR #1444, opened 2026-02-21
CI Update singleton test results

jarzec:
Removes compilation error reports in `pure2-singleton-and-static` regression test for the GCC 13 and Clang 19 compilers.
The generated C++ code no compiles after recent update of the test.


hsutter:
Thanks!

Issue #1446, opened 2026-02-24
[BUG] Cannot handle assignment to u8 and i16 from literal values


jmafc:
**Describe the bug**
If a variable is declared as `u8` and a character literal is assigned to it, or a variable is declared as `i16` and an integer literal is assigned to it, the generated code cannot be compiled by clang++ (g++ compiles but with warnings). I suspect this also applies to the respective signed/unsigned counterparts.

**To Reproduce**
1. Sample code - distilled down to minimal essentials please
```test.cpp2
main: () -> int = {
   foo: u8;
   foo = 'a';
   bar: i16;
   bar = 3;
}
```
 
2. Command lines including which C++ compiler you are using
```
cppfront test.cpp2 -p
clang++ -std=c++20 [or 23] -I<path/to/cppfront/include> test.cpp
```
cppfront compiler v0.8.2 Build B110:2055
Debian clang version 19.1.7 (3+b1)
[also] g++ (Debian 14.2.0-19) 14.2.0

3.  Expected result - what you expected to happen

Clean, no error compilation.

4. Actual result/error

clang++ fails when trying to expand `#define CPP2_FORWARD(x)             std::forward<decltype(x)>(x)` with the following note:
```
test.cpp2:3:8: note: in instantiation of function template specialization 'cpp2::impl::deferred_init<unsigned char>::construct<char>' requested here
    3 |    foo.construct('a');
```

**Additional context**
The first problem can be bypassed if `foo` is declared as `char` (or `_schar` but that is probably platform dependent). However, the second one remains, even if `bar` is declared as `short`, since (AFAIK) there is no way of indicating an integer literal is a `short`.

dima-starosud:
The fix #1449 is on its way, waiting for review.

PR #1449, opened 2026-07-15
Fix deferred init narrowing #1446

dima-starosud:
### Fix deferred init narrowing https://github.com/hsutter/cppfront/issues/1446 by keeping brace-init in caller context.

- Use construct_from(lambda) so constexpr initializers are not lost across a call boundary, matching immediate initialization behavior.
- Note, change out::construct to use brace-init in all paths to be consistent and to align with deferred init.
- Add regression tests for literals (negative too), constexpr, aggregates.
- Update snapshots affected by the change.


include/cpp2util.h
```diff
@@ -1488,11 +1488,19 @@ class deferred_init {
...
     constexpr  deferred_init() noexcept       { }
     constexpr ~deferred_init() noexcept       { destroy(); }
     constexpr auto value()     noexcept -> T& { cpp2_default.enforce(init);  return t(); }
 
-    constexpr auto construct(auto&& ...args) -> void { cpp2_default.enforce(!init);  new (&data) T{CPP2_FORWARD(args)...};  init = true; }
+    constexpr auto construct(auto&& ...args) -> void {
+        construct_from([&]() -> T { return T{CPP2_FORWARD(args)...}; });
+    }
+
+    constexpr auto construct_from(auto&& factory) -> void
```
dima-starosud, line 1501: With the factory, all the values are still on the caller context, which means their constexpr-ness isn't lost during crossing the call boundaries.

include/cpp2util.h
```diff
@@ -1535,10 +1545,16 @@ class out {
...
+        requires std::same_as<decltype(CPP2_FORWARD(factory)()), T>
+    {
         if (has_t || called_construct()) {
-            if constexpr (requires { *t = T(CPP2_FORWARD(args)...); }) {
+            if constexpr (requires { *t = CPP2_FORWARD(factory)(); }) {
                 cpp2_default.enforce( t );
-                *t = T(CPP2_FORWARD(args)...);
+                *t = CPP2_FORWARD(factory)();
```
dima-starosud, line 1557: Here, instead of parens-init we switched to braces-init to align with deferred init, which is also used in `out` down below.

regression-tests/test-results/mixed-autodiff-taylor.cpp
```diff
@@ -140,44 +140,44 @@ return  { std::move(y0.value()), std::move(y.value()) }; }
       cpp2::impl::deferred_init<double> y0;
       cpp2::impl::deferred_init<taylor> y;
 #line 39 "mixed-autodiff-taylor.cpp2"
-  y0.construct(sqrt(x0));
-  y.construct(CPP2_UFCS(sqrt)(x, x0));
+  y0.construct_from([&]() -> typename CPP2_TYPEOF(y0)::value_type { return typename CPP2_TYPEOF(y0)::value_type{sqrt(x0)}; });
+  y.construct_from([&]() -> typename CPP2_TYPEOF(y)::value_type { return typename CPP2_TYPEOF(y)::value_type{CPP2_UFCS(sqrt)(x, x0)}; });
```
dima-starosud, line 144: We could introduce a macro for this. If we decide to, I can bring it in this PR.

regression-tests/test-results/clang-19-c++20/pure2-bugfix-for-deferred-init-narrowing-matrix-error.cpp.output
```diff
@@ -0,0 +1,57 @@
+pure2-bugfix-for-deferred-init-narrowing-matrix-error.cpp2:19:125: error: non-constant-expression cannot be narrowed from type 'cpp2::impl::in<cpp2::i16>' (aka 'const short') to 'signed char' in initializer list [-Wc++11-narrowing]
```
dima-starosud, line 1: This tests errors, i.e. failing cases, when we do want the compilation to fail.

